PR #17, opened 2016-06-15
Add implementedByAsync to PlayRouting endpoints

vdebergue:
**Problem:** If a service is defined as `A => Future[B]` it was not possible to return a result in the play implementation.

This add a new method on `Endpoint`: `implementedByAsync` to solve this use case.


julienrf:
Hi, thanks for pointing this out!

I think we should just change the current `EndpointWithHandler` case class to have a `service: A => Future[B]` member instead of having both `EndpointWithHandlerSync` and `EndpointWithHandlerAsync`. What do you think? I can do these changes if you want :)

julienrf:
Thanks!

Issue #23, opened 2016-11-06
[QUESTION] Using this library with akka-http ?


hajime-moto:
Just wondering if there are plans to add support for akka-http ?

julienrf:
At the moment I don’t plan to take time to do that, but that would be a nice solution to #15. Would you be interested in contributing an akka-http interpreter? I can help you a bit in this work :)

hajime-moto:
I haven't coded in years so I don't think I will be much help here :( .. But I going to ask one of my developers to contribute. Are there any guidelines on how the interpreter should be written or should we just reference the code for Play ?

julienrf:
I would more than warmly welcome such a contribution!

Currently the project is more in the form of an experiment, so there is not much documentation. I’m working to gradually improve the situation :)

Yes, drawing inspiration from the PlayRouting and PlayClient interpreters is a good idea. I can also give help on gitter: https://gitter.im/julienrf/endpoints

ngbinh:
@julienrf this library looks quite interesting to me. I can probably take shot at this issue.

julienrf:
Great! Le me know if you need any additional information.

PR #25, opened 2017-01-07
Partial scala 2.12 support

Krever:
Hi,

I had to add sbt-doge, so different sub-projects can support different scala versions. We need to wait for `faithful` and `play` to fully support scala 2.12. 

I have updated circe(it required some code changes because Xor was replaced by scala.util.Either) and scalajs.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,6 @@ import org.scalajs.sbtplugin.cross.CrossProject
 
 val commonSettings = Seq(
   organization := "org.julienrf",
-  scalaVersion := "2.11.8",
   scalacOptions ++= Seq(
     "-feature",
     "-deprecation",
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -200,6 +224,7 @@ val `example-basic-server` =
 
 val endpoints =
   project.in(file("."))
+    .enablePlugins(CrossPerProjectPlugin)
```
julienrf: It doesn’t work without this line?

julienrf:
Looks great, thanks!

Play not supporting 2.12 is yet another motivation to solve #23!

build.sbt
Krever: It works, but without this line you need to use specific plugin command instead of "+" to perform cross-action(I'm not sure what is actual command, but docs gives `> ;so clean; such test; very publishLocal` as an example). With this line you can just execute `+ publishLocal` and subprojects `crossScalaVersion`s will be taken into consideration.

Krever:
> Play not supporting 2.12 is yet another motivation to solve #23!

Yep, this will be my next step :)

build.sbt
julienrf: Great, thanks for the explanation!

PR #27, opened 2017-01-09
Methods algebra

### LINKED ISSUE #26: Abstract over http method
Currently we have hard coded(in `EndpointAlg`) support for only `get` and `post` methods. I'm currently working on a solution that will allow abstracting over it. I will make a PR once I have some working proof of concept.

Krever:
Fixes #26 

I'm not sure about naming, because `request` crashes a little with `jsonRequest`. 

Any suggestions are very welcome.

### DIFF
--- a/algebra/src/main/scala/endpoints/EndpointAlg.scala
+++ b/algebra/src/main/scala/endpoints/EndpointAlg.scala
@@ -7,7 +7,7 @@ import scala.language.higherKinds
   *
   * Requests and responses contain headers and entity.
   */
-trait EndpointAlg extends UrlAlg {
+trait EndpointAlg extends UrlAlg with MethodsAlg {
 
   /** Information carried by requests’ headers */
   type RequestHeaders[A]
@@ -28,21 +28,25 @@ trait EndpointAlg extends UrlAlg {
   type RequestEntity[A]
 
   /**
-    * Request that uses the GET verb
-    *
-    * @param url Request URL
-    * @param headers Request headers
+    * Empty request.
     */
-  def get[A, B](url: Url[A], headers: RequestHeaders[B] = emptyHeaders)(implicit tupler: Tupler[A, B]): Request[tupler.Out]
+  def emptyRequestEntity: RequestEntity[Unit]
+
 
   /**
-    * Request that uses the POST verb
+    * Request for given parameters
     *
+    * @param method Request method
     * @param url Request URL
     * @param entity Request entity
     * @param headers Request headers
     */
-  def post[A, B, C, AB](url: Url[A], entity: RequestEntity[B], headers: RequestHeaders[C] = emptyHeaders)(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out]
+  def request[A, B, C, AB](
+    method: Method,
+    url: Url[A],
+    entity: RequestEntity[B] = emptyRequestEntity,
+    headers: RequestHeaders[C] = emptyHeaders
+  )(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out]
 
   /** Information carried by a response */
   type Response[A]
--- a/algebra/src/main/scala/endpoints/EndpointAlg.scala
+++ b/algebra/src/main/scala/endpoints/EndpointAlg.scala
@@ -7,7 +7,7 @@ import scala.language.higherKinds
   *
   * Requests and responses contain headers and entity.
   */
-trait EndpointAlg extends UrlAlg with MethodsAlg {
+trait EndpointAlg extends UrlAlg with MethodAlg {
 
   /** Information carried by requests’ headers */
   type RequestHeaders[A]
@@ -47,6 +47,24 @@ trait EndpointAlg extends UrlAlg with MethodsAlg {
     entity: RequestEntity[B] = emptyRequestEntity,
     headers: RequestHeaders[C] = emptyHeaders
   )(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out]
+
+  /**
+    * Helper method to perform GET request
+    */
+  def get[A, C, AUnit](
+    url: Url[A],
+    headers: RequestHeaders[C] = emptyHeaders
+  )(implicit tuplerAB: Tupler.Aux[A, Unit, AUnit], tuplerABC: Tupler[AUnit, C]): Request[tuplerABC.Out] = request(Get, url, headers = headers)
+
+  /**
+    * Helper method to perform POST request
+    */
+  final def post[A, B, C, AB](
+    url: Url[A],
+    entity: RequestEntity[B],
+    headers: RequestHeaders[C] = emptyHeaders
+  )(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out] = request(Post, url, entity, headers)
+
 
   /** Information carried by a response */
   type Response[A]
--- a/algebra/src/main/scala/endpoints/EndpointAlg.scala
+++ b/algebra/src/main/scala/endpoints/EndpointAlg.scala
@@ -51,10 +51,10 @@ trait EndpointAlg extends UrlAlg with MethodAlg {
   /**
     * Helper method to perform GET request
     */
-  def get[A, C, AUnit](
+  def get[A, C](
     url: Url[A],
     headers: RequestHeaders[C] = emptyHeaders
-  )(implicit tuplerAB: Tupler.Aux[A, Unit, AUnit], tuplerABC: Tupler[AUnit, C]): Request[tuplerABC.Out] = request(Get, url, headers = headers)
+  )(implicit tuplerAC: Tupler[A, C]): Request[tuplerAC.Out] = request(Get, url, headers = headers)
 
   /**
     * Helper method to perform POST request
--- a/algebra/src/main/scala/endpoints/MethodsAlg.scala
+++ b/algebra/src/main/scala/endpoints/MethodsAlg.scala
@@ -1,0 +1,16 @@
+package endpoints
+
+
+trait MethodsAlg {
+
+  type Method
+
+  def Get: Method
+
+  def Post: Method
+
+  def Put: Method
+
+  def Delete: Method
+
+}
--- a/play-client/src/main/scala/endpoints/EndpointPlayClient.scala
+++ b/play-client/src/main/scala/endpoints/EndpointPlayClient.scala
@@ -9,7 +9,7 @@ import scala.concurrent.{ExecutionContext, Future}
   * Play’s [[WSClient]] HTTP client.
   * @param wsClient The underlying client to use
   */
-class EndpointPlayClient(wsClient: WSClient)(implicit ec: ExecutionContext) extends EndpointAlg with UrlClient {
+class EndpointPlayClient(wsClient: WSClient)(implicit ec: ExecutionContext) extends EndpointAlg with UrlClient with MethodsClient {
 
   /**
     * A function that, given an `A` and a request model, returns an updated request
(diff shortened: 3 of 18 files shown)


algebra/src/main/scala/endpoints/MethodsAlg.scala
```diff
@@ -0,0 +1,16 @@
+package endpoints
+
+
+trait MethodsAlg {
+
+  type Method
+
+  def Get: Method
```
julienrf: Just a minor comment: I’d rather use lower cased identifiers: `get`, `post`, etc.

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
```diff
@@ -28,21 +28,20 @@ class EndpointPlayClient(wsClient: WSClient)(implicit ec: ExecutionContext) exte
...
-      val (a, b) = tupler.unapply(ab)
-      val wsRequest = wsClient.url(url.encode(a))
-      headers(b, wsRequest).get()
-    }
 
-  def post[A, B, C, AB](url: Url[A], entity: RequestEntity[B], headers: RequestHeaders[C])(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out] =
+  override def request[A, B, C, AB](
+    method: Method, url: EndpointPlayClient.this.Url[A],
+    entity: RequestEntity[B], headers: EndpointPlayClient.this.RequestHeaders[C]
+  )(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): EndpointPlayClient.this.Request[tuplerABC.Out] =
```
julienrf: What is the motivation for using explicitly qualified type names (e.g. `EndpointPlayClient.this.RequestHeaders` vs `RequestHeaders`)?

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
```diff
@@ -28,21 +28,20 @@ class EndpointPlayClient(wsClient: WSClient)(implicit ec: ExecutionContext) exte
...
 
-  def get[A, B](url: Url[A], headers: RequestHeaders[B])(implicit tupler: Tupler[A, B]): Request[tupler.Out] =
-    (ab: tupler.Out) => {
-      val (a, b) = tupler.unapply(ab)
-      val wsRequest = wsClient.url(url.encode(a))
-      headers(b, wsRequest).get()
-    }
 
-  def post[A, B, C, AB](url: Url[A], entity: RequestEntity[B], headers: RequestHeaders[C])(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out] =
+  override def request[A, B, C, AB](
```
julienrf: Also, I prefer to omit the `override` keyword when we *implement* something rather than really *overriding* it. It avoids actually overriding things by mistake.

algebra/src/main/scala/endpoints/MethodsAlg.scala
```diff
@@ -0,0 +1,16 @@
+package endpoints
+
+
+trait MethodsAlg {
```
julienrf: The current naming convention is to use singular nouns for algebra interfaces: `MethodAlg`.

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
Krever: Intellij automation, I had to miss it during reviewing changes.

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
Krever: I like to put `override` because when you make some mistake(in types mostly) compiler warns you about that. So you are safe from cases when you want to override something but in the end you don't.

algebra/src/main/scala/endpoints/MethodsAlg.scala
Krever: That's intentional, to differentiate from methods I have deleted. I will describe more in general comment to PR.

Krever:
I have noticed, that user-facing dsl has changed, and we can safely put it back, by creating functions that will just call `request()` with fixed `method` argument. I will create a draft tomorrow together with other fixes.

Krever:
I have restored previous API. Now I just wonder about few things:
1. Should we put such helper methods for other http methods(`put`, `delete`) ?
2. Should we support other HTTP methods(`patch`, `options`, etc.) ?
3. Should we put `or` combinator for methods to allow defining endpoint that accept more than one method.

For all of these I decided it is not required now and can always be introduced later without significantly bigger cost, but if you have other opinion I can add some or all of these features.

Also, I would like to rename `jsonRequest` to `jsonEntity` or `jsonBody` and I'm curious about your opinion on this.

algebra/src/main/scala/endpoints/EndpointAlg.scala
```diff
@@ -48,6 +48,24 @@ trait EndpointAlg extends UrlAlg with MethodsAlg {
     headers: RequestHeaders[C] = emptyHeaders
   )(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out]
 
+  /**
+    * Helper method to perform GET request
+    */
+  def get[A, C, AUnit](
+    url: Url[A],
+    headers: RequestHeaders[C] = emptyHeaders
+  )(implicit tuplerAB: Tupler.Aux[A, Unit, AUnit], tuplerABC: Tupler[AUnit, C]): Request[tuplerABC.Out] = request(Get, url, headers = headers)
```
julienrf: The `AUnit` type parameter is not useful here, you should have only two type parameters (one for the url and one for the headers).

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
julienrf: > So you are safe from cases when you want to override something but in the end you don't.

IMHO, this is not really a problem because in such a case at some point the compiler will complain that some method is not implemented.

On the other hand, if you always put the `override` keyword you can unintentionally override a method that was already implemented.

I wish I had a `@implement` annotation that would emit an error if you make a mistake in the type signature of the method you implement… That would be far better than waiting for the point where the compiler complains that there are unimplemented members.

julienrf:
> 1. Should we put such helper methods for other http methods(put, delete) ?
> 2. Should we support other HTTP methods(patch, options, etc.) ?

We can add them afterwards, no hurry.

> 3. Should we put `or` combinator for methods to allow defining endpoint that accept more than one method.

What would be the semantic of the client-side? Arbitrarily pick one method? I’m not sure it would be useful because you usually want to pick one method or another for a good reason. What would be your use case? Maybe we can provide such a combinator for the server-side only, though…

julienrf:
> Also, I would like to rename `jsonRequest` to `jsonEntity` or `jsonBody` and I'm curious about your opinion on this.

As I said on gitter, using just `entity` or `body` would be ambiguous with HTTP responses descriptions. I think the right word would be `jsonRequestEntity` but that’s a bit too long :(

Or, maybe we could find a way to make the user facing DSL accept things like:

~~~ scala
endpoint(post(path / "foo", json[Foo]), json[Bar])
// or, explicitly:
endpoint(post(path / "foo", requestEntity = json[Foo]), responseEntity = json[Bar])
~~~

Note that we would use the *same* `json` word for *both* the request and the response. We could probably make this work with some type level trick, but the type signature of `post` would then look like (I’m omitting the request headers for simplicity):

~~~ scala
def post[A, B](
  url: Url[A],
  requestEntity: Entity[A]
)(implicit
  actualRequestEntity: RequestEntity[Entity[A]],
  tupler: Tupler[A, B]
): Request[tupler.Out]
~~~

I’m not sure yet whether it is desirable to add this complexity to the user facing DSL (I would be happy to get rid of tuplers, by the way!).

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
Krever: > IMHO, this is not really a problem because in such a case at some point the compiler will complain that some method is not implemented.

In normal code when you're using and implementing a thing at the same moment it's ok. But when you can implement something in module X and end up with error few days later in module Y. 

> On the other hand, if you always put the override keyword you can unintentionally override a method that was already implemented.

Just curious, could you tell more about such case? If I want to implement some abstract method how can I miss the fact that it is already implemented somewhere? Do you mean some complicated inheritance hierarchy? 

Other thing is that IMHO presence of `override` increase readability, so you know that signature is defined somewhere else and you cannot just change it. 

I think both position have some advantages and disadvantages, and it will be hard to find consensus. If it's important to you just tell me and I will gladly change it without further discussion :)

algebra/src/main/scala/endpoints/EndpointAlg.scala
Krever: I thought compiler won't accept `Tupler[Tupler[A,Unit]#Out, C]#Out` as `Tupler[A,C]#Out` but know I see I was wrong and should've checked it. Thanks

Krever:
> We can add them afterwards, no hurry.
Ok, so I will leave it for now.

> What would be the semantic of the client-side? Arbitrarily pick one method? I’m not sure it would be useful because you usually want to pick one method or another for a good reason. What would be your use case? Maybe we can provide such a combinator for the server-side only, though…

I don't have the use case, and this one of the reason I didn't go with the implementation. I thought it may be useful but let's wait for someone with the use case for this :)

> As I said on gitter, using just entity or body would be ambiguous with HTTP responses descriptions. I think the right word would be jsonRequestEntity but that’s a bit too long :(
Ok, it's 100% legitimate answer. I will leave it for now and maybe try to achieve `json[Foo]` semantics in some future. It may be pretty good feature, because in the end, converting data to/from json is exactly the same no matter if its request or response.

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
julienrf: > In normal code when you're using and implementing a thing at the same moment it's ok. But when you can implement something in module X and end up with error few days later in module Y it is not so nice.

Yes, I totally understand, and that’s why I think it would be useful to have an `@implement` annotation that would give you the error directly in `X`.

> Just curious, could you tell more about such case?

It can be simply because at the beginning of your project class `B` implements a `foo` abstract method of a class `A`, but then after some time this `foo` method is finally implemented directly in `A`. If you used `override` in `B` then you will not notice that there is no need anymore to override the method.

You can also have issues in more complex cases like this one:

~~~ scala
trait Base { def foo: Int }
trait Impl1 extends Base { def foo = 1 }
trait Impl2 extends Base { def foo = 2 }
object Impl extends Base with Impl1 with Impl2 // Does not compile: you have to disambiguate which `foo` you want to use
~~~

But:

~~~ scala
trait Base { def foo: Int }
trait Impl1 extends Base { override def foo = 1 }
trait Impl2 extends Base { override def foo = 2 }
object Impl extends Base with Impl1 with Impl2 // Compiles fine: Impl2 wins
~~~

Krever: Yep, the `@implements` would be great. Curious if someone has already done that. It should be possible with macro, I think.

Now I understand your worries, I will remove the `override`s :)

julienrf:
> > (I would be happy to get rid of tuplers, by the way!).
> 
> Why? And is there any way to do this?

Because they make the type signatures of the user-facing API awful :( But, unfortunately I didn’t find any better solution.

play-client/src/main/scala/endpoints/EndpointPlayClient.scala
```diff
@@ -28,21 +28,20 @@ class EndpointPlayClient(wsClient: WSClient)(implicit ec: ExecutionContext) exte
   /**
     * A function that, given an `A` information and a `WSRequest`, eventually returns a `WSResponse`
     */
-  type RequestEntity[A] = (A, WSRequest) => Future[WSResponse]
+  type RequestEntity[A] = (A, WSRequest) => WSRequest
+
+  override val emptyRequestEntity: RequestEntity[Unit] = { case (_, req) => req }
```
julienrf: Oops, I didn’t catch this one: it should be `lazy val` instead of `override val`.

Krever: I didn't know. What are the implications of it being not lazy `val`? I always assumed that `lazy` is used if we take something from context that may not be initialized, but here we don't take anything from the outside.

Second thing, should I make PR or you can fix it?

julienrf: I would be happy if you make another PR (and remove the remaining `override` keywords at the same time). But if you don’t have time I will do it some day :)

If it’s not a lazy val we can get initialization issues:

~~~ scala
trait SomeEndpoint extends EndpointAlg {
  val foo = endpoint(post(path, emptyRequestEntity), emptyResponse)
}
~~~

Here we are *using* `emptyRequestEntity` while it is still abstract.

Then, when we mix an interpreter:

~~~ scala
object SomeEndpointRouter extends SomeEndpoint with EndpointPlayRouting {
  // …
}
~~~

The initialization semantic in Scala will first run the initialization code of `SomeEndpoint` and then the initialization code of `EndpointPlayRouting`, so we will refer to `emptyRequestEntity` before it will actually be initialized. (I’m just thinking that we could probably prevent this kind of mistakes by making all `Alg` traits *universal traits*).

Also, do you mind renaming `emptyRequestEntity` to `emptyRequest`, for symmetry with `emptyResponse`?

Krever: Sure :) Thanks for explanation.

Issue #29, opened 2017-01-12
User friendliness of algerbas and interpreters composition


Krever:
Currently,  we have a lot of different traits in different modules that can be used to create algebra/server/client and for someone to know what he can use he need to basically browse through source of the library. I think it would be good to address this problem. 
What we have now:
```scala
import endpoints._
trait ApiAlg extends EndpointAlg with CirceCodecAlg with AssetAlg
  with OptionalResponseAlg with BasicAuthenticationAlg 

object Api extends ApiAlg with EndpointPlayRouting with CirceCodecPlayRouting with AssetPlayRouting
  with OptionalResponsePlayRouting with BasicAuthenticationPlayRouting

object Api extends ApiAlg with EndpointXhrClient with CirceCodecXhrClient with AssetXhrClient
  with XhrClientThenable with OptionalResponseXhrClient with BasicAuthenticationXhrClient
```
So to create this, user need to know at least 16 different traits that are not documented anywhere(so need to look at them in the source code, where he will find a lot of supporting traits(UrlAlg, MethodAlg, etc) that are not separated in any way;

I think there are few things we can do:

1. Organise traits in feature-orientend package naming way:
```scala
import endpoints._
trait ApiAlg extends endpoints.Algebra with assets.Algebra
  with optinalresposne.Algebra with authentication.basic.Algebra with marshalling.Circe

object Api extends ApiAlg with routing.Play with routing.play.Circe with routing.play.Assets
  with routing.play.OptionalResponse with routing.play.BasicAuthentication

object Api extends ApiAlg with client.Xhr with client.xhr.Circe with client.xhr.Assets
  with client.xhr.Thenable with client.xhr.OptionalResponse with client.xhr.BasicAuthentication
```
It allows to more intuitive browsing of possibilities, increase readability and divide traits between more descriptive packages. All non user-facing traits should probably go to e.g. `client.xhr.Internal` or something like that.

2. Document the choices:

We should probably do it somehow, probably some list of all available Algebras with their implementations. But I have completely no idea how such list should look like, because e.g `JsonEntityAlg` is partially implemented in `CirceCodecAlg` and partially in external modules. 

3. Encourage composition instead of inheritance

I don't know if it is even possible, but if we define `ApiAlg` like this:
 ```scala
import endpoints._
class ApiAlg(
  endpointAlg: EndpointAlg,
  circeCodecAlg: CirceCodecAlg,
  assetsAlg: AssetAlg,
[...]

Issue #7, opened 2016-04-25
Check that the code given in the README compiles


julienrf:
Consider using [tut](http://github.com/tpolecat/tut) for that.

PR #31, opened 2017-01-16
More user-firendly code organisation

Krever:
Addresses #29 .

I had some doubts during the implementation:
1 . Using plural in `JsonEntites`(instead of `JsonEntity`)
2. `play.routing.circe.JsonEntities` - Circe implementation of `JsonEntities`. I had to make `circe` a package to keep the same class name as algebra.
3. `algebra.CirceEntities` - I had to rename `CirceCodecAlg` and I've chosen `CirceEntities` because of relation to `JsonEntities`.
4. endpoints.PlayCirce - it is neither algebra nor interpreter. I left it without a change, but it should probably live in some other package than root.
5. `endpoints.algebra.BasicAuthentication.Credentials`, `endpoints.algebra.CirceEntities.CirceCodec` - I've put this utility classes in companion objects of algebras, to keep the namespace clean, but I'm not sure if it is good thing. Alternative would be to put them in `util` package or something like that(e.g. `endpoints.algebra.util.CirceCodec`) but that is even worse.

I tried to read the changes carefully, but it is rather big one, so I may have missed something(especially without automated #7 )


play-server-circe/src/main/scala/endpoints/play/routing/CirceEntities.scala
```diff
@@ -1,14 +1,17 @@
...
 
+import endpoints.algebra
+import endpoints.algebra.CirceEntities.CirceCodec
+import endpoints.play
 import io.circe.jawn
-import play.api.libs.concurrent.Execution.Implicits.defaultContext
-import play.api.mvc.{BodyParsers, Results}
-import PlayCirce.circeJsonWriteable
+import _root_.play.api.libs.concurrent.Execution.Implicits.defaultContext
+import _root_.play.api.mvc.{BodyParsers, Results}
```
julienrf: Would it be possible to avoid importing `endpoints.play`, above, so that we don’t need to use the `_root_` package name?

Krever: Fixed

julienrf:
Awesome, thanks!

Maybe we should rename `play.routing` into a single `playrouting` package (and same for `play.client`) to avoid the name clash with the existing `play` package). But we could do that later, if necessary…

PR #32, opened 2017-01-20
Add a more complex example

julienrf:
This PR started with the idea of writing a more realistic example such as the one we can see for the [lagom framework](https://github.com/lagom/online-auction-scala), in order to base the documentation on it.

This example gave me the opportunity to add some more features to the library and I think that would be cool to merge them even though the example is not finished.

This PR introduces:
 - multiplexed endpoints (based on [OlivierBlanvillain/monixwire](https://github.com/OlivierBlanvillain/monixwire)) ;
 - optional querystring parameters ;
 - an `example-cqrs` project.


julienrf:
The diff is huge… I’m sorry about that :-/ That’s also a motivation to merge before making it even bigger.

Issue #10, opened 2016-04-25
Rework documentation


julienrf:
Move the documentation out of the `README`.

Outline:
- Quick start
  - canonical example
  - … ?
- Key features & Motivation
  - first-class values ;
  - separation of definition and evaluation ;
  - extensibility ;
- Reference, by module
  - Endpoints ;
  - Assets ;
  - JSON

PR #33, opened 2017-01-22
[WIP] #10 Bootstrap documentation

### LINKED ISSUE #7: Check that the code given in the README compiles

julienrf:
This is a work in progress, feel free to give feedback.


Krever:
Great piece of work! Single comment I have is that build.sbt is getting really big, but there is probably nothing we can do about that.

julienrf:
> build.sbt is getting really big

Yeah :(

I don’t know how to make it simpler. One option could be to make examples separate projects. But I’d like to do that later and keep the build as it is for now.

Krever:
Another option may be to move configuration for specific projects to that projects, but it has some drawbacks(one have to use `LocalProject("name-of-the-project")` for dependencies, common settings have to be defined in `Build.scala`, Intellij doesn't support such configuration very well and corss projects has to be defined on higher level anyway). I've been using this approach for https://github.com/Krever/scala-startup-template (little side project that is currently suspended until endpoints akka-http backend will be finished) but I agree we should keep the build as it is for now.

julienrf:
I merged it because I think it represents a great improvement over the current state of the documentation. I will continue to work on it over time.

Issue #15, opened 2016-06-07
Implement a lightweight routing semantics


julienrf:
Play is heavy.

PR #35, opened 2017-01-23
Support for akka-http routing

Krever:
It solves #23 and #15 , I think.


akka-http-server/src/main/scala/endpoints/akkahttp/routing/BasicAuthentication.scala
```diff
@@ -0,0 +1,36 @@
...
+import endpoints.algebra.BasicAuthentication.Credentials
+
+trait BasicAuthentication extends algebra.BasicAuthentication with Endpoints {
+
+  /**
+    * Extracts the credentials from the request headers.
+    * In case of absence of credentials rejects request
+    */
+  private[endpoints] lazy val basicAuthentication: RequestHeaders[Credentials] =
+    Directives.headerValue(extractCredentials)
```
julienrf: How is the request rejected if the header is not present?

That’s probably out of the scope of this PR but we should clearly define what the semantic of this method should be. The Play routing rejects the request with a 401 status code, I think we should be consistent with that.

akka-http-server/src/main/scala/endpoints/akkahttp/routing/Endpoints.scala
```diff
@@ -0,0 +1,119 @@
...
+import akka.http.scaladsl.server.{Directive1, Directives, Route}
+import endpoints.algebra.{Decoder, Encoder, MuxRequest}
+import endpoints.{Tupler, algebra}
+
+import scala.concurrent.Future
+import scala.language.higherKinds
+import scala.util.{Failure, Success}
+
+/**
+  * Interpreter for [[algebra.Endpoints]] that performs routing using Play framework.
```
julienrf: Wrong comment :)

examples/basic/shared/src/main/scala/sample/AssetsAlg.scala
```diff
@@ -0,0 +1,15 @@
+package sample
+
+import endpoints.algebra._
+
+/**
+  * Created by wpitula on 1/23/17.
+  */
```
julienrf: This comment is useless

akka-http-server/src/main/scala/endpoints/akkahttp/routing/BasicAuthentication.scala
julienrf: Ah, OK, I see.

At the beginning the `play.routing` interpreter was working exactly like this but at some point it was useful to introduce a distinction between a request that does not match an endpoint and a request that matches it but is invalid.

So, basically, the result of processing a request should be something like `Option[Either[Invalid, Success]]`, where the `Option` models whether the incoming request matched the endpoint or not, and the `Either` models failures that might have happened when the request was processed.

That’s surely something that we should cleary specify, but for now a request *matches* an endpoint if its method, segments and query string match the endpoint definition. The notion of *matching* defines how we associate an incoming request to an endpoint of the application. Then, after a request has matched an endpoint definition, its processing can *fail* if some expected headers are absent or if the request entity is invalid.

julienrf: I think that defining in which conditions the routing process should try the next endpoint or return a failure is a matter of choice and several options can be valid, we just have to agree on a choice because clients will expect servers to behave consistently. I admit that I probably biased the `play.routing` semantics towards the current behavior of Play framework.
However, I still think that in the case of Basic authentication it makes sense to return a 401 status if the header is absent. Web browsers rely on this status and prompt the user for a login and a password. If you don’t return the 401 status then users can not supply their credentials if they use the API from their browser. You can try to run `+example-basic-server/run` and then open http://localhost:9000/auth in your browser to see what I mean.
So, my point is that we should not completely ignore web standards and always use 200 as a status code, for instance. We should take existing HTTP clients into account.

Krever: Sure, with basic auth it makes sense. I will fix it and document in `algebra.BasicAuthentication` and we can discuss general rules for headers and entities outside of this PR, ok? I know it may introduce inconsistency between interpreters, but it's probably not so big problem right now and without automatic testing it will be very difficult to make sure that they are fully consistent.

julienrf:
This is great @Krever, thanks!

PR #44, opened 2017-06-02
Use implicit syntax instead of inheriting from it

julienrf:
I’m not a big fan of implicit conversions. Therefore, the current design puts lower bounds to abstract type members:

~~~ scala
type Url[A] <: UrlOps[A]
~~~

However, in some cases this is very inconvenient:

~~~ scala
type Url[A] = String // does not compile because String is not a subtype of UrlOps[A]
~~~

This PR uses implicit conversions instead.


Krever:
Yep, this is something I've also struggled with :)

Issue #45, opened 2017-06-05
Provide non-play jvm client interpreter


Krever:
Currently we have only one jvm client interpreter, backed by Play Framework. There are two main problems with it:
- Play is still not avaiable for scala 2.12 and already not available for scala 2.10
- It is quite [hard](https://www.playframework.com/documentation/2.3.x/ScalaWS#Using-WSClient) to use it outside of play application

We may consider implementing another intepreter, backed with [scalaj/scalaj-http](https://github.com/scalaj/scalaj-http) or [hmil/RosHTTP](https://github.com/hmil/RosHTTP) (this one will give use scala.js interpreter for free as well).

I'm not 100% sure of this because we already have quite big number of interpreters, and without some generic automated tests the maintenance may become a lot of pain in the future.

julienrf:
I think an akka-http client interpreter would make sense. People using an Akka based stack will be happy to have it.

Krever:
Fixed by #46

PR #46, opened 2017-06-06
Scalaj interpreter + test suite

Krever:
So this quite big. I've made following changes:

1. Added new interpreter that gives us scalaj-http based client. I've chosen scalaj, because it is very simple and published for all scala versions
2. Added new method `stringResponse: Response[String]` to give at least some way to use endpoints without circe(I needed it for tests)
3. Added new project `testsuite` which is supposed to be shared test suite for all intepreters. I've integrated it only with this new scalaj-client and it contains only few basic tests. All of these is because my lack of time. We should probably improve it over time and integrate with other interpreters.
4. Added crosspublishing to scala 2.10 for `algebra` and `algebra-circe`
5. Splitted `algebra.Endpoints` into `algebra.Requests` and `algebra.Responses` for simplification
6. Removed `publishLocal := ()`, because I see no harm in it and it helps with intellij import

This addresses #45


algebra/src/main/scala/endpoints/algebra/Requests.scala
```diff
@@ -0,0 +1,66 @@
...
+    * @param url Request URL
+    * @param entity Request entity
+    * @param headers Request headers
+    */
+  def request[A, B, C, AB](
+                            method: Method,
+                            url: Url[A],
+                            entity: RequestEntity[B] = emptyRequest,
+                            headers: RequestHeaders[C] = emptyHeaders
+                          )(implicit tuplerAB: Tupler.Aux[A, B, AB], tuplerABC: Tupler[AB, C]): Request[tuplerABC.Out]
```
julienrf: Not extremely important but you’ve lost formatting. I prefer to indent parameters with two spaces only.

algebra/src/main/scala/endpoints/algebra/Responses.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+    * Empty response.
+    */
+  def emptyResponse: Response[Unit]
+
+  /**
+    * String response.
+    */
+  def stringResponse: Response[String]
```
julienrf: I would name it `textResponse` to emphasize the fact that the generated response will have content type `plain/text`.

build.sbt
```diff
@@ -27,6 +27,11 @@ val `scala2.12` = Seq(
   crossScalaVersions := Seq("2.11.8", "2.12.1")
 )
 
+val `scala2.12_full` = Seq(
```
julienrf: We should probably adopt a different naming convention. What about `scala 2.11`, `scala 2.11 to 2.12` and `scala 2.10 to 2.12`?

scalaj-client-circe/src/test/scala/endpoints/scalaj/client/CirceEntitiesTest.scala
```diff
@@ -0,0 +1,30 @@
...
+import endpoints.testsuite.client.JsonTestSuite
+import org.scalatest.WordSpec
+
+class TestClient(val address: String ) extends JsonTestApi with Endpoints with CirceEntities {
+
+  import io.circe.generic.auto._
+
+  override implicit def userCodec: CirceCodec[User] = CirceCodec.fromEncoderAndDecoder[User]
+
+  override implicit def addresCodec: CirceCodec[Address] = CirceCodec.fromEncoderAndDecoder[Address]
```
julienrf: I prefer not to put the `override` keyword excepted when we really override a member.

scalaj-client/src/main/scala/endpoints/scalaj/client/Endpoints.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  case class Endpoint[Req, Resp](request: Request[Req], response: Response[Resp]) {
+
+    /**
+      * This method just wraps a call in a Future and is not real async call
+      */
+    def callAsync(args: Req)(implicit ec: ExecutionContext): Future[Resp] =
+      Future {
+        callUnsafe(args)
+      }
```
julienrf: What about also using `concurrent.blocking` here?

testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala
```diff
@@ -0,0 +1,18 @@
+package endpoints.testsuite
+
+import endpoints.algebra
+
+/**
+  * Created by wpitula on 6/5/17.
+  */
```
julienrf: I usually remove these intellij generated comments :)

algebra/src/main/scala/endpoints/algebra/Requests.scala
ShaneDelmore: If you're up for it a checking in a .scalafmt file would be nice so contributors could all use the same format without having to think about it.

julienrf: I’m a bit skeptical about this idea: I don’t have formal formatting rules (ie I have lots of exceptions).

Krever: IMHO having one imperfect but automatically consistent formatting is better than relying of discipline of developers. Anyway I will fix that, no problem, but this is default formatting in Intellij and I cannot promise it won't happen again

scalaj-client/src/main/scala/endpoints/scalaj/client/Endpoints.scala
Krever: Could you elaborate a little? I think I have never used that.

testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala
Krever: So do i, unless I forget, thanks :)

scalaj-client-circe/src/test/scala/endpoints/scalaj/client/CirceEntitiesTest.scala
Krever: Yeah, we have discussed that once in the past, and I even remembered at some point to remove it(it is automatically add by intellij) but then I forgot. Fixed

build.sbt
Krever: Agree, fixed

algebra/src/main/scala/endpoints/algebra/Responses.scala
Krever: Good point, fixed.

scalaj-client/src/main/scala/endpoints/scalaj/client/Endpoints.scala
julienrf: `blocking` “tells” the execution context that this block of code may take several seconds to complete. Then, the execution context can take advantage of this information to manage its thread pool better.

~~~ scala
Future {
  concurrent.blocking {
    callUnsafe(args)
  }
}
~~~

Krever: Nice to learn something new, thanks. Implemented.

play-client/src/main/scala/endpoints/play/client/Endpoints.scala
```diff
@@ -57,6 +57,9 @@ class Endpoints(host: String, wsClient: WSClient)(implicit ec: ExecutionContext)
   /** Successfully decodes no information from a response */
   val emptyResponse: Response[Unit] = _ => Right(())
 
+  /** Successfully decodes no information from a response */
```
julienrf: The Scaladoc is not correct

Issue #42, opened 2017-03-19
Improve modelling of responses in endpoint algebra


afiore:
The `endpoints.algebra` package currently models the concept of response entity, but provides no way to express two other key aspects of the response, both of them relevant to #11.

- Status codes
- Response Headers

Consider the following endpoint:

```scala
val parseInt =
   endpoint(
      post(path / "parseInt", jsonRequest[String]),
      jsonResponse[Either[String, Float]])
```
One would want to express that, for the left case (i.e. the string cannot be parsed into an integer), the server should return a 400 Bad request. While the current implementation allows defining a Circe Encoder for `Either[String, Float]`, in this case, the response status code will still be set to 200 for all the possible request entities.

I reckon that coming up with a good way to handle status codes is not trivial, as ideally, you would want the interface algebra to drive a number of different concerns: i.e. serialisation, server -client interactions, API documentation, etc. ...all of this while reducing boilerplate and keeping a high degree of flexibility!

I am still getting my head around the library and playing around with it, so I have no clear idea on how to tackle the issues above. But it sounds like an interesting challenge!

Krever:
This certainly an interesting problem that should be adressed. 
I think that request headers are relatively easy to implement and may look for example like that:
```
val parseInt =
   endpoint(
      post(path / "parseInt", jsonRequest[String]),
      jsonResponse[Either[String, Float]]
          .withHeader[String]("foo") 
)
```

The status codes are bigger challenge as they target different responses. 
Only thing I can think of right now is something like that:
```
val parseInt =
   endpoint(
      post(path / "parseInt", jsonRequest[String]),
      jsonResponse[Float]
          .withStatusCode(200)
          .or
          .emptyResponse
          .withStatusCode(400) 
)
```
But hat may be a little tricky to implement. 

Anyway I think it is a good point for further study. IMHO we shouldn't take anything for granted, because current model is far not perfect, e.g. with current model basic authentication require special case of endpoint(`authorizedEndpoint` instead of `endpoint`) and this is not very composable.

julienrf:
I agree that we should include response header descriptions.

For the responses’ status codes, that’s already more or less the case:

~~~ scala
endpoint(get(path / "foo"), htmlResponse) // means that we always return a 200 status code
endpoint(get(path / "foo"), option(htmlResponse)) // `option` means that the server could return 404 or the nested response
[...]

PR #47, opened 2017-06-08
Builders API

Krever:
This is a try to introduce a builder api, that will allow more readable endpoint definitions. This may be also a first step to address(at least partially) #42. 

Things worth mentioning:
- I have added ability to join request headers. This allowed me to create more extensible definition of basic authentication
- I have added builders for most of the algebras
- I have added support for OptionalResponses for scalaj interpreter
- I have added a controversial implicit conversion from `Path[A]` to `Url[A]`. I'm not 100% sure about that, but motivation for this was to get rid of intellij error. 

The readability of new api can be easily compared in testsuites.

Also, I'm not 100% about naming(of methods and traits) so any sugestions are very welcome.

### DIFF
--- a/algebra/src/main/scala/endpoints/algebra/Builders.scala
+++ b/algebra/src/main/scala/endpoints/algebra/Builders.scala
@@ -6,7 +6,7 @@ import endpoints.algebra.BasicAuthentication.Credentials
 trait Builders {
   self: Endpoints =>
 
-  def anEndpoint: EndpointBuilder[Unit, Unit, Unit, Unit] = EndpointBuilder(Get, path, emptyHeaders, emptyRequest, emptyResponse)
+  val anEndpoint: EndpointBuilder[Unit, Unit, Unit, Unit] = EndpointBuilder(Get, path, emptyHeaders, emptyRequest, emptyResponse)
 
   case class EndpointBuilder[U, ReqH, ReqE, RespE]
   (method: Method,
@@ -66,7 +66,7 @@ trait JsonBuilders extends Builders {
 trait BasicAuthBuilders extends Builders {
   self: BasicAuthentication =>
 
-  implicit class EndpointJsonOps[U, ReqH, ReqE, RespE](endp: EndpointBuilder[U, ReqH, ReqE, RespE]) {
+  implicit class EndpointBasicAuthOps[U, ReqH, ReqE, RespE](endp: EndpointBuilder[U, ReqH, ReqE, RespE]) {
 
     def withBasicAuth(implicit tupler: Tupler[ReqH, Credentials]): EndpointBuilder[U, tupler.Out, ReqE, Option[RespE]] = {
       val newResponse = authenticated(endp.response)
@@ -82,7 +82,7 @@ trait BasicAuthBuilders extends Builders {
 trait OptionalResponseBuilders extends Builders {
   self: OptionalResponses =>
 
-  implicit class EndpointJsonOps[U, ReqH, ReqE, RespE](endp: EndpointBuilder[U, ReqH, ReqE, RespE]) {
+  implicit class EndpointOptionalResponseOps[U, ReqH, ReqE, RespE](endp: EndpointBuilder[U, ReqH, ReqE, RespE]) {
 
     def withOptionalResponse[NewResp](resp: Response[NewResp]): EndpointBuilder[U, ReqH, ReqE, Option[NewResp]] = {
       val newResponse = option(resp)
--- a/algebra/src/main/scala/endpoints/algebra/Builders.scala
+++ b/algebra/src/main/scala/endpoints/algebra/Builders.scala
@@ -1,0 +1,101 @@
+package endpoints.algebra
+
+import endpoints.Tupler
+import endpoints.algebra.BasicAuthentication.Credentials
+
+trait Builders {
+  self: Endpoints =>
+
+  def anEndpoint: EndpointBuilder[Unit, Unit, Unit, Unit] = EndpointBuilder(Get, path, emptyHeaders, emptyRequest, emptyResponse)
+
+  case class EndpointBuilder[U, ReqH, ReqE, RespE]
+  (method: Method,
+   url: Url[U],
+   requestHeaders: RequestHeaders[ReqH],
+   requestEntity: RequestEntity[ReqE],
+   response: Response[RespE]
+  ) {
+    def withMethod(m: Method): EndpointBuilder[U, ReqH, ReqE, RespE] =
+      this.copy(method = m)
+
+    def withUrl[NewU](u: Url[NewU]): EndpointBuilder[NewU, ReqH, ReqE, RespE] =
+      this.copy(url = u)
+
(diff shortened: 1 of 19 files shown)


algebra/src/main/scala/endpoints/algebra/Urls.scala
```diff
@@ -97,6 +97,9 @@ trait Urls {
   /** An URL path carrying an `A` information */
   type Path[A] <: Url[A]
 
+  // This should remove false positive error in Intellij when using path as an url
+  implicit def pathToUrl[A](path: Path[A]): Url[A] = path
```
julienrf: I prefer this problem to be fixed in IntelliJ :)

Krever: Yes, this would be the best solution, but I don't know if that's gonna happen anytime soon, and this error may be really confusing for the users.

Krever:
> Is the final goal to have only the builder based API?

Not really, because it is not extensible enough, as you have said. My idea was to make it default but optional, e.g. by using builders in most of the docs, and have separate page about underlying API.

algebra/src/main/scala/endpoints/algebra/Builders.scala
```diff
@@ -0,0 +1,101 @@
+package endpoints.algebra
+
+import endpoints.Tupler
+import endpoints.algebra.BasicAuthentication.Credentials
+
+trait Builders {
+  self: Endpoints =>
+
+  def anEndpoint: EndpointBuilder[Unit, Unit, Unit, Unit] = EndpointBuilder(Get, path, emptyHeaders, emptyRequest, emptyResponse)
```
julienrf: Why not a `val`?

algebra/src/main/scala/endpoints/algebra/Builders.scala
```diff
@@ -0,0 +1,101 @@
...
+      endp.copy(response = self.jsonResponse[NewRespE])
+
+  }
+
+}
+
+trait BasicAuthBuilders extends Builders {
+  self: BasicAuthentication =>
+
+  implicit class EndpointJsonOps[U, ReqH, ReqE, RespE](endp: EndpointBuilder[U, ReqH, ReqE, RespE]) {
```
julienrf: Typo: should be named `EndpointBasicAuthOps`

testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala
```diff
@@ -12,4 +13,13 @@ trait BasicAuthTestApi extends algebra.Endpoints with algebra.BasicAuthenticatio
     textResponse
   )
 
+  val protectedEndpointViaBuilder =
+    anEndpoint
+      .withMethod(Get)
+      .withUrl(path / "users")
+      .withTextResponse
+      .withBasicAuth
```
julienrf: If one writes:

~~~ scala
anEndpoint
  .withMethod(Get)
  .withUrl(path / "users")
  .withBasicAuth
  .withTextResponse
~~~

Then the behavior is unexpected: authentication will be handled only for requests, not responses (this is because `.withTextResponse` is called *after* `withBasicAuth`).

We could guard against these kind of incompatible combinations of builder operations by using more refined types, but I’m afraid that would be super complex :(

Krever: Yes, this may cause a lot of trouble when used wrongly.

One solution I see is to make something like this:
```
  case class EndpointBuilder2[Resp, RespMod[_]] {
    def build: Endpoint[???, RespMod[Resp]] = ???
  }
  ```
but this is completely not scalable, because to implement such modifications on all of the elements the number of type arguments grows from 4 to 8(and if we introduce response headers from 5 to 10).

Another approach would be to keep result types in `EndpointBuilder` from the beginning(as type  arguments or type members), like so:
```
  case class EndpointBuilder2[U,RH,RE,Resp, FinalRequest, FinalResponse] {
    def build: Endpoint[FinalRequest, FinalResponse] = ???
  }
```
and the `withResponse` will be awailable when `Resp =:= FinalResponse`

But both of these methods feel more like a hack and I'm not even sure if they will fix anything.

File: algebra/src/main/scala/endpoints/algebra/Builders.scala (hunk 2 above)
Comment (Krever) [inline algebra/src/main/scala/endpoints/algebra/Builders.scala]: Fixed. By the way is there any other reason then performance?

File: algebra/src/main/scala/endpoints/algebra/Builders.scala (hunk 3 above)
Comment (Krever) [inline algebra/src/main/scala/endpoints/algebra/Builders.scala]: Thanks, fixed

File: algebra/src/main/scala/endpoints/algebra/Builders.scala (hunk 2 above)
Comment (julienrf) [inline algebra/src/main/scala/endpoints/algebra/Builders.scala]: No. Actually it’s not that important :)

File: testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala (hunk 4 above)
Comment (ShaneDelmore) [inline testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala]: @julienrf I am confused by your statement about auth happening only for requests, not responses.  Could you show an example of the non-builder syntax of the same bit of code?
Also, isn't authentication only for requests?  I don't think I have heard of an authenticated response.

Comment (julienrf) [inline testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala]: The equivalent without builders would be:

~~~ scala
authenticatedEndpoint(Get, path / "users", response = textResponse)
~~~

“authentication for responses” means that responses are “authentication aware”: they will return a `403` in case of an unauthorized user.

Comment (ShaneDelmore) [inline testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala]: Interesting.  Naively I would have expected the authentication was for the Request => Response instead of one or the other.

Comment (Krever) [inline testsuite/src/main/scala/endpoints/testsuite/BasicAuthTestApi.scala]: @ShaneDelmore Just to clarify, Julien's example would work correctly. To imitate this wrong behaviour, one would have to use `endpoint` instead of `authenticatedEndpoint`, like this:
```
endpoint(request(method, url, requestEntity, basicAuthentication),response)
```
For comparison, `authenticatedEndpoint` is defined like so:
```
endpoint(request(method, url, requestEntity, basicAuthentication), authenticated(response))
```

So, in current API basic auth is safe, because it gives you guarantee that both, response and request, will be authenticated(so, request contains/parse `Authorisation` header and response contains/parse 200/403 status code. With the new API this guarantee is gone, because user can override part of the mechanism. 

I'm experimenting with ways of fixing it, but all the tries either removes composability of builders or complicates the API in extraordinary way.

julienrf: @Krever This is the kind of things we can do to get safe builders: https://fr.slideshare.net/greheine/a-typesafe-builder-a-gentle-introduction-to-type-level-programming

Krever: @julienrf Looks promising. I will try to experiment a little with this approach, thanks.

Krever:
I will close this one as I'm going to give it another shot from scratch soon.

Issue #11, opened 2016-04-25
Add a semantics that produces a documentation of the endpoints


julienrf:
See e.g. [OpenAPI](https://github.com/OAI/OpenAPI-Specification).

PR #49, opened 2017-06-13
Generate an OpenAPI descriptor

julienrf:
Addresses #11.

This PR introduces a new algebra and its interpreter that produces an OpenAPI descriptor from a set of endpoint descriptions.

We have to introduce a new algebra because the current one does not contain enough information to produce a sensible documentation. For instance, in the current algebra, responses have no associated description (which is required to produce an OpenAPI documentation of an endpoint).

The drawback is that most of the time we can not reuse the existing algebras because the methods don’t have exactly the same signatures.

An alternative design would be to reuse exactly the same algebra interfaces but to extend them with new operations that enrich some “things” with documentation:

~~~ scala
val foo = endpoint(get(path / "foo"), textResponse.withDocumentation("A “foo” thing"))
~~~

Here, the `withDocumentation` method would be implicitly provided on `Response[A]` values.

The drawback of this approach would be that endpoints would not be documented *by construction* (users would have to think of calling `withDocumentation`), thus generating the documentation could fail at runtime…


Krever:
Wow, nice to finally see that coming! :+1: 

Here are some questions that arise in my head:
1. I assume it is work in progress, right?
2. Is the new algebra compatible with `algebra.Endpoints` ? I think that it is a must-have to be able to instantiate a server and client based on `DocumentedApi` 
3. Will it require response headers? 
4. I see no lib dependency on open-api, so I assume we are creating docs from sratch base on format documentation, right?
5. How much is still to be done to have it fully functional?

julienrf:
> I assume it is work in progress, right?

Yes, the algebra is not complete yet. I wanted to gather some feedback on the approach before continuing.

> Is the new algebra compatible with `algebra.Endpoints` ? I think that it is a must-have to be able to instantiate a server and client based on `DocumentedApi`

Yes, it is compatible in a sense because you can always get an `algebra.Endpoints` instance from a `algebra.DocumentedEndpoints` instance.

This is illustrated in the example, where I define a [documented endpoint](https://github.com/julienrf/endpoints/blob/53440d4b0294e0cbe22dd90f835afc32e685cc45/examples/basic/shared/src/main/scala/sample/algebra/DocumentedApi.scala#L5-L9) and then I get an [OpenAPI interpretation](https://github.com/julienrf/endpoints/blob/53440d4b0294e0cbe22dd90f835afc32e685cc45/examples/basic/play-server/src/main/scala/sample/openapi/DocumentedApi.scala#L9-L42) as well as a [Play server interpretation](https://github.com/julienrf/endpoints/blob/53440d4b0294e0cbe22dd90f835afc32e685cc45/examples/basic/play-server/src/main/scala/sample/play/server/DocumentedApi.scala#L3-L15) (we could also get a Scala.js client as well). The `delegate` thing is the indirection that turns a `DocumentedEndpoints` into an `Endpoints` (see also [here](https://github.com/julienrf/endpoints/blob/53440d4b0294e0cbe22dd90f835afc32e685cc45/openapi/src/main/scala/endpoints/delegate/DocumentedEndpoints.scala#L14) for more context).

> Will it require response headers?

What do you mean by ”require”? Eventually, all the features supported by the original `algebra.Endpoints` interface should be supported by being mirrored in the `DocumentedEndpoints` version.

> How much is still to be done to have it fully functional?

The next huge task will be to be able to generate JSON schemas of JSON entities carried by requests and responses. If I delay this task for a later time, then very few things remain to support all the features supported the original `algebra.Endpoints` (e.g. documenting headers, query string parameters, etc.)

> Would it be possible, to have some "dummy defaults" in Documented API, if someone don't target proper docs? This way we could potentially avoid having two distinct algebras to maintain.

Yes, maybe. I created a new algebra just because the original one might not provide sensible defaults in its current state. And I didn’t want to change the original algebra because of the OpenAPI interpreter because users who don’t care about documentation would have to supply information that they don’t care about.

For instance, in the original algebra, one can define a URL segment with the `segment` method:

~~~ scala
val url = path / "user" / segment[Long]
~~~

In clients and servers, this `segment[Long]` is interpreted as `Long => Url` and `Url => Option[Long]`, respectively, to produce an URL from a `Long` parameter and to extract a `Long` value from an URL segment.

In OpenAPI we have to give a *name* to this segment, for instance, we want to be able to document `url` as follows, in case this `Long` value is an id:

~~~ json
{
  "/user/{id}": { ... }
}
~~~

So, we have to provide a name for this segment. The `segment` method could take this name as a parameter:

~~~ scala
val url = path / "user" / segment[Long]("id")
~~~

But what about who are not interested in using the OpenAPI interpreter? If we change the original algebra to carry also documentation related information, these users will have to supply this information even though they don’t care about it.

We could make things optional and try to provide sensible defaults, as you suggested, so that users would not have to always supply this information. In this example, the best default we could provide would be to auto-generate a fresh identifier for each segment parameter (e.g. `segment1`, `segment2`, etc.).

Now, can we *always* provide default values? So far, yes: there are only two places where more information is needed: [path segments](https://github.com/julienrf/endpoints/blob/53440d4b0294e0cbe22dd90f835afc32e685cc45/openapi/src/main/scala/endpoints/algebra/DocumentedUrls.scala#L48) and [response descriptions](https://github.com/julienrf/endpoints/blob/53440d4b0294e0cbe22dd90f835afc32e685cc45/openapi/src/main/scala/endpoints/algebra/DocumentedEndpoints.scala#L40). I just explained which default values we could generate for path segments. For response descriptions that’s even easier: we could just generate “No entity”, “Text entity” and “JSON entity” for the `emptyResponse`, `textResponse` and `jsonResponse` methods, respectively.

What would be the drawback of unifying the `DocumentedEndpoints` and `Endpoints` algebras? So far I described the minimal changes to apply to get something work, but in practice OpenAPI can contain a lot more information, and our algebra should make it possible to leverage any feature of OpenAPI. It means, for instance, that we should add parameters to our `endpoint` method to get a short summary, a long description, whether it is deprecated or not, etc. (all these information can be shown in the OpenAPI description). I am not sure I want to put all these things in the original API.

Krever:
> What do you mean by ”require”? Eventually, all the features supported by the original algebra.Endpoints interface should be supported by being mirrored in the DocumentedEndpoints version.
What I meant is doeas OpenAPI require response headers documentation. But now I see that probably not. I'm asking because this is something that is missing from our current `algebra.Endpoints`(we are handling response as a whole, without distinguishing headers, entity and status code).

> What would be the drawback of unifying the DocumentedEndpoints and Endpoints algebras? So far I described the minimal changes to apply to get something work, but in practice OpenAPI can contain a lot more information, and our algebra should make it possible to leverage any feature of OpenAPI. It means, for instance, that we should add parameters to our endpoint method to get a short summary, a long description, whether it is deprecated or not, etc. (all these information can be shown in the OpenAPI description). I am not sure I want to put all these things in the original API.

Understood. As long as we can get algebra.Endpoints from algebra.DocumentedApi it is fine. 

I see that you have missed 4th question, but anyway, thanks for answering. This looks really promising and will potentially increase usability of our library by 30% at least :)

julienrf:
Oh indeed, I missed this one:

> I see no lib dependency on open-api, so I assume we are creating docs from sratch base on format documentation, right?

Is there any useful Scala OpenAPI library? I simply created a Scala ADT modeling an OpenAPI specification (see [here](https://github.com/julienrf/endpoints/blob/doc-interpreter/openapi/src/main/scala/endpoints/openapi/OpenApi.scala))

Krever:
>Is there any useful Scala OpenAPI library? I simply created a Scala ADT modeling an OpenAPI specification (see here)

Probably not, there is some java lib I suppose, but anyway, I agree that this approach is good enough. I've always thought that recreating this ADT from scratch will be more complicated :)

ShaneDelmore:
Regarding segments and naming, what do you think of creating new types for most segments?

Instead of 
Val URL = path / "user" / segment[Long] / pet / segment[Long]

Val URL = path / "user" / UserId / pet / PetId
Or
Val URL = path / "user" / segment[UserId] / pet / segment[PetId]

I don't actually find much value in the segment portion myself, it comes between two /'s, doesn't that strongly imply a segment?  Either way, could you think of a way to use the new types to possibly tie documentation with type (in this case just name is a good default) or require the type to have a name provided by construction?

julienrf:
> val URL = path / "user" / UserId / pet / PetId

This one would not work because you use a type in a term position. Alternatively, you could define an intermediate `val userId = segment[UserId]` and then you could use this `userId` in your path definition. But that requires one more line…

> val URL = path / "user" / segment[UserId] / pet / segment[PetId]
> …
> Either way, could you think of a way to use the new types to possibly tie documentation with type (in this case just name is a good default) or require the type to have a name provided by construction?

“require the type to have a name provided by construction” -> No idea on how to do that!

I know several people like the idea of deriving things from types. In my opinion, this approach falls short: in most cases the type alone does not carry enough information to do really useful things.

With values, I can also add a description to my segment:

~~~ scala
val url = path / "user" / segment[Int](name = "userId", description = "Id of the user")
~~~

Also, values are simpler than types to abstract over and combine. If you have several endpoints that use the same “user id” you can just write the following:

~~~ scala
val userId = segment[Long]("userId", "Id of user")
val read = endpoint(get(path / "user" / userId), jsonResponse[User])
val update = endpoint(put(path / "user" / userId, jsonRequest[UpdateUser]), jsonResponse[User])
~~~

I don’t think that having to define an intermediate `val` requires more boilerplate that defining a new type (as in your suggestion).

ShaneDelmore:
I quite like your last example.  I think just declaring the value separately gets rid of all of the ugliness of adding the details directly in the path that initially had me a little wary of the design.

julienrf:
The PR is becoming huge (more than 1k SLOC…) :(

@Krever Does it look good to you? I will add documentation and then I’d like to merge it.

examples/basic/play-server/src/main/scala/sample/Server.scala
```diff
@@ -25,8 +26,13 @@ object Server extends App with Results {
     }
     case GET(p"/assets/sample-client-fastopt.js") =>
       controllers.Assets.versioned("/", "sample-client-fastopt.js")
+    case GET(p"/api/description") => Action {
```
julienrf: This endpoint shows a JSON description of the `DocumentedApi` (defined in the file below). This JSON description follows the OpenAPI 3 specification.

examples/basic/shared/src/main/scala/sample/algebra/DocumentedApi.scala
```diff
@@ -0,0 +1,40 @@
+package sample.algebra
+
+import io.circe.generic.JsonCodec
+
+trait DocumentedApi
+  extends endpoints.algebra.DocumentedEndpoints
+    with endpoints.algebra.DocumentedOptionalResponses
+    with endpoints.algebra.DocumentedBasicAuthentication
+    with endpoints.algebra.DocumentedJsonEntities {
```
julienrf: This is the definition of the documented endpoints. The idea is that the DSL is the same as for non-documented endpoints, but with additional parameters carrying documentation.

examples/basic/shared/src/main/scala/sample/algebra/DocumentedApi.scala
```diff
@@ -0,0 +1,40 @@
...
+    endpoint(
+      get(path / "items" / segment[String]("category") /? optQs[Int]("page")),
+      jsonResponse[List[Item]]("List all the items of the given category")
+    )
+
+  val itemId = segment[String]("id")
+
+  val item =
+    endpoint(
+      get(path / "item" / itemId),
```
julienrf: This path will be rendered as follows in the documentation: `/item/{id}`

examples/basic/shared/src/main/scala/sample/algebra/DocumentedApi.scala
```diff
@@ -0,0 +1,40 @@
...
+    authenticatedEndpoint(
+      Get,
+      path / "admin",
+      emptyRequest,
+      emptyResponse("Administration page"),
+      "Authentication error"
+    )
+
+  implicit def itemDecoder: JsonResponse[Item]
+  implicit def listDecoder[A : JsonResponse]: JsonResponse[List[A]]
```
julienrf: These implicits are temporary and will be gone when I’ll create the `openapi-circe` module.

openapi/src/main/scala/endpoints/algebra/DocumentedEndpoints.scala
```diff
@@ -0,0 +1,56 @@
...
+/**
+  * Algebra interface for describing endpoints including documentation
+  * (such as human readable descriptions of things).
+  *
+  * This interface is modeled after [[Endpoints]] but some methods
+  * take additional parameters carrying the documentation part.
+  */
+trait DocumentedEndpoints
+  extends DocumentedRequests
+    with DocumentedResponses {
```
julienrf: The diff is huge partly because all the algebra interfaces (e.g. `algebra.Endpoints`) are “mirrored” in “documented” algebra interfaces like this one.

openapi/src/main/scala/endpoints/delegate/DocumentedEndpoints.scala
```diff
@@ -0,0 +1,30 @@
...
+  * Interpreter for [[algebra.DocumentedEndpoints]] that ignores information
+  * related to documentation and delegates to another [[algebra.Endpoints]]
+  * interpreter.
+  */
+trait DocumentedEndpoints
+  extends algebra.DocumentedEndpoints
+    with DocumentedRequests
+    with DocumentedResponses {
+
+  val delegate: algebra.Endpoints
```
julienrf: Users can turn a `DocumentedXxx` into an `Xxx` using delegation. This makes it possible to apply the usual interpreters (e.g. `play.server.Endpoints`) to a `DocumentedEndpoints`.

openapi/src/main/scala/endpoints/openapi/DocumentedEndpoints.scala
```diff
@@ -0,0 +1,64 @@
...
+import endpoints.algebra.MuxRequest
+
+/**
+  * Interpreter for [[algebra.DocumentedEndpoints]] that produces
+  * an [[OpenApi]] instance for endpoints.
+  */
+trait DocumentedEndpoints
+  extends algebra.DocumentedEndpoints
+    with DocumentedRequests
+    with DocumentedResponses {
```
julienrf: Finally, this is the interpreter that builds an OpenAPI description of the endpoint descriptions.

examples/basic/shared/src/main/scala/sample/algebra/DocumentedApi.scala
Krever: Have you considered `endpoints.algebra.docs.Endpoints` instead of `DocumentedEndpoints`. I'm not sure which is better, just wondering.

openapi/src/main/scala/endpoints/delegate/DocumentedRequests.scala
```diff
@@ -0,0 +1,34 @@
+package endpoints
+package delegate
```
Krever: Commenting on random delegate: why delegates are contained inside `openapi` project ?

openapi/src/main/scala/endpoints/openapi/DocumentedBasicAuthentication.scala
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+  * Interpreter for [[algebra.BasicAuthentication]] that produces
+  * OpenAPI documentation.
+  */
+trait DocumentedBasicAuthentication
+  extends algebra.DocumentedBasicAuthentication
+    with DocumentedEndpoints {
+
+  private[endpoints] def basicAuthentication: RequestHeaders[Credentials] =
```
Krever: Have i missed this before or is this something new? I mean limiting access in interpreters. I find it very nice, because namespace of client/server is not cluttered.

openapi/src/main/scala/endpoints/openapi/DocumentedJsonEntities.scala
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+  * Interpreter for [[algebra.DocumentedJsonEntities]].
+  */
+trait DocumentedJsonEntities
+  extends algebra.DocumentedJsonEntities
+    with DocumentedEndpoints {
+
+  // We don’t provide (yet!) a useful JSON schema
+  case class JsonSchema[A]()
```
Krever: Do we have a ticket for adding that?

openapi/src/main/scala/endpoints/openapi/DocumentedRequests.scala
```diff
@@ -0,0 +1,41 @@
...
+  case class DocumentedRequest(
+    method: Method,
+    url: DocumentedUrl,
+    headers: DocumentedHeaders,
+    entity: Option[DocumentedRequestEntity]
+  )
+
+  type RequestEntity[A] = Option[DocumentedRequestEntity]
+
+  case class DocumentedRequestEntity(description: Option[String], content: Map[String, MediaType])
```
Krever: Can we have `String` in `content` either aliased or commented(to know what it expresses without looking at usages)?

openapi/src/main/scala/endpoints/openapi/DocumentedUrls.scala
```diff
@@ -0,0 +1,61 @@
...
+  def combineQueryStrings[A, B](first: QueryString[A], second: QueryString[B])(implicit tupler: Tupler[A, B]): QueryString[tupler.Out] =
+    DocumentedQueryString(first.parameters ++ second.parameters)
+
+  def qs[A](name: String)(implicit value: QueryStringParam[A]): QueryString[A] =
+    DocumentedQueryString(List(DocumentedParameter(name, required = true)))
+
+  def optQs[A](name: String)(implicit value: QueryStringParam[A]): QueryString[Option[A]] =
+    DocumentedQueryString(List(DocumentedParameter(name, required = false)))
+
+  type QueryStringParam[A] = Unit
```
Krever: OpenApi doesn't allow to specify what is the type of parameter? Sad...

openapi/src/main/scala/endpoints/openapi/OpenApi.scala
```diff
@@ -0,0 +1,131 @@
...
+  paths: Map[String, PathItem]
+)
+
+object OpenApi {
+
+  implicit val jsonEncoder: ObjectEncoder[OpenApi] =
+    ObjectEncoder.instance { openApi =>
+
+      def jsonMediaTypes(mediaTypes: Map[String, MediaType]): Json =
+        Json.fromFields(mediaTypes.map { case (tpe, mediaType) => tpe -> Json.obj() /* TODO */ })
```
Krever: What is happening here or what is "to do" ?

openapi/src/main/scala/endpoints/openapi/OpenApi.scala
```diff
@@ -0,0 +1,131 @@
...
+                    }.fold(fields)(_ :: fields)
+
+                  verb -> Json.fromFields(fieldsWithRequestEntity)
+                }
+              )
+            (path, itemObj)
+          }
+        )
+      ))
+    }
```
Krever: This looks quite complicated... Are there any problems with expressing this json as case class and then serializing automatically?

examples/basic/shared/src/main/scala/sample/algebra/DocumentedApi.scala
julienrf: That’s a good idea, I’ll try that.

openapi/src/main/scala/endpoints/delegate/DocumentedRequests.scala
julienrf: I put everything in the `openapi` project: the algebra, the `OpenAPI` interpreter and the delegate interpreters. Alternatively, we could have 3 distinct projects because the algebra doesn’t depend on the interpreters, but I thought that getting everything in a single dependency could make things simpler for users. Also, this dependency is quite cheap because it doesn’t depend on an heavy third-party `OpenAPI` library (in contrast to `circe` modules, which pull circe and cats, and thus justify to be in separate projects). Last, a good reason to separate interfaces and interpreters would be if we wanted to have other interpreters than the `OpenAPI` one (e.g. RAML), however the `openapi.DocumentedEndpoints` algebra is quite specific to the OpenAPI specification, so if we want to support alternative formats we would probably just define another algebra interface.

openapi/src/main/scala/endpoints/openapi/DocumentedBasicAuthentication.scala
julienrf: I just copied what is in the `endpoints.algebra.BasicAuthentication` trait. I don’t think it is new.

The problem is that because both interpreters and users of an algebra interface _extend_ this interface, using `protected` members is not useful to hide them from users. Hence the `private[endpoints]` qualifier.

openapi/src/main/scala/endpoints/openapi/DocumentedJsonEntities.scala
julienrf: I just created #51.

openapi/src/main/scala/endpoints/openapi/DocumentedRequests.scala
julienrf: Good idea!

openapi/src/main/scala/endpoints/openapi/DocumentedUrls.scala
julienrf: Yes it does (cf [spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#parameter-object)), but I skipped that for a later iteration. (cf the TODO line 13)

openapi/src/main/scala/endpoints/openapi/OpenApi.scala
julienrf: Here, for each media type supported by the endpoint we can define a corresponding schema or entity example (see the [spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#mediaTypeObject))

openapi/src/main/scala/endpoints/openapi/OpenApi.scala
julienrf: Then the case classes hierarchy will become complicated as well :)

I can break down the method into simpler parts.

julienrf:
Thanks a lot for your review, @Krever, I’ll address your comments ASAP.

openapi/src/main/scala/endpoints/delegate/DocumentedRequests.scala
Krever: Oh, I have missed that both algebra and interpreter is in the same project. Agree, it is a good idea, but I wonder if we need to separate algebra and interpreter in such case. If connected, one would defined documentation in common project and then mix in delegates in client/server(not sure if that is possible to implement, just throwing ideas).

openapi/src/main/scala/endpoints/openapi/OpenApi.scala
Krever: No need, agree that this is more reliable :)

openapi/src/main/scala/endpoints/delegate/DocumentedRequests.scala
julienrf: It is possible to mess things up but I would leave things as they currently are, we’ll see later if we need to break things down into more projects (we already have *a lot* of sub-projects…)

openapi/src/main/scala/endpoints/openapi/OpenApi.scala
Krever: So maybe we can have some short description of what is intended here to be done(inside `TODO` comment)

openapi/src/main/scala/endpoints/openapi/DocumentedUrls.scala
Krever: Maybe a comment, just not to forget?

PR #52, opened 2017-08-13
Add akka-http client

Fristi:
Rough version for akka-http client. Let me know what you think

I see this as todo:

- [x] Tests
- [x] We need to open issues for other clients which also use this `Urls.scala`.. I've copied it from the play-client project and added a few fixes regarding query string encoding (optional params)
- [x] Documentation in the code (not sure what you all need?)
- [x] Update docs in markdown files

Example of a client: https://github.com/Fristi/endpoints-test/blob/master/client/src/main/scala/com.veon/Client.scala

### DIFF
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/Urls.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/Urls.scala
@@ -1,0 +1,87 @@
+package endpoints.akkahttp.client
+
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets.UTF_8
+
+import endpoints.Tupler
+import endpoints.algebra
+
+/**
+  * [[algebra.Urls]] interpreter that builds URLs.
+  */
+trait Urls extends algebra.Urls {
+
+  val utf8Name = UTF_8.name()
+
+  trait QueryString[A] {
+    def encodeQueryString(a: A): Option[String]
+  }
+
+  def combineQueryStrings[A, B](first: QueryString[A], second: QueryString[B])(implicit tupler: Tupler[A, B]): QueryString[tupler.Out] =
+    (ab: tupler.Out) => {
+      val (a, b) = tupler.unapply(ab)
+
+      (first.encodeQueryString(a), second.encodeQueryString(b)) match {
+        case (Some(left), Some(right)) => Some(s"$left&$right")
+        case (Some(left), None) => Some(left)
+        case (None, Some(right)) => Some(right)
+        case (None, None) => None
+      }
+    }
+
+  def qs[A](name: String)(implicit value: QueryStringParam[A]): QueryString[A] =
+    a => Some(s"$name=${value.apply(a)}")
+
+  def optQs[A](name: String)(implicit value: QueryStringParam[A]): QueryString[Option[A]] = {
+    case Some(a) => qs[A](name).encodeQueryString(a)
+    case None => None
+  }
+
+  type QueryStringParam[A] = A => String
+
+  implicit lazy val stringQueryString: QueryStringParam[String] = s => URLEncoder.encode(s, utf8Name)
+
+  implicit lazy val intQueryString: QueryStringParam[Int] = i => i.toString
+
+  implicit lazy val longQueryString: QueryStringParam[Long] = i => i.toString
+
+
+  trait Segment[A] {
+    def encode(a: A): String
+  }
+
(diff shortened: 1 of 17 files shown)


project/EndpointsSettings.scala
```diff
@@ -67,5 +67,6 @@ object EndpointsSettings {
   // --- Common dependencies
 
   val circeVersion = "0.6.1"
-
+  val akkaHttpVersion = "10.0.1"
+  val akkaHttpJsonVersion = "1.17.0"
```
Krever: Could these be defined in `akka-http/build.sbt` ?

akka-http/client/src/main/scala/endpoints/akkahttp/client/Urls.scala
```diff
@@ -0,0 +1,87 @@
...
+    a => Some(s"$name=${value.apply(a)}")
+
+  def optQs[A](name: String)(implicit value: QueryStringParam[A]): QueryString[Option[A]] = {
+    case Some(a) => qs[A](name).encodeQueryString(a)
+    case None => None
+  }
+
+  type QueryStringParam[A] = A => String
+
+  implicit lazy val stringQueryString: QueryStringParam[String] = s => URLEncoder.encode(s, utf8Name)
```
Krever: There is no akka-http mechanism to handle that? I worry about users implementing their own segments and forgetting to properly encode the string, so maybe we can do it sometime later "globally" ?

ShaneDelmore: There are many useful URI and Query types/methods in https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala#L601

project/EndpointsSettings.scala
Fristi: I'll move, no problem!

akka-http/client/src/main/scala/endpoints/akkahttp/client/Urls.scala
Fristi: > There is no akka-http mechanism to handle that? I worry about users implementing their own segments and forgetting to properly encode the string, so maybe we can do it sometime later "globally" ?

For this matter, I would investigate if we can introduce `Contravariant` for encoding algebras and `Functor` for decoding algebras.

Fristi:
Thanks, the library is a great idea!

> One thing, the example of the client usage you provided looks quite complicated. Typically user code around the client is ~1line for other interpreters.

I think we can provide a simple default, but it's that you have people plug in their own if they want to do custom stuff like;

- Mutual TLS
- Circuitbreakers
- Retries
- Max in-flight requests
- etc

julienrf:
> We need to open issues for other clients which also use this Urls.scala.. I've copied it from the play-client project and added a few fixes regarding query string encoding (optional params)

Can you elaborate on that?

> Documentation in the code (not sure what you all need?)

Yeah, that would be great!

More generally I think we should put more Scaladoc on each trait (`Endpoints`, `OptionalResponses`, `Assets`, etc.). In the `algebra` package we should explain how they work from the user point of view. In interpreters we should just give information about the specificity of each interpreter.

Fristi:
- Added tests
- Artifact entries at the docs
- Squashed commits

akka-http/client/src/main/scala/endpoints/akkahttp/client/AkkaHttpClient.scala
```diff
@@ -0,0 +1,126 @@
...
+        Source.single(request -> 1)
+          .via(poolClientFlow)
+          .map(_._1.get)
+          .runWith(Sink.head)
+    }
+}
+
+
+class AkkaHttpClient(val settings: AkkaHttpClientSettings)
+                    (implicit val EC: ExecutionContext, val M: Materializer) extends algebra.Endpoints with Urls with Methods {
```
julienrf: So far the naming convention is to use the same name for an algebra interface and its implementation (but the interface lives in the `algebra` package, whereas the implementation lives in a different package. Would that convention work for you too?

Fristi: Makes sense, will change!

Fristi:
Processed your comments!

Fristi:
There it is!

julienrf:
Thanks!!!

Issue #51, opened 2017-06-29
Data schema description


julienrf:
When remote actors talk to each other, they have to agree on the format used to exchange data. Currently, the strategy to make both clients and servers agree on data formats is to use the same codec (ie. we require an implicit value which is (generally) unique for a given type to serialize). So far, the concrete types of these codecs are Circe’s `Encoder` and `Decoder` or Akka-http’s `Marshaller` and `Unmarshaller`.

We should add a new codec concrete type that describes the structure of the data so that we can generate more complete documentation of endpoints.

ShaneDelmore:
Do you think it would be possible to have server and client negotiate if a schema is compatible instead of using the same code?  For example, if I have v1Codec[User] client and v2Codec[User] server which is identical but with a new optional field, it would be great if these were known to be compatible if it doesn’t add too much complexity.  I have seen shapeless techniques to determine if one class could be transformed to another than understood how to handle extra optional fields example.

julienrf:
Yes, that would be possible. I would leave that to another ticket, though :)

julienrf:
Closed by #53.

PR #53, opened 2017-08-29
Proof of concept of an algebra interface for defining data type schemas

julienrf:
(See related issue #51 for context)

The diff is not that long (about 500 added lines) but I think a small example might illustrate better what this PR enables.

You can now describe a schema for your data types:

~~~ scala
case class User(name: String, age: Int)

object User {
  // This is the corresponding schema: an object with
  // two fields: “name” and “age”
  implicit val jsonSchema: JsonSchema[User] =
    (field[String]("name") :×: field[Int]("age")).as[User]

  // Or, using automatic derivation:
  // implicit val jsonSchema: JsonSchema[User] = genericJsonSchema[User]
}
~~~

This schema is just a description of a schema and several interpreters can be applied to it. I implemented two interpreters, one that generates a JSON schema and one that generates a circe decoder and encoder to do actual marshalling.

Apart from that there are no other changes, you can define your endpoints as previously:

~~~ scala
val getUser = endpoint(get(path / "user" / segment[UUID]), jsonResponse[User])
~~~

The server will generate the JSON response according to the defined schema. And the client will decode the response according to the schema.

Last but not least, the JSON schema can be used by the OpenAPI interpreters to produce richer descriptions of an API. And then, OpenAPI tools can generate nice things like the following:

image

(this is what we get from the example application in the docs)

Also, a nice thing is that this “data type schema” algebra interface is totally independent of _endpoints_. In fact it could have been defined in a different repository.

I really love the fact that _endpoints_ is modular and extensible, especially because I don’t want to make decisions in place of users. I want them to be able to use whatever server they want (Play, Akka-http, etc.), whatever JSON library they want, whatever documentation system they want, etc. But, on the other hand, this extensibility leads to a jungle of traits that’s hard to follow. WDYT?


openapi/build.sbt
```diff
@@ -3,13 +3,49 @@ import LocalCrossProject._
...
+    .dependsOn(`json-schema`)
+
+val `json-schema-generic-js` = `json-schema-generic`.js
+val `json-schema-generic-jvm` = `json-schema-generic`.jvm
+
+val `json-schema-circe` =
+  crossProject.crossType(CrossType.Pure).in(file("json-schema-circe"))
+    .settings(publishSettings ++ `scala 2.10 to 2.12`: _*)
+    .settings(
+      name := "endpoints-openapi-json-schema-generic",
```
Krever: The name duplicates `json-schema-generic`

openapi/json-schema-circe/src/main/scala/endpoints/documented/circe/JsonSchemas.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  implicit def stringJsonSchema: JsonSchema[String] = CirceCodec.fromEncoderAndDecoder(implicitly, implicitly)
+
+  implicit def intJsonSchema: JsonSchema[Int] = CirceCodec.fromEncoderAndDecoder(implicitly, implicitly)
+
+  implicit def longJsonSchema: JsonSchema[Long] = CirceCodec.fromEncoderAndDecoder(implicitly, implicitly)
+
+  implicit def bigdecimalJsonSchema: JsonSchema[BigDecimal] = CirceCodec.fromEncoderAndDecoder(implicitly, implicitly)
+
+  implicit def booleanJsonSchema: JsonSchema[Boolean] = CirceCodec.fromEncoderAndDecoder(implicitly, implicitly)
```
Krever: Do you think providing explicit values here would make sense? Its always clearer and makes compile time lower(don't know how much though)

openapi/openapi/src/main/scala/endpoints/documented/openapi/Requests.scala
```diff
@@ -2,6 +2,8 @@ package endpoints
 package documented
 package openapi
 
+import endpoints.documented.openapi.model.MediaType
```
Krever: Why this import is needed?

Krever:
This looks great! And turned out to be simpler than I expected it to be. I like the fact that it is really self contained without any dependencies. 

About the traits, I agree that extensibility is a very nice feature, and also agree that it complicates life a lot. I believe a typical user doesn't really care about all these interfaces and want something that "just works". Maybe we could define some additional traits on top of the existing ones that will simplify things. 

In one of my projects we have defined 
```
trait AkkaHttpServer extends server.Endpoints with server.CirceEntities with server.OptionalResponses
```
to increase readability. Maybe we can use similar technique to connect all of the traits available in algebra, and than all of the interpreting traits in given interpreter. So the user doesn't have to deal with the complexity of mixin traits and just use sth like `AkkaHttpCirceBasedServer` or  `AkkaHttpDocumentedCirceBasedServer`. One problem is that I'm not sure where to put this traits as they may depend on few different interpreters(we may need additional sub-module, which will complicate the project even more).

openapi/openapi/src/main/scala/endpoints/documented/openapi/Requests.scala
julienrf: It is used [here](https://github.com/julienrf/endpoints/blob/0cf6d19045b13eea911670d64ae96bf5b93fc48f/openapi/openapi/src/main/scala/endpoints/documented/openapi/Requests.scala#L38). I moved the OpenAPI ADT into a `model` subpackage.

openapi/json-schema-circe/src/main/scala/endpoints/documented/circe/JsonSchemas.scala
julienrf: I started with explicit values and made them implicit at the very end of my work. With explicit values, since their name is explicitly written, it’s better if it’s short and simple. However we have lots of names in the scope when we import several algebras, so we have to make them unique, thus long/complex.

openapi/build.sbt
julienrf: Good catch :)

julienrf:
I’ve updated this work.

I recommend to start with the [`algebra.JsonSchemas`](https://github.com/julienrf/endpoints/pull/53/files#diff-c15a37579664e3a170dda9270525cfa7) definition, which introduces the vocabulary to describe algebraic data types. Then [`generic.JsonSchemas`](https://github.com/julienrf/endpoints/pull/53/files#diff-6de92a85f5be1f2fb4839d63aa65fedd) uses Shapeless generic derivation to automatically build such ADT descriptions from case class and sealed trait definitions. Finally, interpreters producing a JSON schema, a JSON decoder and a JSON encoder are straightforward.

It took me a lot of time to find a nice way to deal with sum types. I tried using the `Alternative` type class, but couldn’t figure out a way to generate the `Encoder`s. I also tried with `Cocartesian` but it didn’t give enough control on the way people could encode sum types (ie. the encoding of alternatives would have been automatic and very different from what people usually do). I tried `Arrow with Choice` but the code written by end users became too complex, so I ended up with something not as powerful or generic as `Arrow with Choice`, but that requires less boilerplate from users and can produce familiar JSON encodings. Here is an example:

~~~ scala
sealed trait Shape
case class Circle(radius: Double) extends Shape
case class Rectangle(width: Double, height: Double) extends Shape

object Shape {
  implicit val schema: JsonSchema[Shape] = genericJsonSchema[Shape] // let automatic derivation do the work for us…
}

// Then the following JSON document represents a `Rectangle`
// { "Rectangle": { "width": 10, "height": 10 } }

// And the following represents a `Circle`
// { "Circle": { "radius": 10 } }
~~~

openapi/json-schema-generic/src/main/scala/endpoints/documented/generic/JsonSchemas.scala
```diff
@@ -0,0 +1,163 @@
...
+      tupler: Tupler.Aux[L, T],
+      gen: Generic.Aux[T, L]
+    ): Record[T] = as[T]
+
+  }
+
+  implicit def toRecordGenericOps[A](record: Record[A]): RecordGenericOps[A :: HNil] =
+    new RecordGenericOps[A :: HNil](record.invmap(_ :: HNil)(_.head))
+
+}
```
Krever: That a bit above my paygrade, but do I understand correctly that by joining `generic.JsonSchemas` and `circe.JsonSchemas` we basically emulate/reimplement `circe-generic`?

openapi/openapi/src/main/scala/endpoints/documented/algebra/JsonSchemaEntities.scala
```diff
@@ -0,0 +1,18 @@
...
+  * `JsonRequest` and `JsonResponse` types to be `JsonSchema`.
+  */
+trait JsonSchemaEntities
+  extends JsonEntities
+    with JsonSchemas {
+
+  type JsonRequest[A] = JsonSchema[A]
+  type JsonResponse[A] = JsonSchema[A]
+
+}
```
Krever: So now the user can choose between `JsonEntities` and `JsonSchemaEntities`. Are there any use cases in which he couldn't go with `JsonSchemaEntites`? In other words, do we still need `JsonEntities`? The only reason I can think of is either for simplicity or for users who really don't want to use `shapeless`.

openapi/openapi/src/main/scala/endpoints/documented/openapi/JsonSchemaEntities.scala
```diff
@@ -0,0 +1,41 @@
...
+          alternatives.map { case (tag, record) =>
+            Schema.Object(Schema.Property(tag, toSchema(record), isRequired = true) :: Nil, None)
+          }
+        Schema.OneOf(alternativesSchemas, None)
+      case Primitive(name) => Schema.Primitive(name)
+      case Array(elementType) => Schema.Array(toSchema(elementType))
+    }
+  }
+
+}
```
Krever: I'm not that familiar with openapi, but do they allow for other content types than json? And if so, would it be possible for the user to have two endpoints defined, one with json and other one with xml(assuming we will support xml or he will wirte an interpreter himself)?

openapi/openapi/src/main/scala/endpoints/documented/openapi/model/OpenApi.scala
```diff
@@ -148,8 +148,53 @@ case class MediaType(schema: Option[Schema])
 object MediaType {
 
   def jsonMediaTypes(mediaTypes: Map[String, MediaType]): Json =
-    Json.fromFields(mediaTypes.map { case (tpe, mediaType) => tpe -> Json.obj() /* TODO Document the media schema */ })
+    Json.fromFields(mediaTypes.map { case (tpe, mediaType) =>
+      tpe -> Json.obj("schema" -> mediaType.schema.fold[Json](Json.obj())(_.asJson))
```
Krever: Tiny, but probably the most important line in the whole PR :+1:

openapi/json-schema-generic/src/main/scala/endpoints/documented/generic/JsonSchemas.scala
julienrf: Exactly! `generic.JsonSchemas` can derives a “schema” for a given algebraic data type and `circe.JsonSchemas` provides circe `Encoder` and `Decoder` for such schemas.

openapi/openapi/src/main/scala/endpoints/documented/algebra/JsonSchemaEntities.scala
julienrf: That’s a good point. I will check if it’s possible/easy to remove `JsonEntities`.

openapi/openapi/src/main/scala/endpoints/documented/openapi/JsonSchemaEntities.scala
julienrf: Yes they support any content-type. And yes we could define somewhere a method `xmlResponse` which would return a `DocumentedResponse` with a `Map("application/xml" -> …)`, in an `openapi` interpreter.

julienrf: We have nothing in place to support content-type negotiation, though…

julienrf:
> If I get this right, now we can present what exact request and response we expose in such tools as swagger UI, right?

Yes, exactly.

PR #66, opened 2017-11-12
Bump scala, circe and play versions

### LINKED ISSUE #54: Upgrade to Play 2.6
Also, that will make #1 more useful because then play-json could be used on client-side too!

Krever:
I bumped some versions to fix #54 and #55 .

The problem is I have no idea what I done with play. They converted `object` to `class` and I just followed the docs to make it compile, but I cant say if thats the correct thing to do. Let me know what can be done better or just apply fixes to the same branch :)

Another thing is that we are very close to being able to update docs to scala 2.12. The only dependency left on 2.11 is faithful, I think.


documentation/examples/basic/play-server/src/main/scala/sample/Server.scala
```diff
@@ -31,8 +31,39 @@ object Server extends App with Results {
...
+          |    <style>
+          |      body {
+          |        margin: 0;
+          |        padding: 0;
+          |      }
+          |    </style>
+          |  </head>
+          |  <body>
+          |    <redoc spec-url='/api/description'></redoc>
+          |    <script src="https://rebilly.github.io/ReDoc/releases/latest/redoc.min.js"> </script>
```
julienrf: Super cool!

play/client-circe/src/main/scala/endpoints/play/client/CirceEntities.scala
```diff
@@ -2,15 +2,20 @@ package endpoints.play.client
...
-import io.circe.jawn
-import endpoints.play.PlayCirce.circeJsonWriteable
 
 /**
   * Implements [[algebra.CirceEntities]] for [[play.client.Endpoints]]
   */
 trait CirceEntities extends algebra.CirceEntities { this: Endpoints =>
 
+  implicit def circeJsonBodyWriteable(implicit codec: Codec): BodyWritable[Json] =
+    new BodyWritable[Json](json => InMemoryBody(codec.encode(json.noSpaces)), ContentTypes.JSON)
```
julienrf: We used to have both `play-client-circe` and `play-server-circe` depend on a common `play-circe` artifact providing a `Writeable[Json]` instance. It seems that now, Play WS doesn’t use `Writeable` anymore? So we should probably remove this `play-circe` module?

documentation/examples/basic/play-server/src/main/scala/sample/Server.scala
Krever: I don't know how I missed this as this was not meant for this PR (I reviewd modified files twice). I will check if that still works and leave it here as it is really nice feature :)

play/client-circe/src/main/scala/endpoints/play/client/CirceEntities.scala
Krever: Your right, will migrate play-circe to play-server-circe.

PR #73, opened 2018-02-12
Add a “Use Cases” documentation page

julienrf:


documentation/examples/cqrs/commands-endpoints/src/main/scala/cqrs/commands/CommandsEndpoints.scala
```diff
@@ -12,14 +12,16 @@ import endpoints.algebra.{CirceEntities, Endpoints}
...
 trait CommandsEndpoints extends Endpoints with CirceEntities {
 
+//#microservice-endpoint-description
   /**
     * Application of a command.
     *
     * Returns the produced event, or `None` in case of failure (aggregate
     * not found or invalid command).
     */
-  val command/*: Endpoint[Command, Option[StoredEvent]]*/ =
```
Krever: Why removing the types?

julienrf: haha, this is embarrassing :)

I wish I could leave these type annotation uncommented but they make type inference fail. If I uncomment them, I get a compilation error, but if I omit them the code compiles fine (and the inferred type is the same as the one written as a comment…).

Since these lines of code are going to be part of the documentation I prefer not to have a maybe confusing commented type annotation.

Krever: Yeah, I struggled with this as well at some point. The only way for me to keep type annotations was to put explicit types in one of the calls(`get` if I remember correctly). 

I don't know if that's a general rule, but having types annotation allows me to learn a new library faster.

julienrf: I’ve added back type information as comments.

PR #79, opened 2018-02-21
fixing some github doc links

bbarker:


julienrf:
This is weird, the doc renders correctly on my machine (and on julienrf.github.io/endpoints). You have to run sbt from the root directory, not from the `documentation` directory. Not sure if that’s the issue…

bbarker:
Sorry, i should have been more descriptive. The problem I have is that on  julienrf.github.io/endpoints, this link (for example) isn't working properly:

image

julienrf:
Oh, indeed, thanks for reporting! We should fix this issue by adding `documentation/` to the path defined here: https://github.com/julienrf/endpoints/blob/master/documentation/manual/src/ornate.conf#L26

bbarker:
Aha. As a matter of curiosity, I guess I'm still a bit confused about why the other urls I changed in this PR don't need the same change... but I've never used Ornate.

bbarker:
Oh right, I see your comment about the relative directory bit.

PR #82, opened 2018-03-04
Unify clients and refactor a bit

Krever:
I added a couple of missing interpreters and fixed one bug in scalaj.

I renamed the type parameters in requests algebra because it's one of the most confusing pieces for the newcomers in my experience.

### DIFF
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
@@ -7,7 +7,10 @@ import endpoints.{Tupler, algebra}
 import scala.concurrent.{ExecutionContext, Future}
 
 class Endpoints(val settings: EndpointsSettings)
-               (implicit val EC: ExecutionContext, val M: Materializer) extends algebra.Endpoints with Urls with Methods {
+  (implicit val EC: ExecutionContext, val M: Materializer)
+  extends algebra.Endpoints
+    with Urls
+    with Methods {
   type RequestHeaders[A] = (A, List[HttpHeader]) => List[HttpHeader]
 
   lazy val emptyHeaders: RequestHeaders[Unit] = (_, req) => req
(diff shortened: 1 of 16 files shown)


akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
```diff
@@ -56,8 +64,10 @@ class Endpoints(val settings: EndpointsSettings)
       for {
         resp <- request(a)
         result <- response(resp).flatMap(futureFromEither)
+        _ = resp.discardEntityBytes() //Fix for https://github.com/akka/akka-http/issues/1495
```
julienrf: Good catch!

play/client/src/main/scala/endpoints/play/client/Endpoints.scala
```diff
@@ -54,10 +54,16 @@ class Endpoints(host: String, wsClient: WSClient)(implicit val executionContext:
...
+    case resp if resp.status >= 200 && resp.status < 300 => Right(())
+    case resp => Left(new Throwable(s"Unexpected status code: ${resp.status}"))
+  }
 
   /** Successfully decodes string information from a response */
-  val textResponse: Response[String] = x => Right(x.body)
+  val textResponse: Response[String] = {
+    case resp if resp.status >= 200 && resp.status < 300 => Right(resp.body)
+    case resp => Left(new Throwable(s"Unexpected status code: ${resp.status}"))
+  }
```
julienrf: That’s probably out of the scope of this PR but it would really help to introduce a separate `ResponseEntity` type, so that we can separate filtering the response status and decoding the response entity.

Krever: Yes, I had exactly the same thought. I even started to think about this but it seems harder than expected. E.g. `jsonResponse` right specifies(or should specify) 3 different things:
* ok response code(200 or maybe 2xx?)
* json encoded entity
* `Content-Type` header value

And this raises a question: how to implement basic algebra for defining e.g. response headers and status codes, so that `jsonRepsonse` can integrate with it easily(so preserve user-defined headers). 

Also much harder topic: how to implement status filtering? Should we allow to accept codes ranges(e.g. 2xx)? Should we introduce some kind of `or` operator and coproducts of payloads in such case? 

Maybe I will figure something out but don't have a good answer yet.

julienrf: I don’t have a good answer either!

Fristi: I have done it here :-) 

https://github.com/Fristi/endpoints-proto/blob/master/example/src/main/scala/itinere/example/ApiEndpoints.scala#L26

PR #83, opened 2018-03-15
Abstract over JSON codecs

### LINKED ISSUE #1: JSON module that uses Play formats
Currently, only circe is supported out of the box for JSON marshalling. It would be great to add a module `PlayOFormats` similar to `CirceCodecs`, but which would use Play’s `OFormat` instead of circe codecs. Then, we could define a `PlayOFormatsClient` and a `PlayOFormatsRouting` modules, similar to the current `CirceCodecsClient` and `CirceCodecsRouting`.

### LINKED ISSUE #58: Json Encoder/Decoder adapter type class
When you want to support a new JSON library in endpoints, you need to implement `JsonEntities` which is defined like this:

```
trait JsonEntities extends Endpoints {

  type JsonRequest[A]

  def jsonRequest[A : JsonRequest]: RequestEntity[A]

  type JsonResponse[A]

  def jsonResponse[A : JsonResponse]: Response[A]

}
```

This will couple the Decoding and Encoding part directly. To avoid such thing we introduce adapter typeclasses. We'll define a `JsonEncoder` and `JsonDecoder` ourselves and convert the decoders/encoder types from other libraries to our own type class.

A example can be found here: https://github.com/Fristi/endpoints-proto/blob/master/json-circe/src/main/scala/itinere/json/circe/CirceJsonCodec.scala#L21
It will use circe Codec to convert it into a itinere JsonCodec. This needs to be split up to Encoder/Decoder in here.

So for each json lib we can support we'll introduce a module which will implement the endpoint Encoder/Decoder type class. By doing this we can drop all the modules like `play-server-circe`, `akka-http-server-circe`, etc

julienrf:
Fixes #58
Fixes #1

In a nutshell, instead of having `CirceEntities` and introducing `PlayJsonEntities`, we have a single [`JsonEntitiesFromCodec`](https://github.com/julienrf/endpoints/pull/83/files#diff-2918303bf555329a13f697a6058f48e1) algebra interface, which uses a [`Codec[String, A]`](https://github.com/julienrf/endpoints/pull/83/files#diff-a5c38521270dfdc96accdc67792f2939) to encode and decode JSON entities. (Thanks to @Fristi for the idea)

Then, this algebra interface is partially interpreted by [`circe.JsonEntitiesFromCodec`](https://github.com/julienrf/endpoints/pull/83/files#diff-a428b8aa643261da7902550c4e91cdd4R58) (respectively [`playjson.JsonEntitiesFromCodec`](https://github.com/julienrf/endpoints/pull/83/files#diff-4b0e129a5d99d4941c275a3e77f10222R58)), which fixes the `JsonCodec` to `CirceCodec` (respectively `play.api.libs.json.Format`).

Client and server interpreters can just implement `JsonEntitiesFromCodec` without knowing the underlying `JsonCodec[A]` (which has been converted to a `Codec[String, A]`). Example for [`play.server.JsonEntitiesFromCodec`](https://github.com/julienrf/endpoints/pull/83/files#diff-9b160142bdf31d7a78ca0b5bfc32fe08).

It allows us to remove most of the `-circe` modules, and to support Play JSON without multiplying our number of modules by two.


play/server-circe/src/main/scala/endpoints/play/server/circe/JsonEntities.scala
```diff
@@ -22,10 +22,10 @@ trait JsonEntities extends Endpoints with algebra.JsonEntities {
...
-    playComponents.playBodyParsers.raw.validate { buffer =>
-      jawn.parseFile(buffer.asFile)
+    playComponents.playBodyParsers.tolerantText.validate { text =>
+      parser.parse(text)
         .right.flatMap(CirceDecoder[A].decodeJson)
-        .left.map(error => Results.BadRequest)
+        .left.map(ignoredError => Results.BadRequest)
     }
 
   def jsonResponse[A : CirceEncoder]: Response[A] = a => Results.Ok(CirceEncoder[A].apply(a))
```
Krever: Why do we want to have direct circe support for play-server but dropped all the others? Just asking.

Krever:
That's really nice! I love how easy it is now to add support for other json libs.

This also made me think about some more general question of "how thick layer of abstraction endpoints want to be". Originally we were just proxying the call and the types directly to backend layer. If we want to embed more and more functionality it inevitably leads to having some abstraction in the middle, like `Codec` here. Can this be a problem for someone users? After `jsonCodec` is called we irretrievably lose the information carried by underlying `JsonCodec`. 

I tried to imagine a scenario in which this could be a problem but I couldn't, so maybe it's not a real problem. 

In general, this comes to the question: when introducing our own abstractions should fix the types or just bounds?
So the choice between
```
type JsonRequest[A] = Codec[String, A]
```
and 
```
type JsonRequest[A] <: Codec[String, A]
```

With the first, once the type is fixed there is no way to extend/modify it. With the second it requires additional line in the interpreter, where in most of the cases it will be
```
type JsonRequest[A] = Codec[String, A]
```
anyway.

I think this question will come more and more as we want to have more features in `endpoints`.

akka-http/client/src/main/scala/endpoints/akkahttp/client/JsonEntitiesFromCodec.scala
```diff
@@ -0,0 +1,28 @@
...
+trait JsonEntitiesFromCodec extends endpoints.algebra.JsonEntitiesFromCodec { this: Endpoints =>
+
+  def jsonRequest[A](implicit codec: Codec[String, A]): RequestEntity[A] = { (a, req) =>
+    req.copy(entity = HttpEntity(ContentTypes.`application/json`, codec.encode(a)))
+  }
+
+  def jsonResponse[A](implicit codec: Codec[String, A]): Response[A] = { response =>
+    for {
+      strictEntity <- response.entity.toStrict(settings.toStrictTimeout)
+      res <- codec.decode(settings.stringContentExtractor(strictEntity)) match {
```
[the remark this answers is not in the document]
julienrf: Oops, that can be simplified, indeed!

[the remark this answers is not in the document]
julienrf: Fixed in 95ac496

play/server-circe/src/main/scala/endpoints/play/server/circe/JsonEntities.scala
julienrf: I find this one useful when I directly write a server, without having a “shared” description of the HTTP API. In such a case I don’t even need a JSON decoder for responses (only an encoder), so I don’t use the `JsonEntitiesFromCodec` thing.

julienrf:
> Originally we were just proxying the call and the types directly to backend layer. If we want to embed more and more functionality it inevitably leads to having some abstraction in the middle, like Codec here. Can this be a problem for someone users? After `jsonCodec` is called we irretrievably lose the information carried by underlying `JsonCodec`.

Yes, you’re right. This is a tradeoff. One limitation of the current form is that the way we handle failures in JSON decoding is fixed by the `jsonCodec`, the `Codec#decode` signature and the `jsonRequest` implementations (in the case of servers) and it might not fit all users needs. For instance, we currently return an empty response entity in case of decoding failure, whereas some applications might want to return a JSON document describing the errors (e.g. which validation constraint failed, which field was missing, etc.). Users can override `jsonCodec` and `jsonRequest` to customize that behaviour according to their needs. However we might want to abstract over the type of decoding failures (currently it’s `Exception`) so that users can take advantage of the information provided by their underlying JSON library.

play/server-circe/src/main/scala/endpoints/play/server/circe/JsonEntities.scala
Krever: I see. It may be a little confusing that only a single interpreter has direct support, but maybe that's not a problem :)

julienrf:
> In general, this comes to the question: when introducing our own abstractions should fix the types or just bounds?

Interesting point. Indeed there is a decision to be made. Initially I wanted to fix as few things as possible. But with time I thought it could be more user-friendly or productive to do some (reasonable) choices for the users. Let me know if you have a different opinion.

Krever:
I think I agree. In general cost of having full extensibility is too high.

PR #85, opened 2018-03-20
Support documentation of JSON schema fields

### LINKED ISSUE #81: Support documentation of json schema fields
Currently the `field` (and `optField`) constructor take only a `name` as parameter. They should also take an optional `documentation: Option[String]` parameter (with a `None` default value), which should be used by the OpenAPI interpreter.

julienrf:
Fixes #81


Fristi:
Looks good, build fails though

Issue #70, opened 2017-12-13
sttp client interpreter


Krever:
[sttp](https://github.com/softwaremill/sttp) is a relatively new but great abstraction over http clients. It provides variety of different clients that can be used([list](http://sttp.readthedocs.io/en/latest/backends/summary.html#))

We would definitely benefit from integration with sttp because then we get multiple backends for single interpreter.

Fristi:
Very good idea, I think we can drop most clients if we implement a sttp based one as it has many backends:

http://sttp.readthedocs.io/en/latest/backends/summary.html

PR #95, opened 2018-04-27
Add sttp client

vdebergue:
This pull request adds support for the [sttp](https://github.com/softwaremill/sttp) client (#70)

The code is mostly similar to play-ws and akka-http clients (`Urls.scala` is the same).

The differences are:
- A backend defines a monad `R[_]` that wraps the responses, this shows in the api.
- `sttp` requires that we define the expected response type when creating a request (with `ResponseAs`). This means that we cannot send the request in `def request` since we don't know yet the response type. In consequence, we send the request in `def endpoint` where we know the response
- an `Endpoint` will return a type `R[Either[String, B]]` whereas the other clients simply returns a `Future[B]`

Otherwise it was quite easy to implement this new backend and to add tests for it 👍

### DIFF
--- a/sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
+++ b/sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
@@ -63,24 +63,24 @@ class Endpoints[R[_]](host: String, val backend: sttp.SttpBackend[R, Nothing]) e
     /**
       * The type of the received body from the server
       */
-    type RB
+    type ReceivedBody
 
     /**
       * To read the response body
       */
-    def responseAs: sttp.ResponseAs[RB, Nothing]
+    def responseAs: sttp.ResponseAs[ReceivedBody, Nothing]
 
     /**
       * Function to validate the response (headers, code). This can also modify the type of the received body
       */
-    def validateResponse(response: sttp.Response[RB]): R[A]
+    def validateResponse(response: sttp.Response[ReceivedBody]): R[A]
   }
 
   type Response[A] = SttpResponse[A]
 
   /** Successfully decodes no information from a response */
   val emptyResponse: Response[Unit] = new SttpResponse[Unit] {
-    type RB = Unit
+    override type ReceivedBody = Unit
     override def responseAs = sttp.ignore
     override def validateResponse(response: sttp.Response[Unit]) = {
       if (response.isSuccess) backend.responseMonad.unit(response.unsafeBody)
@@ -90,7 +90,7 @@ class Endpoints[R[_]](host: String, val backend: sttp.SttpBackend[R, Nothing]) e
 
   /** Successfully decodes string information from a response */
   val textResponse: Response[String] = new SttpResponse[String] {
-    type RB = String
+    override type ReceivedBody = String
     override def responseAs = sttp.asString
     override def validateResponse(response: sttp.Response[String]) = {
       if (response.isSuccess) backend.responseMonad.unit(response.unsafeBody)
@@ -105,7 +105,7 @@ class Endpoints[R[_]](host: String, val backend: sttp.SttpBackend[R, Nothing]) e
 
   def endpoint[A, B](request: Request[A], response: Response[B]): Endpoint[A, B] =
     a => {
-      val req: sttp.Request[response.RB, Nothing] = request(a).response(response.responseAs)
+      val req: sttp.Request[response.ReceivedBody, Nothing] = request(a).response(response.responseAs)
 
       val result = backend.send(req)
       backend.responseMonad.flatMap(result)(response.validateResponse)
--- a/sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
+++ b/sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
@@ -1,0 +1,112 @@
+package endpoints.sttp.client
+
+import java.net.URI
+
+import endpoints.algebra
+import endpoints.Tupler
+import com.softwaremill.sttp
+
+import scala.language.higherKinds
+
+/**
+  * An interpreter for [[endpoints.algebra.Endpoints]] that builds a client issuing requests using
+  * a sttp’s [[com.softwaremill.sttp.SttpBackend]].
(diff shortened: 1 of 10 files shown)


sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
```diff
@@ -0,0 +1,112 @@
...
+  * @param backend  The underlying backend to use
+  * @tparam R       The monad wrapping the response. It is defined by the backend
+  */
+class Endpoints[R[_]](host: String, val backend: sttp.SttpBackend[R, Nothing]) extends algebra.Endpoints with Urls with Methods {
+
+  /**
+    * A function that, given an `A` and a request model, returns an updated request
+    * containing additional headers
+    */
+  type RequestHeaders[A] = (A, sttp.Request[_, Nothing]) => sttp.Request[_, Nothing]
```
Krever: can we type alias `sttp.Request[_, Nothing]`? It seems to clutter the code

sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
```diff
@@ -0,0 +1,112 @@
...
+    override def validateResponse(response: sttp.Response[String]) = {
+      if (response.isSuccess) response.body
+      else Left(s"Unexpected status code: ${response.code}")
+    }
+  }
+
+  /**
+    * A function that, given an `A`, eventually attempts to decode the `B` response.
+    */
+  type Endpoint[A, B] = A => R[Either[String, B]]
```
Krever: I'm not sure about that. The idea, at least for me, behind endpoints was that properly defined protocol should never fail because of expected reasons. And here we say something different. Maybe that's ok, but I would like to understand in which cases we will receive `Left`.

vdebergue: According to the [documentation](http://sttp.readthedocs.io/en/latest/responses/basics.html#obtaining-the-response-body):

```
The response.body will be a:
Left(errorMessage) if the request is successful, but response code is not 2xx.
Right(deserializedBody) if the request is successful and the response code is 2xx.
``` 

If you think that defined endpoints should not fail because of expected reasons, `.unsafeBody` may work better in our case

Krever: Maybe we could leave it up to the user? So have an `apply` which will call `unsafeBody` and some other method that will give back the `Either`. I wouldn't like to force the user to write `right.get` every time even if he is sure that server and client are consistent. Nevertheless, if this comes because of underlying API I leave it up to you to decide :)

julienrf:
This is cool, thanks Vincent ;)

Just to play the devil’s advocate: what’s the advantage of this backend? FWIU sttp has a nice user-facing API, but in our case it does not really matter because end users don’t use it directly since they use the *endpoints* API.

The only advantage I see is that if you already use sttp in some other parts of your program, then it’s good to also have *endpoints* use it.

sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
julienrf: I’m not familiar with sttp, what’s the role of these two type parameters?

sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+  /**
+    * Trait that indicates how a response should be interpreted
+    */
+  trait SttpResponse[A] {
+    /**
+      * The type of the received body from the server
+      */
+    type RB
```
julienrf: I would go for `ReceivedBody`, for clarity.

Krever:
> Just to play the devil’s advocate: what’s the advantage of this backend? FWIU sttp has a nice user-facing API, but in our case it does not really matter because end users don’t use it directly since they use the endpoints API.

It has a couple of different backends, so we get for free whatever they develop :)

sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
julienrf: `unsafeBody` doesn’t sound appealing because it throws.

That being said I agree with @Krever, we tend to give `Endpoint[A, B]` a concrete type that doesn’t require users to deal with failures due to communications. Typically, it is represented by a function `A => F[B]`, where `F` has a `MonadError[F, Exception]` instance. Do we have the same guarantees for `R`, here?

julienrf: (by failures due to communication, I mean network failures but also data marshalling failures)

sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+      * To read the response body
+      */
+    def responseAs: sttp.ResponseAs[RB, Nothing]
+
+    /**
+      * Function to validate the response (headers, code). This can also modify the type of the received body
+      */
+    def validateResponse(response: sttp.Response[RB]): Either[String, A]
```
julienrf: I’ve just had a look at sttp’s documentation and it seems that `R` is not just a `Monad` but even a `MonadError`, so I suggest that you return a `R[A]` here, instead of a `Either[String, A]`.

sttp/client/src/main/scala/endpoints/sttp/client/Endpoints.scala
vdebergue: The first type is the type of the body in the request: `String` `Array[Int]` `Inputstream`...

The second allows to specify if the response should be streamed and should match the second parameter of the backend. For strict backend it uses `Nothing`, and for streaming backend like akka-http it will be `Source[ByteString, Any]`

Since it's the first implementation and I'm not really familiar with the streaming part, I prefered focusing on the strict backend.

Krever:
Thanks!

PR #98, opened 2018-06-20
Update dependencies (cats and circe)

julienrf:
This PR picks @easel’s work and fixes one remaining dependency. Sbt is kept to 0.13 because ornate is not yet available for 1.0.


julienrf:
It seems that the CI has been stabilized (I hope, at least…). Do you agree to upgrade the dependencies before we do a new release? (@easel and @Krever)

PR #102, opened 2018-08-01
[Experiment] Compose algebras via composition instead of inheritance

Krever:
I wanted to check if this would work and it seems that it would indeed. 

The biggest benefit for me is a clear distinction between base algebra of an interpreter and dependent algebras. It becomes much easier to reason about the interpreters and to create such. The confusion about which traits need to be mixed in should be much smaller with this approach.

The user still has all the power to extend/override the implementation. The biggest drawback I see is the necessity of importing contents of all the algebras separately.

The consistency between interpreters can be enforced by type refinements if needed.

In general, I'm in favour of this encoding if there would be no other problems. I believe that being able to easily reason about algebras pays the cost of some additional boilerplate.

One unknown is: should interpreter instantiate the dependent interpreters or leave it to the end user? It already needs to override the type of those, so creating an instance wouldn't change much. (rephrase: should `Endpoints` instantiate `Requests`)?

### DIFF
--- a/algebras/algebra-circe/src/test/scala/endpoints/algebra/circe/JsonFromCirceCodecTestApi.scala
+++ b/algebras/algebra-circe/src/test/scala/endpoints/algebra/circe/JsonFromCirceCodecTestApi.scala
@@ -3,8 +3,9 @@ package endpoints.algebra.circe
 import endpoints.algebra.{Address, JsonFromCodecTestApi, User}
 
 trait JsonFromCirceCodecTestApi
-  extends JsonFromCodecTestApi
-    with endpoints.algebra.circe.JsonEntitiesFromCodec {
+  extends JsonFromCodecTestApi {
+
+  val entities: endpoints.algebra.circe.JsonEntitiesFromCodec
 
   import io.circe._
   import io.circe.generic.semiauto._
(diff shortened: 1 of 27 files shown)


julienrf:
I think the biggest drawback (besides having to explicitly import things) is that it can become complicated to *align* types (wire together parts of interpreters that agree on the underlying carrier types) because of the nesting, leading to complex path-dependent type refinements.

Krever:
Yes, that's definitely an issue, but I rarely ever see this problem in practice. AFAIK the refinements would only be needed if we would like to align on algebra level. When we do this on interpreter level, the types become aligned automatically via proper construction. Of course, there is nothing stopping the user from constructing invalid(unaligned) interpreter but this would mean he somehow need to get 2 different interpreters for a given algebra and this would be hard to achieve if he is manually creating all of them(it would be more probable if the interpreter instantiate the dependent algebras, so maybe that's a good argument against it). 

Do you have any particular example that I could play with to see how problematic it is in practice? One that I can think of is composing some custom `Validation`, `Authorization` and `Serialization` algebras but these would probably be independent of one another and would not require any alignment on algebra level (they all would probably depend on `Endpoints` directly).

julienrf:
I don’t have it off the top of my head, I remember to have tried your design and to have struggled with that issue. But maybe you found a variation that doesn’t suffer from that problem… I would recommend trying to get a small example compiling (preferably one that uses `JsonSchemaEntities`). If you don’t hit the issue, that’s great!

Then to answer the following question:

> One unknown is: should interpreter instantiate the dependent interpreters or leave it to the end user? It already needs to override the type of those, so creating an instance wouldn't change much. (rephrase: should Endpoints instantiate Requests)?

I would say yes. That seems to provide a sensible default, that could always be overridden by users if needed.

Krever:
I may get back to it at some point but lets close it for now.

Issue #105, opened 2018-08-10
Basic support for OpenApi tags


krzemin:
It'd be really nice if endpoints could support grouping operations by tags in OpenApi documentation, as in the example here: https://swagger.io/docs/specification/grouping-operations-with-tags/

I think it's sufficient to annotate operations only with optional tag list (empty by default). Providing more detailed tag description on root level would be also nice, but not necessarily in scope of this ticket.

After quick investigation, I think it may be quite simple to implement, although IMHO it requires modifying `endpoint` signature to:

```scala
def endpoint[A, B](
    request: Request[A],
    response: Response[B],
    summary: Documentation = None,
    description: Documentation = None,
    tags: List[String] = Nil
  ): Endpoint[A, B]
```

Are you OK to go this way?

Krever:
For me its ok. OpenAPI is more or less a standard for describing endpoints so aligning our API with it seems reasonable, as long as optional parts can be ommited.

PR #106, opened 2018-08-10
Basic support for OpenApi tags

krzemin:
This PR addresses #105


algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
```diff
@@ -39,7 +39,8 @@ trait Endpoints extends Requests with Responses {
     request: Request[A],
     response: Response[B],
     summary: Documentation = None,
-    description: Documentation = None
+    description: Documentation = None,
+    tags: List[String] = Nil
```
julienrf: Can you add a `@param` scaladoc documentation?

krzemin: sure, added

PR #107, opened 2018-08-11
Support for empty objects in JsonSchemas

krzemin:
This PR is follow up to discussion on gitter: https://gitter.im/julienrf/endpoints?at=5b6eac942a8e6c6083aaf21e

* added `emptyRecord` JSON schemas dsl primitive
* generic JSON schema derivation for both singleton objects and case classes with no parameters should be supported

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
@@ -33,6 +33,9 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   }
 
+  def emptyRecord: DocumentedRecord =
+    DocumentedRecord(Nil)
+
   def field[A](name: String, docs: Documentation)(implicit tpe: DocumentedJsonSchema): DocumentedRecord =
     DocumentedRecord(Field(name, tpe, isOptional = false, docs) :: Nil)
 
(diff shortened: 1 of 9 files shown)


krzemin:
Weird - scala 2.10 tests pass, while 2.11 and 2.12 fail at
```
scala.reflect.internal.MissingRequirementError: object scala.runtime in compiler mirror not found.
```

Krever:
I allowed myself to bump sbt version in your branch as suggested in https://github.com/sbt/sbt/issues/2958. Lets see if that helps.

Krever:
It has not. I also cleared the cache in travis. Let's see if that helps. If not, then I'm out of ideas.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -33,6 +33,9 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   }
 
+  def emptyRecord: DocumentedRecord =
+    DocumentedRecord(Nil)
```
julienrf: Can you do a quick check that `DocumentedRecord` usages don’t expect the list of fields to not be empty?

krzemin: Checked in the IDE and didn't find such case ✅

krzemin:
@Krever something actually helped - now the test jobs are green, but the coverage fails at:

```
[info] EndpointsTest:
[info] Client interpreter
[info] - should return server response *** FAILED ***
[info]   A timeout occurred waiting for a future to complete. Queried 119 times, sleeping 10000000 nanoseconds between each query. (EndpointsTestSuite.scala:24)
```

but it doesn't look like something caused by this change

julienrf:
I don’t understand why when we run the tests with coverage enabled they take forever and fail the CI server…

PR #109, opened 2018-08-17
Fix decoder for empty object

krzemin:
This PR is follow up to https://github.com/julienrf/endpoints/pull/107

* I defined circe decoder for `emptyRecord` that was using `io.circe.Decoder.decodeUnit` which actually expects empty object as an input and fails otherwise (in case it was given object with few fields)
* As we use this decoder instance in [base case](https://github.com/julienrf/endpoints/blob/master/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala#L55) of generic derivation, using such derived decoder was failing to decode any non-empty record
* This PR introduces decoder that is a bit more liberal - it accepts any json object, not necessarily empty which seems to fix the problem

### DIFF
--- a/json-schema/json-schema/src/test/scala/endpoints/algebra/JsonSchemasTest.scala
+++ b/json-schema/json-schema/src/test/scala/endpoints/algebra/JsonSchemasTest.scala
@@ -16,6 +16,14 @@ trait JsonSchemasTest extends JsonSchemas {
       field[String]("name", Some("Name of the user")) zip
       field[Int]("age")
     ).invmap((User.apply _).tupled)(Function.unlift(User.unapply))
+
+    val schema2: JsonSchema[User] = (
+      emptyRecord zip
+      field[String]("name", Some("Name of the user")) zip
+      field[Int]("age")
+    )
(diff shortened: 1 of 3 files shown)


json-schema/json-schema/src/test/scala/endpoints/algebra/JsonSchemasTest.scala
```diff
@@ -16,6 +16,14 @@ trait JsonSchemasTest extends JsonSchemas {
       field[String]("name", Some("Name of the user")) zip
       field[Int]("age")
     ).invmap((User.apply _).tupled)(Function.unlift(User.unapply))
+
+    val schema2: JsonSchema[User] = (
+      emptyRecord zip
+      field[String]("name", Some("Name of the user")) zip
+      field[Int]("age")
+    )
+      .invmap[(String, Int)](p => (p._1._2, p._2))(p => (((), p._1), p._2))
```
krzemin: ideally would like to use `.tupled[User]`, but `RecordGenericOps` are only available in `json-schema-generic`

PR #114, opened 2018-09-09
[WIP] update core algebra to scala 2.13.0-M2

pjazdzewski1990:
Done as part of https://github.com/julienrf/endpoints/issues/112
Work in progress

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,16 @@ cache:
     - $HOME/.ivy2/cache
     - $HOME/.sbt/boot
 
+# based on https://github.com/jruby/jruby/blob/master/.travis.yml
+before_install:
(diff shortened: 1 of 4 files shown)


pjazdzewski1990:
PR blocked on a (most likely) issue in Scalacheck. Working on a solution

pjazdzewski1990:
@Krever Thanks. Please wait a bit with the merge. There's one more thing I want to take a look at - the 2.13 build is much faster than the others and it might not be running the tests as it should be. I'm trying to figure it out.

Krever:
@pjazdzewski1990 I believe the diff in time is because of the number of modules being built. I will merge it now because you fixed the scoverage so we can have the nice green badge in readme :). If you want to look at anything else feel free to come with next PR :)

Issue #63, opened 2017-11-10
Urls.scala fixes


Fristi:
In #52 I've applied some changes to `Urls.scala` which fixes some issues while using query strings in urls. When there were no query string params it would encode with `path?` and when there was only query string present it would encode as `path?qs=1&`.

I see it's used for clients at:
- play-client
- scalaj

Basically the diffs boils down to these changes: https://www.diffchecker.com/c4FWHCnk (left is play-client, right is akka-http-client).

Krever:
It may be nice in the end to have it implemented in a similar way in all client interpreters. I'm not fully convinced with encoding is better but probably `None` is better than an empty string as it allows to create more reasonable urls(with an empty string we can end up with `www.example.com/?&x=1` if I recall correctly and this was probably the origin of change, right?)

The more important question that comes to my mind in regard to this change is: should we try to introduce common skeleton for client interpreters and the second one for server ones? 

As we can see here, they tend to implement some parts in the exact same way...

julienrf:
> The more important question that comes to my mind in regard to this change is: should we try to introduce common skeleton for client interpreters and the second one for server ones?

@Krever I wouldn’t do that unless we really need it.

Krever:
Agree, but everything has it cost. In case of duplication, we don't have a way to make sure that bug fixed once won't appear in other interpreters... Or maybe we do, with the tests!

Another problem is that at some point we will need to decide if we want to support "all possible interpreters" or just "reasonable subset of possible interpreters", because in case of many similar interpreters without common parts maintenance will become hell of a burden.

I agree that it's probably too soon to introduce such skeleton, and also it kills one of the main points of the lib, freedom of interpretation.

Krever:
Fixed in #116

PR #116, opened 2018-09-09
Urls.scala changed according to discussion in #63

piotrkosecki:
added test which checks if the optional parameter works correctly, changed play-client Urls.scala implementation, for scalaj it looks it's not required(it is being resolved by the lib)


Krever:
Scoverage has problem again, merging anyway.

julienrf:
Thank you so much @piotrkosecki !

PR #117, opened 2018-09-09
Test suite for server interpreters

Yuzvin:
Wrote test suite for server interpreter and tested it using Akka-Http server.

Fixes #117


julienrf:
Fixes #117

algebras/algebra/src/test/scala/endpoints/algebra/server/EndpointsTestSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+        val server = serveEndpoint(serverApi.smokeEndpoint, mockedResponse)
+        server.start()
+
+        val response  = sttp.get(uri"http://localhost:${server.port}/user/userId/description?name=name1&age=18").send()
+        assert(response.body.isRight)
+        assert(response.body.right.get == mockedResponse)
+        assert(response.code == 200)
+
+        server.stop()
+      }
```
julienrf: What do you think if, instead, `serveEndpoint` takes an additional function parameter that is called after the server is started and before it is stopped? Test writers wouldn’t have to manually take care of starting and stopping the server.

algebras/build.sbt
```diff
@@ -8,6 +8,7 @@ val algebra =
       name := "endpoints-algebra",
       libraryDependencies ++= Seq(
         "com.github.tomakehurst" % "wiremock" % "2.6.0" % Test,
+        "com.softwaremill.sttp" %% "core" % "1.3.2" % Test,
```
julienrf: The CI failed with an error related to this dependency. Can you check that we don’t miss another dependency?

NK91: It seems `sttp` has dependencies to `akka-http 10.1.4` and `akka-stream 2.5.16`, maybe it's the reason of failure?

NK91: I think changing the version of the `sttp` to `1.1.13` can help.

PR #128, opened 2018-10-04
[WIP] Add documentation support for generic schema derivation

guersam:
This commits allows to add field documentations using annotation.

example:
```scala
case class User(
  @documentation("The name of the user") name: String,
  @documentation("The age of the user") age: Int
)
object User {
  implicit val schema: JsonSchema[User] = genericJsonSchema[User]
}
```

It doesn't support coproducts yet. Maybe we can add `docs` parameter to  `JsonSchema#taggedField`?

### DIFF
--- a/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
@@ -79,7 +79,7 @@ trait JsonSchemas extends algebra.JsonSchemas {
     implicit def consRecord[L <: Symbol, H, T <: HList](implicit
       labelHead: Witness.Aux[L],
       jsonSchemaHead: JsonSchema[H],
-      jsonSchemaTail: GenericRecord[T],
+      jsonSchemaTail: GenericRecord[T]
     ): GenericRecord[FieldType[L, H] :: T] =
       new GenericRecord[FieldType[L, H] :: T] {
         def jsonSchema(docs: List[Option[documentation]]): Record[FieldType[L, H] :: T] =
@@ -90,7 +90,7 @@ trait JsonSchemas extends algebra.JsonSchemas {
     implicit def consOptRecord[L <: Symbol, H, T <: HList](implicit
       labelHead: Witness.Aux[L],
       jsonSchemaHead: JsonSchema[H],
-      jsonSchemaTail: GenericRecord[T],
+      jsonSchemaTail: GenericRecord[T]
     ): GenericRecord[FieldType[L, Option[H]] :: T] =
       new GenericRecord[FieldType[L, Option[H]] :: T] {
         def jsonSchema(docs: List[Option[documentation]]): Record[FieldType[L, Option[H]] :: T] =
--- a/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
@@ -2,8 +2,8 @@ package endpoints
 package generic
 
 import shapeless.labelled.{FieldType, field => shapelessField}
-import shapeless.ops.hlist.Tupler
-import shapeless.{:+:, ::, CNil, Coproduct, Generic, HList, HNil, Inl, Inr, LabelledGeneric, Witness}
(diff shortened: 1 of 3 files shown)


json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
```diff
@@ -120,29 +121,35 @@ trait JsonSchemas extends algebra.JsonSchemas {
...
+    implicit def taggedGeneric[A, R, D <: HList](implicit
       gen: LabelledGeneric.Aux[A, R],
       tagged: GenericTagged[R],
+      docAnns: Annotations.Aux[documentation, A, D],
+      docsToList: ToList[D, Option[documentation]],
       ct: ClassTag[A]
     ): GenericTagged[A] =
       new GenericTagged[A] {
-        def jsonSchema: Tagged[A] = nameSchema(tagged.jsonSchema.invmap[A](gen.from)(gen.to))
+        def jsonSchema(docs: List[Option[documentation]]): Tagged[A] =
```
guersam: Unfortunately it involves somewhat weird runtime propagation like this. Maybe I should add a comment that we're ignoring the method argument on purpose or even an assertion to make sure that `docs == Nil`.

json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
```diff
@@ -103,9 +104,9 @@ trait JsonSchemas extends algebra.JsonSchemas {
       taggedTail: GenericTagged[T]
     ): GenericTagged[FieldType[L, H] :+: T] =
       new GenericTagged[FieldType[L, H] :+: T] {
-        def jsonSchema: Tagged[FieldType[L, H] :+: T] = {
-          val taggedHead = recordHead.jsonSchema.tagged(labelHead.value.name)
-          taggedHead.orElse(taggedTail.jsonSchema).invmap[FieldType[L, H] :+: T] {
+        def jsonSchema(docs: List[Option[documentation]]): Tagged[FieldType[L, H] :+: T] = {
+          val taggedHead = recordHead.jsonSchema(Nil).tagged(labelHead.value.name)
```
guersam: This `Nil` makes me a bit sad :(

julienrf: Why don’t we use `docs.head`? (I’m not familiar with `shapeless.Annotation`)

guersam: Shortest reason:
`docs.head` returns `Option[documentation]`, not `List[Option[documentation]]`.

Longer reason:
I haven't established a strategy to document each coproduct branch yet. In this case, the result of `shapeless.Annotations[documentation, A]` is the list of `@documentation` annotating each branch of `sealed trait A` like this:

```scala
sealed trait Foo
@documentation("barDoc") case object Bar extends Foo
case object Baz extends Foo

assert(Annotations[documentation, A].apply == List(Some(documenation("barDoc")), None))
```

Since `recordHead` is already `GenericRecord`, I'd rather expand `@documentation` to annotate the class itself and let `recordGeneric` handle it, which is not implemented in this PR yet.

guersam: Related discussion: https://github.com/milessabin/shapeless/issues/866#issuecomment-426913817

guersam:
Regarding sealed trait support, I'd wait for #135 for proper discriminator support and then split `def genericJsonSchema` into `def genericJsonRecord[A]` and `def genericJsonTagged[A](discriminatorName: String)`. What do you think?

julienrf:
That looks fine to me.

julienrf:
@guersam Would you be interested in finishing that work? Otherwise, maybe @jchapuis you would like to help?

julienrf:
BTW, what do you think of using `@docs` instead of `@documentation`? (just for the sake of conciseness)

guersam:
`@docs` looks good to me. I still want to finish this work but am suffering from lack of free time these days, so I'll be really thankful if there's someone who can drive it to the goal.

julienrf:
@guersam Thanks for your work, I’ve included it in #393, I’m closing this PR.

PR #129, opened 2018-10-04
Rewrite the “Overview” documentation page

julienrf:
I think the previous version was going a little bit too much into irrelevant technical details (like how to configure the build). Also, we were not demonstrating OpenAPI documentation generation.

I’ve also inverted the order of the “Use cases” and “Overview” pages: if readers follow the left menu, I think it makes more sense to start with the “Use case” page, which gives a really high-level view of what *endpoints* does, and then go to the “Overview” page, which shows more precisely what the programming experience look like.

I think I should eventually give more pointers on how to go further on some specific things (like adding more descriptions to the generated documentation, using another HTTP library than Play, implementing a custom part in an endpoint encoding/decoding), but I will first overhaul more parts of the documentation (in a separate PR).

### DIFF
--- a/documentation/manual/src/ornate/overview.md
+++ b/documentation/manual/src/ornate/overview.md
@@ -1,43 +1,33 @@
 # Overview
 
-Typically, a project is broken down into several sub-projects:
+The central idea of the *endpoints* library is that you first describe your communication endpoints
+and then the library provides you:
 
-~~~ mermaid
-graph BT
-  endpoints
-  server -.-> endpoints
-  client -.-> endpoints
-~~~
-
-The `endpoints` sub-project contains the *description* of the communication
-protocol. The `server` sub-project *implements* this communication protocol.
-The `client` sub-project *uses* the protocol to communicate with the `server`.
+- a server implementation decoding requests and building responses,
+- a client implementation building requests and decoding responses,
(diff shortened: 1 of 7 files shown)


documentation/manual/src/ornate/overview.md
```diff
@@ -47,25 +37,80 @@ And also:
...
+            "content" : {
+              "application/json" : {}
+            }
+          }
+        }
+      }
+    }
+  }
+}
+~~~
```
julienrf: I wish I could generate this JSON document from our code, but that would require a bit more work. So, I’ve manually copy-pasted the JSON document.

julienrf:
Has anyone any feeling about this PR?

julienrf:
> One thing that could be easily improved (and would love to see in this PR) is the information on how to add the lib to the project. `Installation` chapter could be either embedded in `Introduction` or placed as a second one on the list. Or we could add some very basic thing to `Introduction` and link to the `Installation` from there. Nevertheless, when I google for a library I know about already, I like to get sbt installation snippet with as little effort as possible.

In our case, the hard part is that we have quite a lot of artifacts. I’m not sure what’s the best way to present them. I think it is useful to list them exhaustively somewhere, but that’s long (and painful to maintain).

Issue #134, opened 2018-10-08
Support for discriminator in `Tagged`


guersam:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#discriminator-object

Many real world json schema involves a discriminator field such as `type` to implement sum types. As it affects both serialization and documentation, it would be nice to have it in the core algebra.

krzemin:
I'm totally in!

I tried from the different perspective, here comes my summary:
* few days ago I was trying code generation through https://github.com/openapitools/openapi-generator from endpoints-generated OpenApi json
* everything worked well, except sealed trait hierarchies which it seems was not supported using our encoding
* on a branch, I added support for discriminator instead our existing encoding for tagged instances - it didn't help openapi-generator as it seems they don't properly support `oneOf` yet
* I had to change openapi schema for instances, adding referenced base type (using `allOf`) - only then the code generator worked well, also for sum types!

As @guersam noticed, this will affect also circe/play serialization. When I align them, I'll issue a WiP PR so that we can discuss few more details there.

guersam:
Great to know you're already working on it! Looking forward to the PR :)

PR #135, opened 2018-10-08
Support for OAS3 discriminator

krzemin:
This PR addresses #134:

- added `discriminatorName` to `JsonSchemas` algebra
  * do we need more fine-grained granularity of specifying discriminator name (i.e. per union type) or is it just fine?
  * does default value (`"type"`) at algebra level make sense in this case? user has ability to override it, but we don't require defining this value, especially if sum types are not used
- changed encoding of coproducts from `{ "InstanceName": { <instance encoding> }}` to `{<discriminatorName>: "InstanceName", <instance encoding>}`
   * this is now reflected in both circe and play json schema interpreters
   * such encoding is not resistant to border cases where instance case class contains field with the same name as `discriminatorName` value
- coproduct instances are now encoded using `allOf` + reference to base coproduct type
  * semantically this shouldn't change anything, but enables [openapi-generator](https://github.com/openapitools/openapi-generator) to correctly generate code for coproducts, as it [doesn't seem to support `oneOf`](https://github.com/openapitools/openapi-generator/issues/15)

- when fixing Play Json codecs, I also explicitly modeled `Record` to have `def writes: OWrites[A]` instead of `def writes: Writes[A]` which let me avoid few ugly `.asInstanceOf[JsObject]`

### DIFF
--- a/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
@@ -1,7 +1,7 @@
 package endpoints
 package circe
 
-import endpoints.algebra
+import endpoints.algebra.JsonSchemas
 import endpoints.algebra.circe.CirceCodec
 import io.circe._
 
@@ -11,7 +11,7 @@ import scala.language.higherKinds
 /**
   * An interpreter for [[endpoints.algebra.JsonSchemas]] that produces a circe codec.
   */
-trait  JsonSchemas
+trait JsonSchemas
   extends algebra.JsonSchemas {
 
   trait JsonSchema[A] {
@@ -25,30 +25,30 @@ trait  JsonSchemas
 
     implicit def toCirceCodec[A](implicit jsonSchema: JsonSchema[A]): CirceCodec[A] =
       CirceCodec.fromEncoderAndDecoder(jsonSchema.encoder, jsonSchema.decoder)
-
   }
 
   type Record[A] = JsonSchema[A]
 
   trait Tagged[A] extends Record[A] {
+    def discriminator: String = JsonSchemas.defaultDiscriminatorName
     def taggedEncoded(a: A): (String, Json)
     def taggedDecoder(tag: String): Option[Decoder[A]]
     def encoder: Encoder[A] =
       Encoder.instance { a =>
         val (tag, json) = taggedEncoded(a)
-        json.deepMerge(Json.obj(discriminatorName -> Json.fromString(tag)))
+        json.deepMerge(Json.obj(discriminator -> Json.fromString(tag)))
       }
     def decoder: Decoder[A] =
       Decoder.instance { cursor =>
         cursor.as[JsonObject].right.flatMap { jsonObject =>
-          jsonObject(discriminatorName).flatMap(_.asString) match {
+          jsonObject(discriminator).flatMap(_.asString) match {
             case Some(tag) =>
               taggedDecoder(tag) match {
                 case Some(dec) => dec.decodeJson(cursor.value)
                 case None => Left(DecodingFailure(s"No decoder for discriminator '$tag'!", Nil))
               }
             case None =>
-              Left(DecodingFailure(s"Missing type discriminator field '$discriminatorName'!", Nil))
+              Left(DecodingFailure(s"Missing type discriminator field '$discriminator'!", Nil))
           }
         }
       }
@@ -79,6 +79,13 @@ trait  JsonSchemas
     new Tagged[A] {
       def taggedEncoded(a: A) = (tag, recordA.encoder.apply(a))
       def taggedDecoder(tagName: String) = if (tag == tagName) Some(recordA.decoder) else None
+    }
+
+  def withDiscriminator[A](tagged: Tagged[A], discriminatorName: String): Tagged[A] =
+    new Tagged[A] {
+      override def discriminator: String = discriminatorName
+      def taggedEncoded(a: A): (String, Json) = tagged.taggedEncoded(a)
+      def taggedDecoder(tag: String): Option[Decoder[A]] = tagged.taggedDecoder(tag)
     }
 
   def choiceTagged[A, B](taggedA: Tagged[A], taggedB: Tagged[B]): Tagged[Either[A, B]] =
@@ -133,5 +140,4 @@ trait  JsonSchemas
       io.circe.Encoder.encodeIterable[A, C](jsonSchema.encoder, implicitly),
       io.circe.Decoder.decodeIterable[A, C](jsonSchema.decoder, cbf)
     )
-
 }
--- a/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
@@ -24,18 +24,40 @@ trait JsonSchemas
 
     implicit def toCirceCodec[A](implicit jsonSchema: JsonSchema[A]): CirceCodec[A] =
       CirceCodec.fromEncoderAndDecoder(jsonSchema.encoder, jsonSchema.decoder)
+
+    implicit def toCirceEncoder[A](implicit jsonSchema: JsonSchema[A]): Encoder[A] =
+      jsonSchema.encoder
+
+    implicit def toCirceDecoder[A](implicit jsonSchema: JsonSchema[A]): Decoder[A] =
+      jsonSchema.decoder
   }
 
-  type Record[A] = JsonSchema[A]
+  trait Record[A] extends JsonSchema[A] {
+    override def encoder: ObjectEncoder[A]
+  }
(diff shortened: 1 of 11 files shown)


json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -82,6 +82,9 @@ trait JsonSchemas {
   /** The JSON schema of a record with a single optional field `name` of type `A` */
   def optField[A](name: String, documentation: Option[String] = None)(implicit tpe: JsonSchema[A]): Record[Option[A]]
 
+  /** Name of the field that acts as discriminator for sum types (coproducts) */
+  def discriminatorName: String = "type"
```
guersam: For sake of flexibility, it might be preferable to make `discriminatorName` an additional parameter of `def tagged` rather than an algebra-wise value, so that the users are able to specify different discriminator names like `pet_type` in the corresponding example in OAS3 document.
 
image

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -82,6 +82,9 @@ trait JsonSchemas {
...
+
   /** Tags a schema for type `A` with the given tag name */
   def taggedRecord[A](recordA: Record[A], tag: String): Tagged[A]
```
julienrf: What do you think of taking the `discriminatorName` as a parameter of `taggedRecord` (with a default value to `"type"`)? I don’t know if this is useful but that would allow users to have schemas with different discriminator names in a same project.

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
guersam: oops, `taggedRecord` I mean...

openapi/openapi/src/main/scala/endpoints/openapi/Endpoints.scala
```diff
@@ -106,7 +106,7 @@ trait Endpoints
     } yield recSchema
 
     allReferencedSchemas
-      .map { ref => ref.name -> ref.original }
+      .collect { case Schema.Reference(name, Some(original)) => name -> original }
```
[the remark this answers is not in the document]
krzemin: I extended `Reference` to have optional original schema - that's why I'm collecting only references with `Some(original)`.

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
krzemin: I agree it would be useful to have control on discriminator field name per concrete union type, but I don't think it should be a property of `taggedRecord`, which reflects to coproduct instance. You could have two tagged records with different discriminator names and when summing them with `choiceTagged` we would need to make arbitrary choice which discriminator name to use which would be misleading.

A bit better would be to extend `choiceTagged` signature to:
```scala
def choiceTagged[A, B](taggedA: Tagged[A], taggedB: Tagged[B], discriminator: String = "type"): Tagged[Either[A, B]]
```
Not perfect, but at least we would have clear semantics: discriminator from the outermost `choiceTagged` is used.

Alternatively, we could introduce another combinator to the algebra, something similar to already existing `named`:

```scala
def withDiscriminator[A](tagged: Tagged[A], discriminator: String): Tagged[A]
```
which could be used for providing own discriminator field name (by default we would use just `"type"`).

WDYT?

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
krzemin: See https://github.com/julienrf/endpoints/pull/135#discussion_r223656666 which concerns the same issue.

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
julienrf: > Alternatively, we could introduce another combinator to the algebra, something similar to already existing `named`:
> 
> ```scala
> def withDiscriminator[A](tagged: Tagged[A], discriminator: String): Tagged[A]
> ```
> 
> which could be used for providing own discriminator field name (by default we would use just `"type"`).

Indeed, the `taggedRecord` method is not the right place to put this information. Your solution seems better.

krzemin:
> I need to have a look at the OAS to understand why the encoding you propose is correct (it could be something specific to the generator you use).

Basically, `allOf` enables us to model inheritance hierarchies in OpenApi spec. There is [example in specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#models-with-polymorphism-support) which encodes simple hierarchy (Pet/Cat/Dog) in this way.

> Other than that, do you think it would be possible to take as parameter the encoding of the alternative (to support both the current encoding and the encoding you propose, and users could choose between each one)?

I think it's crucial to distinguish the styles of coproduct encoding first:
a) `oneOf` + discriminator
b) `allOf` with reference to base type + discriminator
c) both at the same time (as they don't seem to contradict each other)

So far a) without discriminator, but with inline object schema was used. This PR currently aims to model c).

What's specific to generator I'm using is that it doesn't properly support a)-only style.

I'm OK with giving user a choice, if we find it valuable.
I'm also OK with leaving only a) in `endpoints` as the most canonical encoding. Then I could lift to b)/c) by post-processing in my project only or wait for https://github.com/openapitools/openapi-generator/issues/15 is implemented.

guersam:
If we have to choose one, I'd prefer b) or c) as many popular languages including Scala 2 rely on subtype polymorphism and lack proper support for union type.

Adding `def withDiscriminator` to the core algebra seems fine to me because of the fundamental difference of `oneOf` and `allOf` as @krzemin pointed out.

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
krzemin: Implemented ✅

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -145,3 +145,8 @@ trait JsonSchemas {
   ): JsonSchema[C[A]]
 
 }
+
+object JsonSchemas {
+
+  val defaultDiscriminatorName: String = "type"
```
guersam: Given the scope of `Tagged`, it would be fine to get this back into `trait JsonSchemas` with current name so that the user can set algebra-wise default name as well as each schema specific value.

guersam: (sorry for somewhat inconsistent feedback, I had no idea of the default name encoding at first.)

krzemin:
@julienrf any preference how do we want to proceed with coproduct encoding in `openapi`? a/b/c? give user a choice or support only single one?

json-schema/json-schema-playjson/src/main/scala/endpoints/playjson/JsonSchemas.scala
```diff
@@ -19,40 +19,46 @@ trait JsonSchemas
     def writes: Writes[A]
   }
 
+  trait Record[A] extends JsonSchema[A] {
+    override def writes: OWrites[A]
+  }
```
julienrf: :+1:

json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
```diff
@@ -25,29 +24,30 @@ trait  JsonSchemas
...
 
   trait Tagged[A] extends Record[A] {
+    def discriminator: String = defaultDiscriminatorName
     def taggedEncoded(a: A): (String, Json)
     def taggedDecoder(tag: String): Option[Decoder[A]]
     def encoder: Encoder[A] =
       Encoder.instance { a =>
         val (tag, json) = taggedEncoded(a)
-        Json.obj(tag -> json)
+        json.deepMerge(Json.obj(discriminator -> Json.fromString(tag)))
```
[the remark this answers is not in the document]
krzemin: reworked this and unified representation for Records so that encoder is now `ObjectEncoder` - pretty similar to what is in play interpreter

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -85,6 +85,12 @@ trait JsonSchemas {
   /** Tags a schema for type `A` with the given tag name */
   def taggedRecord[A](recordA: Record[A], tag: String): Tagged[A]
 
+  /** Default discriminator field name for sum types */
```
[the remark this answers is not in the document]
krzemin: updated scaladoc ✅

julienrf:
Awesome, thanks!

Issue #132, opened 2018-10-08
Support for `format` property of OpenAPI primitive types


guersam:
OpenAPI specifies `format` property in addition to `type` [1]. There are a few primitive types such as date or dateTime that have different `format`s, with the same `string` type.

It would be nice to add additional `format: Option[String]` member to `endpoints.openapi.model.Schema.Primitive` so that we can increase openapi spec coverage.

[1] https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#data-types

julienrf:
Fixed by #136

PR #136, opened 2018-10-09
#132 Add `format` property to OpenAPI primitive schema

guersam:
This commit adds preliminary support for OpenAPI `format` property [1]
with a few default values for `Int`, `Long` and `Double`.

[1] https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md#data-types

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
@@ -77,6 +77,8 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   lazy val bigdecimalJsonSchema: DocumentedJsonSchema = Primitive("number")
 
+  lazy val floatJsonSchema: DocumentedJsonSchema = Primitive("number", format = Some("float"))
+
   lazy val doubleJsonSchema: DocumentedJsonSchema = Primitive("number", format = Some("double"))
 
   lazy val booleanJsonSchema: DocumentedJsonSchema = Primitive("boolean")
--- a/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
@@ -28,7 +28,7 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
     case class DocumentedCoProd(alternatives: List[(String, DocumentedRecord)],
                                 name: Option[String] = None) extends DocumentedJsonSchema
 
-    case class Primitive(name: String) extends DocumentedJsonSchema
+    case class Primitive(name: String, format: Option[String] = None) extends DocumentedJsonSchema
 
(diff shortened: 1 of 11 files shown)


guersam:
Users still have to define each implementation and documentation schema by themselves, but it must be better than nothing.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -71,13 +71,13 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
...
-  lazy val intJsonSchema: DocumentedJsonSchema = Primitive("integer")
+  lazy val intJsonSchema: DocumentedJsonSchema = Primitive("integer", format = Some("int32"))
 
-  lazy val longJsonSchema: DocumentedJsonSchema = Primitive("integer")
+  lazy val longJsonSchema: DocumentedJsonSchema = Primitive("integer", format = Some("int64"))
 
   lazy val bigdecimalJsonSchema: DocumentedJsonSchema = Primitive("number")
 
-  lazy val doubleJsonSchema: DocumentedJsonSchema = Primitive("number")
+  lazy val doubleJsonSchema: DocumentedJsonSchema = Primitive("number", format = Some("double"))
```
krzemin: What do you think about adding `floatJsonSchema`, following the spec?
https://swagger.io/docs/specification/data-models/data-types/#numbers

guersam: That must be harmless, I'll add it. Thanks for pointing out!

guersam:
Following questions::

1. How to unite the marshalling and the documentation of custom formats?

      Or do we need it at all?

2.  How to deal with other well-known formats, especially date(Time)?

      Not to involve 3rd party lib too early, I'd suggest factor the default json schemas out into something like `trait DefaultJsonSchemas` extending `ScalaVersionSpecificDefaultJsonSchemas`, that contains default `java.time` instances in Scala 2.12+ and remains empty in Scala 2.11. 

      Alternatively we can make `endpoints-java8` module, but unlike Circe's case we need `-java8` module for each interpreter.

      And we also need to determine if we'll make those additional instances implicit or not. Personally I'd prefer the former because even if it's implicit, the users can override it when they want another datetime representation such as unix timestamp. But no strong opinion here.

julienrf:
>     1. How to unite the marshalling and the documentation of custom formats?

Do you have a specific example in mind? If the operations exposed at the algebra level do not guarantee that documentation and marshaling are consistent, then the solution for custom formats would be to introduce new vocabulary at the algebra level and consistently implement their interpreters.

>     2. How to deal with other well-known formats, especially date(Time)?

We should add a `dateJsonSchema: JsonSchema[Date]` and a `dateTimeJsonSchema: JsonSchema[DateTime]` methods to the algebras. We could make this module optional (in a separate trait) or maybe ask users to use the [threeten backport](https://www.threeten.org/threetenbp/) (I’ve never used it, I don’t know how compatible/reliable it is, that’s just an idea).

PR #139, opened 2018-10-10
Make it possible to write explicit result types of endpoint constructors

julienrf:
Follow-up of our [discussion on gitter](https://gitter.im/julienrf/endpoints?at=5bbdcb92ef4afc4f284b7a0f).

It seems that by adding an additional type parameter (instead of using a path-dependent type), the compiler happily accepts our type annotations.


julienrf:
I guess I can just go ahead and merge it :smile:

Krever:
Awesome!

Issue #131, opened 2018-10-07
Documentation overhaul


julienrf:
I think it’s time to redesign the documentation. It has evolved to a state where things are too much spread out. In particular:

- I don’t think it is anymore relevant to document how to produce OpenAPI docs in a separate “tutorial” page,
- I don’t think this idea of using tutorials is that good, after all. I initially wanted to have a kind of reference documentation that users could read like a “story”, progressively introducing all our features, in the context of a single example project. But we have too many features, which makes it hard to explain all of them in the context of the same example project.

Also, I think the documentation should cover the (currently missing) following topics:

- how to extend the algebra interfaces with application-specific needs (such as authentication),
- how to properly mix the JSON partial interpreters.

IMHO the following existing pages correctly do their job:
- **Introduction**: explain the purpose of the library in one sentence, highlight its main features, and provide pointers to the other documentation pages,
- **Use cases**: showcase the salient features of the library, without going into technical details,
- **Overview**: show the programming experience of the library,
- **Installation**: concrete steps to add *endpoints* to your project build,
- **Design in a nutshell**: explain our tagless-final based design,
- **Comparison with similar tools**: highlight the differences with similar tools,
- **Talks and publications**: videos or articles complementing the docs.

But the most important pages, namely the tutorials and the “Algebra interfaces and their interpreters” page, are not easy to read or are incomplete or out of date. I’m not sure yet how we should organize their content. The goal of these pages is to serve as a reference listing (almost) all the features provided by the different algebra interfaces.

Ideas:

- Show a diagram of the various algebra interfaces and their relationships
- Write the reference documentation for a given algebra interface in its API docs

julienrf:
Note that the existing scaladoc already provides a diagram of the hierarchy of algebra interfaces:

![algebra-interfaces](https://user-images.githubusercontent.com/332812/46580995-f8213680-ca2f-11e8-8e2e-81503a4aac6c.png)

PR #141, opened 2018-10-11
Reference documentation overhaul

julienrf:
As a first step to address #131, this PR replaces the tutorials by “reference” documentation pages for each algebra and for each family (Play, Akka, etc.) of interpreters. The “Overview” and “Installation” pages have also been merged into a single “Quick start” page.

BTW, I didn’t follow the idea that I suggested in #131, consisting in documenting algebras or interpreters in their scaladoc. I prefer the approach that consists in inserting code snippets into .md files, because this allows us to show only the relevant parts of the code and to not show the boring “setup” code (which can be seen if people click on the link to the source file).

Let me know what you think.


julienrf:
I’ve tried to add tests for the added Scala.js interpreter, but it seems that the combination of jsdom and scoverage doesn’t work. Or maybe it’s something else, but that doesn’t work. I think the cost of supporting these tests is higher than the benefits they provide, therefore I will exclude the last commit from the PR.

Issue #142, opened 2018-10-11
Add `put` and `delete` shorthand methods for defining PUT and DELETE requests


julienrf:
Like we have `get` and `post`, in `endpoints.algebra.Endpoints`, we should have `put` and `delete`.

PR #146, opened 2018-10-30
Adding put and delete shorthand methods for UPDATE and DELETE Requests

stankoua:
Linked to the issue #142 
It need the PR for tests on Requests.


julienrf:
Maybe let’s wait for #117 to be merged and then add tests for `post`, `put` and `delete` methods?

algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
```diff
@@ -89,5 +89,27 @@ trait Requests extends Urls with Methods with InvariantFunctorSyntax with Semigr
...
   )(implicit tuplerUB: Tupler.Aux[UrlP, BodyP, UrlAndBodyPTupled], tuplerUBH: Tupler.Aux[UrlAndBodyPTupled, HeadersP, Out]): Request[Out] = request(Post, url, entity, headers)
 
+  /**
+    * Helper method to perform UPDATE request
+    * @tparam UrlP Payload carried by url
+    * @tparam BodyP Payload carried by body
+    * @tparam HeadersP Payload carried by headers
+    * @tparam UrlAndBodyPTupled Payloads of Url and Body tupled together by [[Tupler]]
+    */
+  final def update[UrlP, BodyP, HeadersP, UrlAndBodyPTupled, Out](
```
julienrf: Actually, since the name of the HTTP verb is “PUT” we should name this method `put` as well, instead of `update`.

PR #148, opened 2018-10-30
Renamed the `option` response combinator to `wheneverFound`

### LINKED ISSUE #140: Rename the `option` response combinator to `wheneverFound`
404 response status codes are actually handled by the `option` response combinator, that turns a `Response[A]` into a `Response[Option[A]]` (`None` being mapped to a 404 response).

However, the `Option` type can be used for other means, in responses (for instance, we also use it for authenticated endpoints, to mean that the request was not authenticated).

I think that we should rename the `option` combinator to `wheneverFound`, so that it conveys better its meaning:

~~~ scala
// currently
val userById =
  endpoint(get(path / "user" / segment[Long]), option(jsonResponse[User]()))

// with the proposed change
val userById =
  endpoint(get(path / "user" / segment[Long]), wheneverFound(jsonResponse[User]()))
~~~

NK91:
Fixes #140

### DIFF
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
@@ -108,7 +108,7 @@ class Endpoints(val settings: EndpointsSettings)
       Future.failed(new Throwable(s"Unexpected status code: ${x.status.intValue()}"))
     }
 
-  override def option[A](inner: HttpResponse => Future[Either[Throwable, A]], notFoundDocs: Documentation): HttpResponse => Future[Either[Throwable, Option[A]]] = {
+  override def wheneverFound[A](inner: HttpResponse => Future[Either[Throwable, A]], notFoundDocs: Documentation): HttpResponse => Future[Either[Throwable, Option[A]]] = {
     {
       case resp if resp.status.intValue() == 404 => Future.successful(Right(None))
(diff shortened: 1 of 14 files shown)


julienrf:
Thanks @NK91!

As I explained in the corresponding issue (#140), I prefer “wheneverFound” over “option”, so I’m in favor of merging this PR. @Krever proposed having a `.orNotFound` operation. I think that would be fine only in addition to this one (as a nicer syntactic sugar), but I wouldn’t only have `.orNotFound` because extension methods don’t generally provide a nice dev experience. WDYT?

NK91:
I think it would be great to have something like that:
```scala
import endpoints.syntax._ // or mix trait for instance `ResponseSyntax`
...
val userById =
  endpoint(get(path / "user" / segment[Long]), jsonResponse[User]().wheneverFound())
...

```
WDYT?

julienrf:
I have a preference for `.orNotFound` over `.wheneverFound`, because it uses the same words as the status code description.
We can include the support for the “syntax” in the `Endpoints` trait (no need to import or mix an additional trait).

NK91:
You mean in the trait `endpoints.algebra.Responses`:
```scala
/**
  * Extension syntax to the response.
  */
implicit class ResponseExtension[A](response: Response[A]) {
    /**
      * syntax for `wheneverFound`
      */
    def orNotFound(notFoundDocs: Documentation = None): Response[Option[A]] = wheneverFound(response, notFoundDocs)
}
```
?

PR #187, opened 2018-11-20
Fixes #186

### LINKED ISSUE #186: Add support for JsonSchema[UUID], Segment[UUID] and QueryStringParam[UUID]
It is rather inconvenient to have to write this in an endpoint definition:

```
trait MyEndpoints 
  extends algebra.Endpoints
  with algebra.JsonSchemaEntities
  with generic.JsonSchemas {

 // some endpoint with a segment[UUID] 
 // some endpoint with a jsonResponse[X] or jsonRequest[X] where X has a field of type UUID.

 implicit def uuidSegment: Segment[UUID]
 implicit def uuidJsonSchema: JsonSchema[UUID]
}
```

and then implement these implicit definitions everywhere.

There is a legitimate concern about avoiding a proliferation of cases for many data types;
however, I believe this special case  is reasonable given the pervasiveness of UUIDs and its support in the technologies used in the current algebra implementations.

NicolasRouquette:


documentation/examples/cqrs/web-client/src/main/scala/cqrs/webclient/PublicEndpoints.scala
```diff
@@ -1,17 +1,8 @@
...
 
 object PublicEndpoints
   extends cqrs.publicserver.PublicEndpoints
     with xhr.faithful.Endpoints
-    with xhr.JsonEntitiesFromCodec {
-
-  //#segment-uuid
-  implicit lazy val uuidSegment: Segment[UUID] =
-    (uuid: UUID) => stringSegment.encode(uuid.toString)
-  //#segment-uuid
```
julienrf: Same for this snippet.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -81,6 +81,8 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   def invmapJsonSchema[A, B](jsonSchema: DocumentedJsonSchema, f: A => B, g: B => A): DocumentedJsonSchema = jsonSchema
 
+  lazy val uuidJsonSchema: DocumentedJsonSchema = Primitive("uuid")
```
julienrf: I think it should be `Primitive("string", format = Some("uuid"))` instead?

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -212,6 +212,7 @@ object Schema {
       s"#/components/schemas/$name"
   }
 
+  val simpleUUID = Primitive("uuid", None)
```
julienrf: Same here, it should be `Primitive("string", format = Some("uuid"))`.

akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala
```diff
@@ -33,6 +35,8 @@ trait Urls extends algebra.Urls {
   // Query strings
   //***************
 
+  implicit lazy val uuidQueryString: QueryStringParam[UUID] = Unmarshaller.strict[String, UUID](UUID.fromString)
```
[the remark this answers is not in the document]
NicolasRouquette: Yes, this could be rewritten like this:

```
  implicit lazy val uuidQueryString: QueryStringParam[UUID] = Unmarshaller(
    implicit ec =>
      (s: String) =>
        Try(Future(UUID.fromString(s))).fold(Future.failed, identity))
```

documentation/examples/cqrs/public-server/src/main/scala/cqrs/publicserver/PublicServer.scala
```diff
@@ -79,14 +79,6 @@ class PublicServer(
 
     )
 
-  //#segment-uuid
-  implicit lazy val uuidSegment: Segment[UUID] =
-    new Segment[UUID] {
-      def decode(segment: String): Option[UUID] = Try(UUID.fromString(segment)).toOption
-      def encode(uuid: UUID): String = URLEncoder.encode(uuid.toString, utf8Name)
-    }
-  //#segment-uuid
```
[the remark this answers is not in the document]
NicolasRouquette: hum, ok, i'll check on my next update.

NicolasRouquette:
Regarding your suggestion w.r.t: 
```
def refineSegment[A, B](sa: Segment[A])(f: A => Option[B])(g: B => A): Segment[B]
def refineQueryStringParam[A, B](pa: QueryStringParam[A])(f: A => Option[B])(g: B => A): QueryStringParam[B]
def refineJsonSchema[A, B](sa: JsonSchema[A])(f: A => Option[B])(g: B => A): JsonSchema[B]
```

I agree with the direction of your suggestion; however, I've been puzzled by the return type of the forward conversion, `Option[B]`. I was wondering whether you were thinking of using `None` to report a conversion failure. But then, your question about signaling failure to the Unmarshaller led me to wonder about reusing your `InvariantFunctor` machinery where conversion failures would be reported as exceptions as is done elsewhere.

For `algebra.Urls`, this could look like this:

```
  type QueryStringParam[A]

  implicit def queryStringParamInvFunctor: InvariantFunctor[QueryStringParam]

  /** Ability to define `UUID` query string parameters */
  implicit def uuidQueryString: QueryStringParam[UUID]

  /** Ability to define `String` query string parameters */
  implicit def stringQueryString: QueryStringParam[String]

  /** Ability to define `Int` query string parameters */
  implicit def intQueryString: QueryStringParam[Int]

  /** Query string parameter containing a `Long` value */
  implicit def longQueryString: QueryStringParam[Long]
```


```
  type Segment[A]

  implicit def segmentInvFunctor: InvariantFunctor[Segment]

  /** Ability to define `UUID` path segments */
  implicit def uuidSegment: Segment[UUID]

  /** Ability to define `String` path segments */
  implicit def stringSegment: Segment[String]

  /** Ability to define `Int` path segments */
  implicit def intSegment: Segment[Int]

  /** Segment containing a `Long` value */
  implicit def longSegment: Segment[Long]
```

Except for `stringQueryString` and `stringSegment`, all other
definitions could be defined in terms of `queryStringParamInvFunctor` and `segmentInvFunctor` respectively. 

I'll experiment with this idea to see how it goes...

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
NicolasRouquette: Ah, yes; thanks.

julienrf:
>  I've been puzzled by the return type of the forward conversion, `Option[B]`. I was wondering whether you were thinking of using `None` to report a conversion failure.

Yes, that’s what I intended.

> But then, your question about signaling failure to the Unmarshaller led me to wonder about reusing your `InvariantFunctor` machinery where conversion failures would be reported as exceptions as is done elsewhere.

There are two types of failures:

1. invalid content and malformed parameters should be explicit in the types (e.g., by using `Try`, `Either`, `Option`, `Future`, etc.)
2. unexpected errors (e.g., bugs in the implementation of an interpreter) should use exceptions.

So, we should not throw exceptions in the mapping functions passed to an `InvariantFunctor`, and we should not have existing code relying on that.

NicolasRouquette:
Yes but `openapi.Urls` is the interpreter layer.

On closer look, I realized that we can use the same trick than what you've used for defining `urlInvFunctor` for defining the `refineXxx` methods there. Since `Segment[A] == Segment[B]` in that interpreter, `refineSegment` is the identity of its 1st argument.

julienrf:
Sorry, I thought you were talking about the algebra level. At the interpreter level, yes these methods are implementable because all the types are known. The solution you pushed is the right one :)

algebras/algebra/src/main/scala/endpoints/algebra/Urls.scala
```diff
@@ -73,6 +75,11 @@ trait Urls {
     */
   type QueryStringParam[A]
 
+  def refineQueryStringParam[A, B](pa: QueryStringParam[A])(f: A => Option[B])(g: B => A): QueryStringParam[B]
+
+  /** Ability to define `UUID` query string parameters */
+  implicit def uuidQueryString: QueryStringParam[UUID]
```
julienrf: You should now be able to implement `uuidQueryString` here, instead of reimplementing it in each interpreter ;)

algebras/algebra/src/main/scala/endpoints/algebra/Urls.scala
```diff
@@ -73,6 +75,11 @@ trait Urls {
     */
   type QueryStringParam[A]
 
+  def refineQueryStringParam[A, B](pa: QueryStringParam[A])(f: A => Option[B])(g: B => A): QueryStringParam[B]
```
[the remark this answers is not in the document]
NicolasRouquette: Done.

algebras/algebra/src/main/scala/endpoints/algebra/Urls.scala
NicolasRouquette: Right, I removed most of them except for thost in openapi.Urls and uuidSegment in server.Urls.

NicolasRouquette:
Sorry about the formatting changes. We should configure scalafmt for this project to avoid these problems.

akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala
```diff
@@ -60,6 +69,8 @@ trait Urls extends algebra.Urls {
   // Paths
   // ********
 
+  override implicit def uuidSegment: Segment[UUID] = JavaUUID
```
julienrf: Are we sure that this behaves like the default (inherited) implementation?

julienrf: In particular, the current server implementation assumes that a segment with the wrong format (e.g. a string which is not a valid UUID) does not match the path description.

NicolasRouquette: I think the two are equivalent.

In `algebra.Urls`, we have:

```
  implicit def uuidSegment: Segment[UUID] =
    refineSegment[String, UUID](stringSegment)((s: String) => Try.apply(UUID.fromString(s)).toOption)(_.toString)
```

which in the context of `server.Urls`, we have:

```
  def refineSegment[A, B](sa: Segment[A])(f: A => Option[B])(g: B => A): Segment[B] =
    sa.tflatMap[Tuple1[B]]((a: Tuple1[A]) => f(a._1).map(Tuple1.apply))
```

So this becomes:

```
  implicit def uuidSegment: Segment[UUID] =
stringSegment.tflatMap[Tuple1[UUID]](a: Tuple1[String]) => Try.apply(UUID.fromString(a._1)).toOption)(_.toString)).map(Tuple1.apply))
```

Compare this with:

```
  val JavaUUID: PathMatcher1[UUID] =
    PathMatcher("""[\da-fA-F]{8}-[\da-fA-F]{4}-[\da-fA-F]{4}-[\da-fA-F]{4}-[\da-fA-F]{12}""".r) flatMap { string ⇒
      try Some(UUID.fromString(string))
      catch { case _: IllegalArgumentException ⇒ None }
    }
```

If anything, the `JavaUUUID` version does redundant parsing of the UUID string once via the regex `PathMatcher` and if that succeeds, then with `UUID.fromString(_)`. In the refine case, the UUID string is parsed only once in `UUID.fromString(_)`. 

I'll delete the redundant override.

akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala
```diff
@@ -22,8 +24,13 @@ trait Urls extends algebra.Urls {
 
   type QueryStringParam[T] = FromStringUnmarshaller[T]
 
+  def refineQueryStringParam[A, B](pa: QueryStringParam[A])(f: A => Option[B])(g: B => A): QueryStringParam[B] =
+    pa.flatMap(ec => _ => (a: A) => f(a).fold[Future[B]](Future.failed(Unmarshaller.NoContentException))((b: B) => Future.successful(b)))
```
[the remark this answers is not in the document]
NicolasRouquette: Yes, I noticed this difference as well; however, look at what the current primitives do in `akkahttp.server.Urls`; e.g. `intQueryString`, `longQueryString`, which call:

```
  private def numberUnmarshaller[T](f: String ⇒ T, target: String): Unmarshaller[String, T] =
    Unmarshaller.strict[String, T] { string ⇒
      try f(string)
      catch numberFormatError(string, target)
    }

  private def numberFormatError(value: String, target: String): PartialFunction[Throwable, Nothing] = {
    case e: NumberFormatException ⇒
      throw if (value.isEmpty) Unmarshaller.NoContentException else new IllegalArgumentException(s"'$value' is not a valid $target value", e)
  }
```

So, for `refineQueryStringParam`, it seemed to me sensible to mimic the strategy used for these primitive cases; i.e., throw `Unmarshaller.NoContentException`.

I agree with you that the difference in behavior w.r.t. `play.server` is puzzling; however, this seems to me a topic for a separate issue than that of providing support for `UUID` w/ a refinement-based strategy.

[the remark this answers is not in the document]
julienrf: I agree

Issue #138, opened 2018-10-10
Support for OAS3 security scheme


guersam:
reference: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#securitySchemeObject

https://github.com/OpenAPITools/openapi-generator adds authentication methods in client implementations only if it's defined in the spec. Even if we don't provide support for server impl, a simple model change might be helpful in this case.

stefanholzmueller:
Just adding it to the openapi model sounds good to me. Similar to the `Info` case class, supplied directly by the user?

But don't endpoint also need to reference the security schema as their requirement? https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-requirement-object

PR #192, opened 2018-11-28
Support for OAS3 security schemes

krzemin:
This is preview PR addressing proposal #138.

The idea is that openapi interpreters for specific authentication method algebras might specify security schemes for each operation. Then openapi generator captures all such schemes and put them in `components/securitySchemes` section, referring to them by name (this is similar to what was implemented for referenced schemas in #113).

Regarding [Security Scheme Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#securitySchemeObject), I defined only fields that are required to support `http` and `apiKey` methods. I believe if we agree on final approach, remaining fields (for oauth2/openid) might be added in a follow-up PR.

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
@@ -50,14 +50,18 @@ object PathItem {
 
 }
 
-case class Components(schemas: Map[String, Schema])
+case class Components(schemas: Map[String, Schema],
+                      securitySchemes: Map[String, SecurityScheme])
 
 object Components {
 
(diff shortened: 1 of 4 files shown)


openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -104,6 +110,17 @@ object Operation {
 
 }
 
+case class OperationSecurity(name: String,
```
julienrf: What do you think of naming this class `SecurityRequirement` instead, to follow the naming convention of the OAS?

krzemin: Good idea 👍

PR #194, opened 2018-11-29
Add JsonSchemaEntities for sttp client

krzemin:


Krever:
Do you think it would be possible not to rely on particular json implementation (`circe`) but allow user to use sttp client with whatever json lib they already have?

sttp/client/src/main/scala/endpoints/sttp/client/JsonSchemaEntities.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def jsonRequest[A : JsonRequest](docs: Documentation = None): RequestEntity[A] = { (a, request) =>
+    request
+      .body(JsonSchema.toCirceEncoder[A].apply(a).noSpaces)
+      .contentType("application/json")
+  }
+
+  def jsonResponse[A : JsonResponse](docs: Documentation = None): Response[A] = new SttpResponse[A] {
+    override type ReceivedBody = Either[Exception, A]
+    override def responseAs = sttp.asString.map { str =>
```
julienrf: What happens if the response entity is binary?

sttp/build.sbt
```diff
@@ -14,4 +15,8 @@ val `sttp-client` =
         "com.typesafe.akka" %% "akka-stream" % "2.5.18" % Test
       )
     )
-    .dependsOn(`algebra-jvm` % "compile->compile;test->test", `algebra-playjson-jvm` % "test->test")
+    .dependsOn(
+      `algebra-jvm` % "compile->compile;test->test",
+      `algebra-playjson-jvm` % "test->test",
+      `json-schema-circe-jvm`
+    )
```
julienrf: Should we introduce a `sttp-client-circe` module instead of adding a dependency to the `json-schema-circe` module here?

julienrf:
Thanks @krzemin for your initial work. I’m closing the PR due to inactivity. I’m happy to review and merge an updated version of the contribution.

PR #197, opened 2018-12-01
Add support for JSON schemas of recursive data types

julienrf:
It is now possible to refer to a currently being defined JSON schema
by wrapping it in a `lazySchema` constructor:

~~~ scala
case class Rec(next: Option[Rec])

val recSchema: JsonSchema[Rec] = (
  optField("next")(lazySchema(recSchema, "Rec"))
).invmap(Rec)(_.next)
~~~

The `json-schema-generic` module does not use this construct, therefore
it does not support recursive data types.


julienrf:
Do you have any comments? Maybe @stefanholzmueller, @Krever or @krzemin?

PR #211, opened 2019-01-03
Align akkahttp.server.JsonEntities signatures with the algebra

### LINKED ISSUE #210: Akka-Http JsonEntities signatures don't align with algebra
Looking at the implementation for [akkahttp.server.JsonEntities](https://github.com/julienrf/endpoints/blob/v0.8.0/akka-http/server/src/main/scala/endpoints/akkahttp/server/JsonEntities.scala#L21) the signatures of the (purportedly) implemented methods of the corresponding [algebra.JsonEntities](https://github.com/julienrf/endpoints/blob/master/algebras/algebra/src/main/scala/endpoints/algebra/JsonEntities.scala#L31) don't actually correspond correctly, as the documentation parameter is missing, e.g.

```scala
//from algebra.JsonEntities
def jsonRequest[A : JsonRequest](docs: Documentation = None): RequestEntity[A]
```

```scala
//from akkahttp.server.JsonEntities
def jsonRequest[A : JsonRequest]: RequestEntity[A] = Directives.entity[A](implicitly)
```

The net result is that it's not possible to easily implement an endpoint with akka-http mixin-in the  `server.JsonEntities` to the corresponding `server.Endpoint`, which results in missing implementations, and thus a compilation error

ivanopagano:
Fix #210

### DIFF
--- a/akka-http/server/src/main/scala/endpoints/akkahttp/server/JsonEntities.scala
+++ b/akka-http/server/src/main/scala/endpoints/akkahttp/server/JsonEntities.scala
@@ -18,11 +18,11 @@ trait JsonEntities extends algebra.JsonEntities with Endpoints {
 
   type JsonRequest[A] = FromRequestUnmarshaller[A]
 
-  def jsonRequest[A : JsonRequest]: RequestEntity[A] = Directives.entity[A](implicitly)
+  def jsonRequest[A : JsonRequest](docs: algebra.Documentation = None): RequestEntity[A] = Directives.entity[A](implicitly)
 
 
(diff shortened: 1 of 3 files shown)


ivanopagano:
@julienrf I hope this was what you had in mind?

julienrf:
Awesome, can you squash the two commits?

ivanopagano:
> Awesome, can you squash the two commits?

Do you want me to force push? Is there any cleaner way to do this, other than a new PR from totally different branch?

julienrf:
Sure, just force push.

PR #214, opened 2019-01-07
Implement JsonSchema[OpenApi] instance

### LINKED ISSUE #126: Rendering the OpenAPI model relies on circe
This issue arised from a discussion in #122. Quoting @julienrf:

> One issue is that we only have an encoder of the OpenAPI model for circe. We could abstract over the underlying JSON library by using a “generic” JSON AST (such as scalajson). Another solution could be to provide the OpenAPI JSON encoder in the modules tied to a specific JSON library.

Needs discussion. I have no solution in mind. It just feels odd that users have the choice between circe and Play JSON for regular endpoints, but the documentation endpoint only works with circe.

julienrf:
- Remove previous circe encoder from OpenApi model
- Remove circe dependency from `openapi` and `example-documented`
  modules
- Align circe encoding of optional fields with play-json: a `None`
  value is encoded with a missing field rather than a `null` field
- All Schemas can now have an attached “description”

Fixes #126


julienrf:
It’s simpler for me to work from my repository so I’m closing this PR in favor of #219.

PR #230, opened 2019-01-26
Update to Scala 2.13.0-M5

julienrf:
This required to update the Scala.js version too. I took this
opportunity to switch to sbtcrossproject.


julienrf:
I’m merging this PR although no interpreters do support 2.13. We can do that later… I’ve already started ([here](https://github.com/julienrf/endpoints/compare/play2.7.x)) but not all our dependencies support 2.13 so I’m stuck.

PR #252, opened 2019-03-10
Support repeated query parameters

### LINKED ISSUE #237: Support binding query string parameters to a list of values
The `qs` (and `qsOpt`) method can create a `QueryString[A]`, but in some cases it is useful to produce a `QueryString[List[A]]`. We could do it in two ways:

Either by introducing a `qsList` method in `endpoints.algebra.Requests`:

~~~ scala
def qsList[A: QueryStringParam](name: String, docs: Option[String] = None): QueryString[List[A]]
~~~

And then an URL taking several query string parameters could be defined as follows:

~~~ scala
path / "foo" /? qsList[Int]("ids")
~~~

The second solution would be to introduce the following `listQueryStringParam` combinator in `endpoints.algebra.Requests`:

~~~ scala
implicit def listQueryStringParam[A: QueryStringParam]: QueryStringParam[List[A]]
~~~

This would allow users to define an URL taking several query string parameters as follows:

~~~ scala
path / "foo" /? qs[List[Int]]("ids")
~~~

I think the second solution is preferable but it would require more changes in the way we implement interpreters.

julienrf:
Repeated query parameters can be encoded from and decoded to any
standard collection (e.g., `List`, `Set`, etc.). A query parameter
is marked as repeated when its type is a such collection:

~~~ scala
// Matches `/foo?id=1&id=2&id=3`, for instance
path / "foo" /? qs[List[Int]]("id")
~~~

I decided to also model optional query parameters with a similar
“type-directed” approach: a query parameter is marked as optional
when its type is wrapped by an `Option` (e.g., `Option[Int]`). I
deleted the `optQs` operation.

This work led me to refine the way we deal with errors when server
interpreters decode URLs. Indeed, what should be decoded by the
above URL definition for the `/foo?id=1&id=bar` URL candidate?
Note that the second occurrence of `id` is not a valid `Int`.

I decided to return an error in such a case. I’ve reviewed the
semantics that our server interpreters used to have in various
situation and made the following changes:

- when a query parameter is not successfully decoded, the server
  returns a “client error” (an HTTP response with status 400),
- when a path segment is not successfully decoded, the server
  also returns a “client error”

Very few tests did exist on that part of the code, so I added more
tests. I invite you to read them to know more precisely the new
semantics. Note that the new semantics should not affect users of
our client interpreters, which always construct well-formed requests.

This PR fixes #237


akka-http/client/src/main/scala/endpoints/akkahttp/client/Urls.scala
```diff
@@ -31,25 +34,28 @@ trait Urls extends algebra.Urls {
       }
     }
 
-  def qs[A](name: String, docs: Documentation)(implicit value: QueryStringParam[A]): QueryString[A] =
-    a => Some(s"$name=${value.apply(a)}")
+  def qs[A](name: String, docs: Documentation)(implicit param: QueryStringParam[A]): QueryString[A] =
+    a => {
+      val params = param(a)
+      if (params.isEmpty) None
+      else Some(params.map(v => s"$name=$v").mkString("&"))
```
tdroxler: Nothing wrong here, just for the style I usually find nicer to use pattern matching:
``param(a) match {
  case Nil => None
  case params => Option(params.map(v => s"$name=$v").mkString("&"))
}
```

tdroxler:
Nice I like it :+1: good job for the `qs[Option[A]]` too

PR #272, opened 2019-04-15
Add `remainingSegments` path constructor

### LINKED ISSUE #251: There is no way to define an URL having a dynamic part spanning several segments
For instance, given the `path / "foo" / segment[String]()` URL definition, an incoming request to `/foo/bar/baz` does not match it because the `segment[String]()` matches only `"bar"`, not `"bar/baz"`.

Actually, the [`assetSegments`](https://github.com/julienrf/endpoints/blob/6bbc7a8bd79d77d24a8a77356b192b5a036bbf42/algebras/algebra/src/main/scala/endpoints/algebra/Assets.scala#L29) operation in the `Assets` algebra does match path parts spanning several segments, but we miss a more general operation, in the `Urls` algebra interface, returning a `Segment[String]`.

julienrf:
- The `remainingSegments` constructor matches all the remaining segments
  of a path (fixes #251)
- The `remainingSegments` semantics can not be represented in OpenAPI,
  we model it as a *single* segment,
- Replace the `refineXxx` operations by `PartialInvariantFunctor` instances,
  which provide an equivalent `xmapPartial` operation,
- Implement `intSegment`, `longSegment` and `doubleSegment` in terms of
  `xmapPartial`


akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala
```diff
@@ -24,6 +24,18 @@ trait Urls extends algebra.Urls {
     joinDirectives(pathPrefix, Directives.pathEndOrSingleSlash.tmap(Tuple1(_)))
   )
 
+  implicit lazy val pathPartialInvariantFunctor: PartialInvariantFunctor[Path] = new PartialInvariantFunctor[Path] {
+    def xmapPartial[A, B](fa: Path[A], f: A => Option[B], g: B => A): Path[B] =
+      new Path(fa.pathPrefix.flatMap { a =>
+        f(a) match {
+          case Some(b) => Directives.pass.tmap(_ => b)
```
Krever: I think there is `Directives.provide`

PR #298, opened 2019-06-11
fixes #296 consistent syntax of xmap and xmapPartial

### LINKED ISSUE #296: Syntax of xmap and xmapPartial are not consistent
`xmap` takes two function parameters (e.g. `xmap(f, g)`), whereas `xmapPartial` has two parameter lists, taking a function each (e.g. `xmapPartial(f)(g)`).

We should use a similar syntax for both operations. I have a preference for using two parameter lists because it often helps with type inference, and because it avoids nesting braces in parens (ie, `xmap({ case (a, b) => e }, ...)` becomes `xmap { case (a, b) => e }(...)`)

fabianpage:


algebras/algebra/src/test/scala/endpoints/algebra/EndpointsTestApi.scala
```diff
@@ -60,20 +60,20 @@ trait EndpointsTestApi extends algebra.Endpoints {
     textResponse()
   )
 
-  val headers2 = header("C").xmap[Int](_.toInt, _.toString)
+  val headers2 = header("C").xmap[Int](_.toInt)(_.toString)
```
julienrf: Does it also work now if we remove the `[Int]` explicit type?

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -26,7 +26,7 @@ import scala.language.higherKinds
   *     implicit val schema: JsonSchema[User] = (
   *       field[String]("name") zip
   *       field[Int]("age")
-  *     ).xmap((User.apply _).tupled)(Function.unlift(User.unapply))
+*     ).xmap((User.apply _).tupled)(Function.unlift(User.unapply))
```
julienrf: The indentation looks wrong.

fabianpage:
Yes they can be removed.

PR #302, opened 2019-06-11
295: started with documentation

fabianpage:


julienrf:
@fabianpage I’ve given another round to this PR. I found the initial examples not compelling enough. Let me know what you think of the new examples and the accompanying documentation. Here is how it renders:

image

fabianpage:
@julienrf Looks pretty clear to me. (I only had time to read your comment, haven't read it in context of the documentation)

julienrf:
Thanks for your feedback!

PR #305, opened 2019-06-11
Add PartialInvariantFunctor implementation for QueryString

### LINKED ISSUE #294: QueryString values can not been transformed with xmap and xmapPartial
We should:
- add an abstract defintion `implicit def queryStringPartialInvFunctor: PartialInvariantFunctor[QueryString]` in `endpoints.algebra.Urls`,
- implement it in all the interpreters,
- add a test for it in `endpoints.algebra.client.EndpointsTestSuite`,
- add a test for it in `endpoints.algebra.server.EndpointsTestSuite`,
- add an example in the documentation `endpoints.algebra.EndpointsDocs`.

For instance, the example could be (and it could also be used for tests):

~~~ scala
case class Location(longitude: Double, latitude: Double)

val locationQueryString: QueryString[Location] =
  (qs[Double]("lon") & qs[Double]("lat"))
    .xmap(
      { case (lon, lat) => Location(lon, lat) },
      location => (location.longitude, location.latitude)
    )
~~~

Another one, illustrating `xmapPartial`, could be the following:

~~~ scala
sealed trait BlogId
case class BlogUuid(uuid: UUID) extends BlogId
case class BlogSlug(slug: String) extends BlogId

val blogIdQueryString: QueryString[BlogId] =
  (qs[Option[UUID]]("uuid") & qs[Option[String]]("slug"))
    .xmapPartial {
      case (maybeUuid, maybeSlug) =>
        maybeUuid.map(BlogUuid).orElse(maybeSlug.map(BlogSlug))
    } {
      case BlogUuid(uuid) => (Some(uuid), None)
      case BlogSlug(slug) => (None, Some(slug))
    }
~~~

Swalkyn:
fix #294

### DIFF
--- a/algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
+++ b/algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
@@ -241,7 +241,11 @@ trait EndpointsTestSuite[T <: EndpointsTestApi] extends ClientTestBase[T] {
 
           val locationQueryString: QueryString[Location] =
             (qs[Double]("lon") & qs[Double]("lat"))
-              .xmap({ case (lon, lat) => Location(lon, lat) })(location => (location.longitude, location.latitude))
+              .xmap { 
+                case (lon, lat) => Location(lon, lat) 
+              } {
+                location => (location.longitude, location.latitude)
+              }
           
           encodeUrl(path /? locationQueryString) (Location(12.0, 32.9)) shouldEqual "?lon=12.0&lat=32.9"
           encodeUrl(path /? locationQueryString) (Location(-12.0, 32.9)) shouldEqual "?lon=-12.0&lat=32.9"
@@ -263,10 +267,10 @@ trait EndpointsTestSuite[T <: EndpointsTestApi] extends ClientTestBase[T] {
                 case BlogSlug(slug) => (None, Some(slug))
               }
           
-          val testUUID: String = "f4b9defa-1ad8-453f-9a06-2683b8564b8d"
+          val testUUID: UUID = UUID.randomUUID()
           val testSlug: String = "test-slug"
 
-          encodeUrl(path /? blogIdQueryString) (BlogUuid(UUID.fromString(testUUID))) shouldEqual s"?uuid=$testUUID"
+          encodeUrl(path /? blogIdQueryString) (BlogUuid(testUUID)) shouldEqual s"?uuid=$testUUID"
           encodeUrl(path /? blogIdQueryString) (BlogSlug(testSlug)) shouldEqual s"?slug=$testSlug"
         }
 
--- a/algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
+++ b/algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
@@ -241,9 +241,7 @@ trait EndpointsTestSuite[T <: EndpointsTestApi] extends ClientTestBase[T] {
 
           val locationQueryString: QueryString[Location] =
             (qs[Double]("lon") & qs[Double]("lat"))
-              .xmap(
-                { case (lon, lat) => Location(lon, lat) },
-                location => (location.longitude, location.latitude))
+              .xmap({ case (lon, lat) => Location(lon, lat) })(location => (location.longitude, location.latitude))
           
           encodeUrl(path /? locationQueryString) (Location(12.0, 32.9)) shouldEqual "?lon=12.0&lat=32.9"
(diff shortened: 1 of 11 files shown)


scalaj/client/src/main/scala/endpoints/scalaj/client/Urls.scala
```diff
@@ -28,9 +28,13 @@ trait Urls extends algebra.Urls {
...
 
   class Url[A](val toReq: A => HttpRequest)
+  
+  implicit lazy val queryStringPartialInvFunctor: PartialInvariantFunctor[QueryString] = new PartialInvariantFunctor[QueryString] {
+    def xmapPartial[A, B](fa: QueryString[A], f: A => Option[B], g: B => A): QueryString[B] = fa compose g
+  }
 
   implicit lazy val queryStringParamPartialInvFunctor: PartialInvariantFunctor[QueryStringParam] = new PartialInvariantFunctor[QueryStringParam] {
-    def xmapPartial[A, B](fa: A => List[String], f: A => Option[B], g: B => A): B => List[String] = fa compose g
+    def xmapPartial[A, B](fa: QueryStringParam[A], f: A => Option[B], g: B => A): QueryStringParam[B] = fa compose g
```
Swalkyn: We noticed that `QueryStringParam` was not used in in `xmapPartial`

algebras/algebra/src/test/scala/endpoints/algebra/server/EndpointsTestSuite.scala
```diff
@@ -99,10 +101,66 @@ trait EndpointsTestSuite[T <: endpoints.algebra.Endpoints] extends ServerTestBas
...
+    "xmap query strings" should {
+
+      "xmap urls of locations" in {
+        
+
+        val locationQueryString =
+          (qs[Double]("lon") & qs[Double]("lat"))
+            .xmap[Location](
+              { case (lon, lat) => Location(lon, lat) },
+              location => (location.longitude, location.latitude))
```
julienrf: Same comment here about the syntax.

scalaj/client/src/main/scala/endpoints/scalaj/client/Urls.scala
julienrf: FYI, both are correct since `QueryStringParam[A]` is an alias for `A => List[String]` in `scalaj.client.Urls`. That being said, I prefer using the `QueryStringParam` name because it reflects better the intent. Thanks for fixing this!

algebras/algebra/src/test/scala/endpoints/algebra/server/EndpointsTestSuite.scala
```diff
@@ -99,10 +101,66 @@ trait EndpointsTestSuite[T <: endpoints.algebra.Endpoints] extends ServerTestBas
...
+          (qs[Option[UUID]]("uuid") & qs[Option[String]]("slug"))
+            .xmapPartial {
+              case (maybeUuid, maybeSlug) =>
+                maybeUuid.map[BlogId](BlogUuid).orElse(maybeSlug.map[BlogId](BlogSlug))
+            } {
+              case BlogUuid(uuid) => (Some(uuid), None)
+              case BlogSlug(slug) => (None, Some(slug))
+            }
+        
+        val testUUID: String = "f4b9defa-1ad8-453f-9a06-2683b8564b8d"
```
julienrf: Maybe it’d be easier to work with an `UUID` value from the beginning? (it would save some `UUID.fromString` calls)

~~~ scala
val testUUID: UUID = UUID.randomUUID()
~~~

algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
```diff
@@ -233,6 +233,47 @@ trait EndpointsTestSuite[T <: EndpointsTestApi] extends ClientTestBase[T] {
...
+            (qs[Option[UUID]]("uuid") & qs[Option[String]]("slug"))
+              .xmapPartial {
+                case (maybeUuid, maybeSlug) =>
+                  maybeUuid.map[BlogId](BlogUuid).orElse(maybeSlug.map[BlogId](BlogSlug))
+              } {
+                case BlogUuid(uuid) => (Some(uuid), None)
+                case BlogSlug(slug) => (None, Some(slug))
+              }
+          
+          val testUUID: String = "f4b9defa-1ad8-453f-9a06-2683b8564b8d"
```
julienrf: Here too, you probably want to work with a value of type `UUID` from the beginning.

Swalkyn:
Thank you for your comments @julienrf. I made changes according to what you wrote in the last commit. I think this is ready to be merged.

Again, thanks for your help today!

julienrf:
Thanks!!

PR #306, opened 2019-06-11
rename emptyRequest to emptyBody and document missing doc possibilities

robstoll:
We could add the possibility to document an empty body but then we
should also provide the possibility to document empty headers.
I kept it as is (no way to document an empty body or empty headers) but
mentioned that it can be documented in description of Endpoints.endpoint

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
@@ -17,6 +17,8 @@ trait Requests extends Urls with Methods with PartialInvariantFunctorSyntax with
     * be empty. Just that, from a server point of view no information will
     * be extracted from them, and from a client point of view no particular
     * headers will be built in the request.
+    *
+    * Use `description` of [[endpoints.algebra.Endpoints#endpoint]] to document empty headers.
     */
   def emptyHeaders: RequestHeaders[Unit]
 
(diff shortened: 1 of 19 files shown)


algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
```diff
@@ -37,13 +39,13 @@ trait Requests extends Urls with Methods with PartialInvariantFunctorSyntax with
   implicit def reqEntityInvFunctor: InvariantFunctor[RequestEntity]
 
   /**
-    * Empty request.
+    * Request without a body.
+    * Use `description` of [[endpoints.algebra.Endpoints#endpoint]] to document an empty body.
     */
-  //TODO rename to emptyBody and maybe make uniform with all other by adding ()
-  def emptyRequest: RequestEntity[Unit]
+  def emptyBody: RequestEntity[Unit]
```
julienrf: I’m not a fan of this change because this is inconsistent with the other constructors for `RequestEntity`, which all contain `Request` in their name (e.g. `textRequest`, `jsonRequest`, etc.).

Also, another problem is that we have to find a name that is unambiguous between requests and responses. An `emptyBody` could mean an embty request body as well as an empty response body.

robstoll: fair enough, I am going to close this pull request and open a new one for just the improved docs.

PR #311, opened 2019-06-14
[WIP] Http4s 0.20.x

jkobejs:
Here is first try of implementation of #283 

Things that are not implemented:
- tests
- client
- support other JSON libs (currently it has only circe)
- probably some other features of which I'm not currently aware 😄


http4s/server/src/main/scala/endpoints/http4s/server/Urls.scala
```diff
@@ -0,0 +1,196 @@
...
+        }
+      })
+    }
+  }
+
+  private def pathExtractor[A](path: Path[A], uri: http4s.Uri): Option[Either[ErrorResponse, A]] = {
+    val segments =
+      uri.path
+        .split("/")
+      .map(URLDecoder.decode(_, utf8Name))
```
Krever: We definietely need to reconsider scalafmt for everyone benefit. @julienrf Are you sitll against it ?

julienrf: We can try, OK!

julienrf:
Hey @jkobejs!

> Things that are not implemented:
> 
>     * tests
>     * client
>     * support other JSON libs (currently it has only circe)
>     * probably some other features of which I'm not currently aware smile

It is perfectly fine to support only circe and to not have a client interpreter ;)

I’ll review the PR ASAP!

documentation/examples/basic/http4s-server/src/main/scala/sample/Api.scala
```diff
@@ -0,0 +1,28 @@
...
+      action.implementedBy { _ =>
+        ActionResult("Action")
+      } orElse
+      actionFut.implementedByEffect { _ =>
+        IO.pure(ActionResult("Action"))
+      }
+  ) <+> HttpRoutes.of(auth.implementedBy { credentials =>
+    println(s"Authenticated request: ${credentials.username}")
+    if (Random.nextBoolean()) Some(()) else None // Randomly return a forbidden
+  })
```
julienrf: Should we add the following convenience method?

~~~ scala
def routesFromEndpoints(endpoints: PartialFunction[http4s.Request[F], F[http4s.Response[F]]]*) =
  endpoints.reduceLeft(_ orElse _)
~~~

Also, is there any difference between using `orElse` and `<+>`?

http4s/server-circe/src/main/scala/endpoints/http4s/server/circe/JsonEntitiesFromCodec.scala
```diff
@@ -0,0 +1,44 @@
...
+trait JsonEntitiesFromCodec[F[_]]
+    extends Endpoints[F]
+    with endpoints.algebra.circe.JsonEntitiesFromCodec {
+
+  def jsonRequest[A](docs: Documentation = None)(
+      implicit codec: JsonRequest[A]): RequestEntity[A] =
+    req => {
+      def transform(t: Either[DecodeFailure, Json]): Either[DecodeFailure, A] =
+        for {
+          str <- t
```
julienrf: This will not compile with 2.11. The workaround consists in adding a `.right` call (`str <- t.right`).

http4s/server-circe/src/main/scala/endpoints/http4s/server/circe/JsonEntitiesFromCodec.scala
```diff
@@ -0,0 +1,44 @@
...
+          a <- codec
+            .decode(str.toString)
```
julienrf: Why do we need to transform the `Json` value into a `String` value?

http4s/server-circe/src/main/scala/endpoints/http4s/server/circe/JsonEntitiesFromCodec.scala
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+  def jsonResponse[A](docs: Documentation = None)(
+      implicit codec: JsonResponse[A]): Response[A] =
+    a => {
+      implicit val encoder: http4s.EntityEncoder[F, A] =
+        jsonEncoderOf[F, Json].contramap(
+          a =>
+            // it should be safe call get here: A => Json => String => Either[Error, Json] => Json
+            parse(codec.encode(a)).right.get)
```
julienrf: We should not need to parse JSON when we as a server produce a response. There must be something wrong somewhere.

http4s/server/src/main/scala/endpoints/http4s/server/BasicAuthentication.scala
```diff
@@ -0,0 +1,34 @@
...
+      headers
+        .get(Authorization)
+        .map { authHeader =>
+          authHeader.credentials match {
+            case BasicCredentials(username, password) =>
+              Right(Credentials(username, password))
+            case _ => Left(badRequestResponse)
+          }
+        }
+        .getOrElse(Left(badRequestResponse))
```
julienrf: In case of failure we should return a `401` status code and a header `WWW-Authenticate`.

http4s/server/src/main/scala/endpoints/http4s/server/Urls.scala
```diff
@@ -0,0 +1,196 @@
...
+import endpoints.algebra.Documentation
+import endpoints.{PartialInvariantFunctor, Tupler, algebra}
+import org.http4s
+import org.http4s.Uri
+
+import scala.collection.compat._
+import scala.collection.mutable
+
+
+trait Urls[F[_]] extends algebra.Urls {
```
julienrf: Do we want to use an abstract type member instead of a type parameter? This would remove the burden of threading it everywhere.

~~~ scala
type Effect[A]
~~~

Also, it seems that `F` is not used here, we should probably introduce it only in `Endpoints`.

documentation/examples/basic/http4s-server/src/main/scala/sample/Api.scala
jkobejs: There is no difference between `orElse` and `<+>`. 
Convenience method looks good to me, I'll add it and update example

http4s/server-circe/src/main/scala/endpoints/http4s/server/circe/JsonEntitiesFromCodec.scala
jkobejs: My idea here was  to reuse decoder `jsonOf` from org.http4s.circe._ since it returns errors when it tries to decode invalid JSON request but I couldn't use `jsonOf[F, String]` because it requires payload to be String and not JSON object.
Because of that I'm decoding it into Circe's Json datatype and converting it to String to that I can use our `codec` to decode it to wanted type.

I'll see what can I do here to remove this intermediate step since it looks weird to me too 🙂

http4s/server-circe/src/main/scala/endpoints/http4s/server/circe/JsonEntitiesFromCodec.scala
jkobejs: Yep, idea here was similar to idea from `jsonRequest` method. I'll see how to get rid of this.

http4s/server/src/main/scala/endpoints/http4s/server/Urls.scala
jkobejs: That makes sense to me, I'll remove type parameter

jkobejs: I need effect type in `Urls` since I'm returning error response when I'm decoding uri and path segments and error response needs effect type.

julienrf:
Superseded by #489

PR #313, opened 2019-06-17
Feature/response entity

tdroxler:
As discussed face to face, let's create a PR for this so we can iterate on it and merge this :tada: 

This PR introduce `ResponseEntity`, which split the current behavior of `Response` with a entity part and the status code part. This allows us to return a response with different status codes and not only 200.

### DIFF
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/StatusCodes.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/StatusCodes.scala
@@ -13,11 +13,15 @@ trait StatusCodes extends algebra.StatusCodes {
   type StatusCode = AkkaStatusCode
 
   def OK = AkkaStatusCodes.OK
+  def Created = AkkaStatusCodes.Created
+  def Accepted = AkkaStatusCodes.Accepted
+  def NoContent = AkkaStatusCodes.NoContent
 
   def BadRequest= AkkaStatusCodes.BadRequest
-
   def Unauthorized = AkkaStatusCodes.Unauthorized
-
+  def Forbidden = AkkaStatusCodes.Forbidden
   def NotFound = AkkaStatusCodes.NotFound
+
+  def InternalServerError = AkkaStatusCodes.InternalServerError
 
 }
--- a/akka-http/client-circe/src/main/scala/endpoints/akkahttp/client/circe/JsonSchemaEntities.scala
+++ b/akka-http/client-circe/src/main/scala/endpoints/akkahttp/client/circe/JsonSchemaEntities.scala
@@ -16,15 +16,15 @@ import io.circe.parser.decode
   */
 trait JsonSchemaEntities extends algebra.JsonSchemaEntities with circe.JsonSchemas { this: Endpoints =>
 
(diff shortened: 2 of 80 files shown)


tdroxler:
Note that this is a breaking change, as we need to wrap the current responses with `ok`

Krever:
I like the direction, as responses encoding was something to be improved for a long time (#42). Thanks a lot! 

Regarding breaking change, I recently learned that scala-steward is able to automatically create PRs with scalafix migrations applied. I don't know how hard it is to write those migrations but might be worth checking out. 

I've not been following the codebase thoroughly enough to review, so @julienrf will hopefully do that, but looks good for me :)

julienrf:
The code looks good to me! Now we need volunteers to update the remaining interpreters. I can do the `play` and `xhr` ones. Anyone interested in the `scalaj` or `sttp` ones?

julienrf:
The changes introduced by this PR are not backward compatible. Should I work on a rewrite rule to automate the upgrade of existing codebases?

The changes are simple to apply manually, though. If I take the endpoint definition shown in the first page of the documentation:

~~~ scala
val increment: Endpoint[Increment, Unit] =
  endpoint(post(path / "increment", jsonRequest[Increment]()), emptyResponse())
~~~

With this PR, It becomes:

~~~ scala
val increment: Endpoint[Increment, Unit] =
  endpoint(post(path / "increment", jsonRequest[Increment]), ok(emptyResponse))
~~~

There are two changes:
- responses have to specify both the entity and status code (hence the additional `ok(...)` call)
- request and entities don’t take documentation anymore (hence the removal of parenthesis)

Documentation about requests and responses is done when calling the request or response constructor. Before:

~~~ scala
val someDocumentedResource: Endpoint[Int, String] =
  endpoint(
    get(path / "some-resource" / segment[Int]("id")),
    textResponse(docs = Some("The content of the resource"))
  )
~~~

With this PR:

~~~ scala
val someDocumentedResource: Endpoint[Int, String] =
  endpoint(
    get(path / "some-resource" / segment[Int]("id")),
    ok(textResponse, docs = Some("The content of the resource"))
  )
~~~

Krever:
I wouldn't bother with the rewrite rules unless its pretty easy to do. I never tried ...

akka-http/client/src/main/scala/endpoints/akkahttp/client/StatusCodes.scala
```diff
@@ -13,11 +13,15 @@ trait StatusCodes extends algebra.StatusCodes {
   type StatusCode = AkkaStatusCode
 
   def OK = AkkaStatusCodes.OK
+  def Created = AkkaStatusCodes.Created
```
gmarti:
```suggestion
  val Created = AkkaStatusCodes.Created
```

julienrf: There is a subtlety here. If we make it a `val` we become subject to initialization errors, in case the val is used before it is effectively initialized. So, it should be a `lazy val`. However, `lazy val`s have some overhead when they are read. Since, in this case, we are just forwarding to something else, I’ve kept a `def`: there is nothing expensive to compute, so in case the JIT does not inline the `def`, we don’t pay a huge price.

PR #314, opened 2019-06-17
Introduce `StatusCodes` abstraction

tdroxler:
Even tho this might not seems very useful now, I think when we'll have this `ResponseEntity` separation defined  in #313 , it will allows us to write something like this in the algebra:
```  
def responseWithCode(code: StatusCodes, entity: ResponseEntity): Response

def ok(entity:ResponseEntity): Response = responseWithCode(OK, entity)
def notFound(entity:ResponseEntity): Response = responseWithCode(NotFound, entity)
```
so only `responseWithCode` will have to be defined in the interpreters.

Let's of course discuss it, as it comes a bit from nowhere.


algebras/algebra/src/main/scala/endpoints/algebra/StatusCodes.scala
```diff
@@ -0,0 +1,21 @@
...
+  def OK: StatusCode
+
+  /** 4xx Client Error */
+  def Unauthorized: StatusCode
+
+  def NotFound: StatusCode
+
+  def BadRequest: StatusCode
+
+}
```
julienrf: Do we want to add `statusCodeFromInt(value: Int): StatusCode` convenience method? (I’m not sure it will be implementable in all interpreters, though)

openapi/openapi/src/main/scala/endpoints/openapi/StatusCodes.scala
```diff
@@ -0,0 +1,31 @@
...
+  * Interpreter for [[endpoints.algebra.StatusCodes]]
+  *
+  * @group interpreters
+  */
+trait StatusCodes extends endpoints.algebra.StatusCodes {
+
+  sealed trait StatusCode {
+    def value: Int
+    override def toString = value.toString
+  }
```
julienrf: I would just go with `type StatusCode = Int`. What is the benefit of adding a `StatusCode` class?

In addition, it should not be `sealed`: we really want users to be able to add their own status codes if they want to.

play/client/src/main/scala/endpoints/play/client/StatusCodes.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  type StatusCode = Int
+
+  def OK = 200
+
+  def NotFound = 404
+
+  def Unauthorized = 401
+
+  def BadRequest = 404
```
julienrf: Should be `400`

scalaj/client/src/main/scala/endpoints/scalaj/client/StatusCodes.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  type StatusCode = Int
+
+  def OK = 200
+
+  def NotFound = 404
+
+  def Unauthorized = 401
+
+  def BadRequest = 404
```
julienrf: Same here.

sttp/client/src/main/scala/endpoints/sttp/client/StatusCodes.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  type StatusCode = Int
+
+  def OK = 200
+
+  def NotFound = 404
+
+  def Unauthorized = 401
+
+  def BadRequest = 404
```
julienrf: Here too.

xhr/client/src/main/scala/endpoints/xhr/StatusCodes.scala
```diff
@@ -0,0 +1,22 @@
...
+
+  type StatusCode = Int
+
+  def OK = 200
+
+  def NotFound = 404
+
+  def Unauthorized = 401
+
+  def BadRequest = 404
```
julienrf: And here.

openapi/openapi/src/main/scala/endpoints/openapi/StatusCodes.scala
tdroxler: Indeed no benefit, probably I had an idea to had something, but can't remember what. Good catch for the `sealed`

scalaj/client/src/main/scala/endpoints/scalaj/client/StatusCodes.scala
tdroxler: copy  paste is baaaaaad

algebras/algebra/src/main/scala/endpoints/algebra/StatusCodes.scala
tdroxler: Could be added latter if we see it would much easier to add a code with it, but honestly most of the time it's either already an Int, or it's the exact same same as the algebra.

tdroxler:
The second commit was meant to be squashed #cleangithistory :wink:

julienrf:
I did squash your commits ;)

tdroxler:
oh indeed, I'm used to rebase/squash before merging, rather than squashing on merge. Thx :+1:

PR #315, opened 2019-06-17
Add support for type parameters in generic JsonSchemas

gmarti:
When using generic JsonSchemas with classes using type parameters, the generated name will be the same for all typed classes
Documentation will be overwritten and be wrong for all but one reference.

This PR try to fix this by adding type parameters in the name of the schema if there are some.

For `case class GenericFoo[A, B]` in package `com` with type parameters `GenericFoo[Int, com.foo.Bar]` generated name will be `com.GenericFoo-int_com.foo.Bar`

I had to use a Manifest, couldn't import a TypeTag... probably some cross compilation problem ?

Happy reviewing

### DIFF
--- a/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
@@ -130,7 +130,7 @@ trait JsonSchemas extends algebra.JsonSchemas {
     implicit def recordGeneric[A, R](implicit
       gen: LabelledGeneric.Aux[A, R],
       record: GenericRecord[R],
-      ct: ClassTag[A]
+      ct: Manifest[A]
     ): GenericRecord[A] =
       new GenericRecord[A] {
         def jsonSchema: Record[A] = nameSchema(record.jsonSchema.xmap[A](gen.from)(gen.to))
@@ -139,7 +139,7 @@ trait JsonSchemas extends algebra.JsonSchemas {
     implicit def taggedGeneric[A, R](implicit
       gen: LabelledGeneric.Aux[A, R],
       tagged: GenericTagged[R],
-      ct: ClassTag[A]
+      ct: Manifest[A]
     ): GenericTagged[A] =
       new GenericTagged[A] {
         def jsonSchema: Tagged[A] = nameSchema(tagged.jsonSchema.xmap[A](gen.from)(gen.to))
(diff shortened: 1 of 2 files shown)


gmarti:
About what to use in prod or not, my vision is : If you have good tests you can deploy anything :)
If something change in the library and serialization change, your tests should cover that change.

About my use case, I'm doing  :
```
 implicit lazy val schema: JsonSchema[GenericFoo[Int, Long]] = genericJsonSchema[GenericFoo[Int, Long]]
 implicit lazy val schema2: JsonSchema[GenericFoo[Int, String]] = genericJsonSchema[GenericFoo[Int, String]]
```
Everything is working fine for akka http server (at least)
Only the documentation is not working because of the name generated that is only "GenericFoo" and is not "unique" since I use GenericFoo multiple time with different parameters.

So this was only to fix the naming calculation and nothing else since everything is working fine.

the tests are running for JVM and JS right ? so we are covered when testing "Manifest" and "ClassTag" ??

julienrf:
> If you have good tests you can deploy anything

I’m sympathetic to this idea, but writing tests requires effort and writing robust tests requires a huge effort. This has to be put in the equation.

> the tests are running for JVM and JS, right? so we are covered when testing "Manifest" and "ClassTag"?

Yes, tests are running on both JVM and JS. But, they might not cover all the cases… I’ve seen a few cases (although I don’t remember them off the top of my head) where `ClassTag#getSimpleName` returned something different on JVM and JS. Also, tests covering communication between JVM and JS are sparse.

Actually, I think I was wrong here:

> if the class is polymorphic maybe the schema should be polymorphic as well? At deserialization-time your codecs won’t be able to distinguish between two instantiations of a type parameter of your class, right?

I think it doesn’t make sense to have a polymorphic schema. So, if this fix works for you, I’m fine with your PR.

julienrf:
Do you mind rebasing on `master` (the CI was broken yesterday)

julienrf:
Hmm, maybe using `Manifest` is not a good idea: https://github.com/scala/scala/pull/8162.

gmarti:
I try to use classTags... i wanted to but was not able

julienrf:
Do you mean `TypeTag`? I’m afraid this will not be supported by Scala.js either :-/

Does the workaround I’ve suggested work?

gmarti:
Which workaround ? if Manifest don't work and TypeTag don't work we have only erased ClassTag so we cannot use "typeArguments" method.
It will probably not work for scala.js world

julienrf:
> Which workaround?

Using a newtype for each instantiation of your polymorphic type:

~~~ scala
class GenericFoo[T](val value: T)
case class GenericFooInt(value: Int) extends GenericFoo[Int]
implicit def schemaPrimitive: JsonSchema[GenericFooInt] = genericJsonSchema
~~~

I agree that it’s less convenient to use, though…

gmarti:
Created a PR with a nicer workaround, I close this one

PR #326, opened 2019-07-11
Add streaming support

### LINKED ISSUE #6: Add support for streamed entities
Streamed entities is useful for long polling, server-sent events, websockets, and for serving large entities.

Consider using [monix](monix.io) to give a consistent programming experience on server and client sides.

julienrf:
Add algebras for describing endpoints whose responses are streamed (using HTTP1.1 chunked transfer-encoding), and WebSocket endpoints.

Add interpreters for these algebras for `akka-http-client`, `akka-http-server`, `xhr-client`, `play-server`, and `openapi`.

Since Wiremock server does not support streaming, we use Akka-HTTP in tests to implement either a server (when testing client interpreters) or a client (when testing a server interpreter).

Fixes #6

The PR is not ready to merge, but I’d like to know what you think about the design. In particular, streaming is achieved by introducing a new type `ChunkedEndpoint[A, B]`, as opposed to reusing the existing `Endpoint` type (ie, this would give something like `Endpoint[A, Chunks[B]]`). One limitation of the current design is that only *responses* can be streamed, not *requests*. This was motivated mostly because `XMLHttpRequest` (in web browsers) doesn’t support streaming requests. But the more I think about it, the more I find it too limiting. Also, that design would not be a good basis to add support for streamed requests afterward (we would have to have 3 types: `ChunkedRequestEndpoint`, `ChunkedResponseEndpoint` and `ChunkedRequestAndResponseEndpoint`). Alternatively, if we only introduce a `Chunks` type constructor like in `Endpoint[A, Chunks[B]]` it will be easy to add support for streamed requests (`Endpoint[Chunks[A], B]`) and streamed request and response (`Endpoint[Chunks[A], Chunks[B]]`). But that design would have drawbacks too. For instance, client interpreters fix the `Endpoint[A, B]` type to be `A => Future[B]`. Therefore, a streamed endpoint of type `Endpoint[A, Chunks[B]]` would expand to `A => Future[Source[B, _]]`, which is less convenient to use than just `A => Source[B, _]` (which is what we have in the PR).


Krever:
Hey, sorry for not responding earlier. 
In my experience responses are streamed far more often than requests so current design is acceptable, yet the more generic alternative seems better in a long term. Regarding `A => Future[Source[B, _]]` I don't think its problematic in any way but raise one other concern: would `Endpoint[Chunks[A], B]` expand to ` Future[Source[A, _]]` => B` ? If so I don't think such signature is implementable without `Await.result` which is rather bad.

To sum up, I vote for more generic design even at the cost of some additional complexity at use site. Unless of course there are other reasons not to use it. 

I didnt have time to look at the code but I will try to do next week.

julienrf:
Thanks for your feedback Wojtech :)

> would `Endpoint[Chunks[A], B]` expand to ` Future[Source[A, _]] => B`?

No, it would expand to `Source[A, _] => Future[B]`.

I will try to play a bit with an alternative design, to see where it goes…

julienrf:
I’m closing this one in favor of #465

PR #331, opened 2019-07-18
Add JSON schema for a  Byte

kirankbs:
closes: https://github.com/julienrf/endpoints/issues/330

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
@@ -111,6 +111,8 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   lazy val booleanJsonSchema: DocumentedJsonSchema = Primitive("boolean")
 
+  lazy val byteJsonSchema: DocumentedJsonSchema = Primitive("byte")
+
   def arrayJsonSchema[C[X] <: Seq[X], A](implicit
     jsonSchema: JsonSchema[A],
     factory: Factory[A, C[A]]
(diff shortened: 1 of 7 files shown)


openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -111,6 +111,8 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   lazy val booleanJsonSchema: DocumentedJsonSchema = Primitive("boolean")
 
+  lazy val byteJsonSchema: DocumentedJsonSchema = Primitive("byte")
```
julienrf: There is no `byte` type in OpenAPI. We should just use `integer`. Unfortunately, there is no `int8` *format* (as there is `int32` for `Int` values).

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -111,6 +111,8 @@ trait JsonSchemas extends endpoints.algebra.JsonSchemas {
 
   lazy val booleanJsonSchema: DocumentedJsonSchema = Primitive("boolean")
 
+  lazy val byteJsonSchema: DocumentedJsonSchema = Primitive("integer", format = Some("int32"))
```
julienrf: Unfortunately, `int32` is not appropriate for `Byte` and it seems that the OpenAPI specification does not have `int8`. Therefore, we should just use `None` for the `format` parameter.

```suggestion
  lazy val byteJsonSchema: DocumentedJsonSchema = Primitive("integer")
```

julienrf:
Thanks!

Issue #353, opened 2019-09-10
algebra.Urls is not mixing PartialInvariantFunctorSyntax


gmarti:
We need to mix a trait in 0.10.1 to be able to use xmap in algebra.Urls : `trait Urls extends algebra.Urls with PartialInvariantFunctorSyntax`

And i think algegra.Urls should mix PartialInvariantFunctorSyntax and use it instead of : 
```
implicit def uuidQueryString: QueryStringParam[UUID] =
    queryStringParamPartialInvFunctor.xmapPartial(stringQueryString, (x: String) => Try(UUID.fromString(x)).toOption, _.toString)
```

We have afterwards` trait Requests extends Urls with Methods with PartialInvariantFunctorSyntax with SemigroupalSyntax` but we should be able to extends only Urls and go with it right ? withtout mixing some syntax sugar

julienrf:
Fixed by #354

PR #354, opened 2019-09-11
Add choiceResponse operation

### LINKED ISSUE #42: Improve modelling of responses in endpoint algebra

julienrf:
Also add `InvariantFunctor[Response]` instance.

Re-implement `wheneverFound`, `authenticated`, and `wheneverValid`
operations in terms of `choiceResponse` and `xmap`.

Fixes #42 and #353.

### DIFF
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/BasicAuthentication.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/BasicAuthentication.scala
@@ -1,12 +1,8 @@
 package endpoints.akkahttp.client
 
-import akka.http.scaladsl.model.{StatusCodes => AkkaStatusCodes}
 import akka.http.scaladsl.model.headers._
 import endpoints.algebra
 import endpoints.algebra.BasicAuthentication.Credentials
-import endpoints.algebra.Documentation
-
-import scala.concurrent.Future
 
 /**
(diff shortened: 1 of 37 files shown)


julienrf:
The docs look like the following:

response-doc

Should I expand the last two examples into more complete endpoint examples, or are the short versions enough? (cc @fduraffourg)

PR #389, opened 2019-10-18
Support documenting webhooks

### LINKED ISSUE #385: Support documenting webhooks
See https://swagger.io/docs/specification/callbacks/

julienrf:
Also group together all documentation related parameters of the `endpoint` constructor in single case class `EndpointDocumentation`

Fixes #385

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
@@ -43,16 +43,43 @@ trait EndpointsWithCustomErrors extends Requests with Responses with Errors {
     *
     * @param request  Request
     * @param response Response
-    * @param summary optional summary documentation
-    * @param description optional description documentation
-    * @param tags list of OpenApi tags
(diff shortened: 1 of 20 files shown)


julienrf:
The test example produces the following documentation in Swagger UI:

webhook-doc

algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
```diff
@@ -43,16 +43,43 @@ trait EndpointsWithCustomErrors extends Requests with Responses with Errors {
...
+    docs: EndpointDocumentation = EndpointDocumentation()
+  ): Endpoint[A, B]
+
+  /**
+    * @param summary     Short description
+    * @param description Detailed description
+    * @param tags        OpenAPI tags
+    * @param callbacks   Callbacks indexed by event name
+    */
+  case class EndpointDocumentation(
```
julienrf: Maybe we should name this type `EndpointDocs`, for conciseness and consistency with the `docs` parameter name of the `endpoint` operation.

PR #392, opened 2019-10-24
Support documenting endpoints as “deprecated”

### LINKED ISSUE #345: Support documenting endpoints as “deprecated”
Since [OpenAPI 3.0,](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md) it's possible to deprecate endpoints/fields etc.

I suppose for then endpoints we could do something like: 
```scala
  def endpoint[A, B](
    request: Request[A],
    response: Response[B],
    summary: Documentation = None,
    description: Documentation = None,
    tags: List[String] = Nil,
    deprecated: Boolean = false,
  ): Endpoint[A, B]
```
For the fields I haven't thought about it yet

julienrf:
Fixes #345

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
@@ -56,12 +56,14 @@ trait EndpointsWithCustomErrors extends Requests with Responses with Errors {
     * @param description Detailed description
     * @param tags        OpenAPI tags
     * @param callbacks   Callbacks indexed by event name
+    * @param deprecated  Indicates whether this endpoint is deprecated or not
     */
   case class EndpointDocs(
     summary: Documentation = None,
     description: Documentation = None,
     tags: List[String] = Nil,
-    callbacks: Map[String, CallbacksDocs] = Map.empty
+    callbacks: Map[String, CallbacksDocs] = Map.empty,
+    deprecated: Boolean = false
   )
 
   /**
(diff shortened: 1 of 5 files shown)


algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
```diff
@@ -52,16 +52,18 @@ trait EndpointsWithCustomErrors extends Requests with Responses with Errors {
...
+    * @param callbacks    Callbacks indexed by event name
+    * @param isDeprecated Declares this endpoint to be deprecated.
     */
   case class EndpointDocs(
     summary: Documentation = None,
     description: Documentation = None,
     tags: List[String] = Nil,
-    callbacks: Map[String, CallbacksDocs] = Map.empty
+    callbacks: Map[String, CallbacksDocs] = Map.empty,
+    isDeprecated: Boolean = false
```
tdroxler: To stay consistent with what we have in `Operation` shouldn't we have simply `deprecated`? If don't find the `is` necessary.

tdroxler:
Kewl :+1:  easy but very useful PR. Thx

algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
julienrf: I thought it would be more “scala-ish”, but I agree that it’s probably simpler to use the OpenAPI vocabulary everywhere.

PR #393, opened 2019-10-25
Support documenting fields in generically derived JSON schemas

### LINKED ISSUE #249: Field descriptions via tagged types
Our documentation requirements include providing some descriptive text for each model field. We could use the JSON schema description DSL but that means introducing a separate schema: we would much rather use automatic generation with `genericJsonSchema`.

So how about taking advantage of tagged types and a dedicated typeclass allowing for providing descriptions for each tag?

For instance, with the following definitions using shapeless tagged types:

```scala
object tags {
  sealed trait CardNumber
  sealed trait CVV
}
import tags._
case class CreditCard(number: String @@ CardNumber, 
                      holder: String, 
                      cvv: String @@ CVV)
```

And a typeclass definition and instances: 

```scala
 trait TypeDescriptor {
    def apply(): String
  }

  implicit val cardNumberDescriptor: TypeDescriptor =  () => "Credit card number"
  implicit val cvvDescriptor: TypeDescriptor = () => "Card verification value"
```

With that in place, `genericJsonSchema` could grab instances of this typeclass for a type or tag and include the description in the generated schema. This also allows for reuse of descriptions btw.
Does it make sense?

julienrf:
Fixes #249 (and supersedes #128)

### DIFF
--- a/documentation/manual/src/doc/algebras/json-schemas.md
+++ b/documentation/manual/src/doc/algebras/json-schemas.md
@@ -103,7 +103,7 @@ by wrapping it in the `lazySchema` constructor:
 ## Generic derivation of JSON schemas (based on Shapeless) {#generic-derivation-of-json-schemas}
 
 The module presented in this section uses Shapeless to generically derive JSON schemas
-for algebraic data type definitions.
+for algebraic data type definitions (sealed traits and case classes).
 
 ~~~ scala expandVars=true
 "org.julienrf" %% "endpoints-json-schema-generic" % "{{version}}"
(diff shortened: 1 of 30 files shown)


julienrf:
Review by @guersam and @jchapuis?

documentation/manual/src/doc/algebras/json-schemas.md
```diff
@@ -118,9 +118,22 @@ reduced to the following:
...
 
 The `genericJsonSchema` operation builds a JSON schema for the given
-type, mapping each case class field to a JSON object field of the same name,
-and each leaf of a sealed trait to a tagged JSON object whose type discriminator
-uses the case class fully qualified name.
+type. The rules for deriving the schema are the following:
+
+- the schema of a case class is a JSON object,
+- the schema of a sealed trait is the alternative (`oneOf`) of its leaf case
+  class schemas, discriminated by the case class names,
```
jchapuis: Maybe you could mention that overriding 'classTagToSchemaName' is now supported, if relevant?

json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
```diff
@@ -51,59 +50,56 @@ trait JsonSchemas extends algebra.JsonSchemas {
...
+      jsonSchemaTail: DocumentedGenericRecord[T, DT]
+    ): DocumentedGenericRecord[FieldType[L, Option[H]] :: T, DH :: DT] =
+      new DocumentedGenericRecord[FieldType[L, Option[H]] :: T, DH :: DT] {
+        def record(docs: DH :: DT) =
+          (optField(labelHead.value.name, docs.head.map(_.text))(jsonSchemaHead) zip jsonSchemaTail.record(docs.tail))
             .xmap[FieldType[L, Option[H]] :: T] { case (h, t) => shapelessField[L](h) :: t }(ht => (ht.head, ht.tail))
       }
 
-    implicit def consCoproduct[L <: Symbol, H, T <: Coproduct](implicit
+    implicit def consCoproduct[L <: Symbol, H, T <: Coproduct, D](implicit
```
guersam: what's the new type param `D` for? I can't find its actual usage.

julienrf: Good catch, that’s for the next PR ;) I’ll remove it.

documentation/manual/src/doc/algebras/json-schemas.md
julienrf: Indeed, I should document that. I was thinking of doing it in another PR, where I will also introduce a way to customize the name of a generically derived schema.

PR #394, opened 2019-10-27
Remove `namedGenericJsonSchema` and introduce `genericRecord` and `genericTagged`

julienrf:
`genericRecord[Xxx].named("xxx")` is the new syntax for the previous `namedGenericJsonSchema[Xxx]("xxx")`

Also, make the `named` operation available only on `Record`, `Tagged`, and `Enum`.

### DIFF
--- a/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
@@ -89,9 +89,11 @@ trait JsonSchemas
     )
   }
 
-  def named[A, S[T] <: JsonSchema[T]](schema: S[A], name: String): S[A] = schema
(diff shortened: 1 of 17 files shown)


json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
```diff
@@ -100,6 +102,9 @@ trait JsonSchemas
     }
   }
 
+  def lazyRecord[A](schema: => Record[A], name: String): JsonSchema[A] = lazySchema(schema, name)
```
gmarti: Should they return a record / tagged ?

julienrf: I believe returning a `JsonSchema` is enough. We can change that later, if needed.

PR #395, opened 2019-10-31
Increase configurability of generic schema derivation process

### LINKED ISSUE #137: Shorter derived schema tag name in OpenAPI
`endpoints.openapi.JsonSchemas` uses the fully qualified class name as the tag of a derived json schema. Unfortunately many openapi client generators doesn't support for package stripping, so that the generated models tend to have unnecessarily long class names such as `ComExampleMyApiProtocolRealClassName` from `com.example.my.api.protocol.RealClasssName`.

Suggestions:
  1. Use simple name instead of FQCN
  2. Let users choose between simple name and FQCN via
      a. an additional argument of `genericJsonSchema`
      b. implicit `Configuration` class like [circe-generic-extras](https://circe.github.io/circe/codecs/custom-codecs.html#custom-key-mappings-via-annotations)
  3. Use FQCN by default and let users specify the prefix to strip via something like `openapi.JsonSchema#packageNameToStrip`

julienrf:
Fixes #137

- Adds a `@discriminator` annotation allowing users to define the
  discriminator field name of a sealed trait schema (if absent,
  fallback to `defaultDiscriminatorName`)
- Adds a `@name` annotation allowing users to define the schema name
  of a sealed trait schema or a case class schema (if absent, fallback
  to `classTagToSchemaName`)
- Make `withDiscriminator` an operation available on `Tagged` schemas

### DIFF
--- a/documentation/manual/src/doc/algebras/json-schemas.md
+++ b/documentation/manual/src/doc/algebras/json-schemas.md
@@ -150,10 +150,16 @@ type. The rules for deriving the schema are the following:
   the same name and type (for instance, the generic schema for the `Rectangle`
   type has a `width` required property of type `integer`),
 - each case class field of type `Option[A]` for some type `A` has a corresponding
-  optional JSON object property of the same name and type.
+  optional JSON object property of the same name and type,
+- documentation specific to case class fields can be defined by annotating the fields
+  with the `@docs` annotation,
+- for sealed traits, the discriminator field name can be defined by the `@discriminator`
+  annotation, otherwise the `defaultDiscriminatorName` value is used,
+- the schema is named by the `@name` annotation, if present, or by invoking the
+  `classTagToSchemaName` operation with the `ClassTag` of the type for which the schema
+  is derived.
 
-Documentation specific to case class fields can be defined by annotating the fields
-with the `@docs` annotation:
+Here is an example that illustrates how to configure the generic schema derivation process:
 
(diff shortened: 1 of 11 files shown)


julienrf:
>  I don't remember if this was discussed already but another approach would have been type classes. But for fields this would have been frustrating without an easy way to refine types in the language.

Do you mean something like this?

~~~ scala
case class User(name: String, age: Int)

object User {
  implicit val userDocs: Docs[User] = Docs("A user of the system")
  implicit val nameDocs: Docs[User @@ "name"] = Docs("The name of the user")
}
~~~

(Assuming you compile with `-Yliteral-types`)

I’m not sure that approach would be better.

jchapuis:
> (Assuming you compile with `-Yliteral-types`)

Ah yes we have literal types now in 2.13! but the tagging operator is from shapeless, right? so that's out of the scope of vanilla scala anyway
 
> I’m not sure that approach would be better.

Me neither. I just thought if we wanted to avoid annotations at all costs...but the resulting expressions are anyways more verbose. Maybe with dotty that could become an option, given that we no longer need to name the implicits. But we would need some built-in type-tagging ability, I don't know if that's planned.

julienrf:
“type-tagging” can be implemented at the library-level (and annotations processing in this PR relies on macros, but that’s okay).

Nevertheless, I think it is a little bit simpler to decorate the fields with annotations rather than writing these type class instances.

PR #402, opened 2019-11-07
Add `as` operation to `JsonSchema`

### LINKED ISSUE #379: In `generic.JsonSchemas`, operation `as` only available on `Record[A]`
It should be available on `JsonSchema[A]` as well, because this type also supports `xmap`.

julienrf:
Fixes #379

- Make `Record#zip` return a flat tuple (by using `Tupler`) instead
  of nested pairs,
- Rename `Validated#tuple` to `Validated#zip`, to be consistent with
  `Record#zip`,
- Remove `:*:` operation from `Record`: we can just use `zip`,
- Remove `tupled` operation from `Record`: it is not needed anymore
  since `zip` returns a flat tuple,
- Add `as[B]` operation to `Record[A]` and `JsonSchema[A]`. This
  operation takes an implicit `Generic.Aux[A, B]`, which is summoned
  by shapeless (although, to make it work we had to introduce a rule
  that checks that the generic representation of `A` is the same as
  the generic representation of `B`),
- Add implicit `Tupler` instances appending an element to a tuple
  from 3 to 21 elements.

### DIFF
--- a/akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala
+++ b/akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala
@@ -85,7 +85,7 @@ trait Urls extends algebra.Urls with StatusCodes { this: EndpointsWithCustomErro
 
   def urlWithQueryString[A, B](path: Path[A], qs: QueryString[B])(implicit tupler: Tupler[A, B]): Url[tupler.Out] = { (segments: List[String], query: Map[String, List[String]]) =>
     path.validate(segments).flatMap {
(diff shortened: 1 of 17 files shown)


julienrf:
Note that `as` is still not available on `Tagged`, which means that you will still have to use `xmap` in [this case](https://github.com/julienrf/endpoints/blob/b5dce23d03efd0474c2bfb7fd40d09b7a94afd2e/json-schema/json-schema/src/test/scala/endpoints/algebra/JsonSchemasDocs.scala#L35-L45)…

I’ve tried to have `as` on `Tagged` but this requires more work than expected, so I’m delaying it for another PR.

Issue #399, opened 2019-11-01
Should we provide our own JSON codecs for the JsonSchemas algebra?


julienrf:
Following up the discussion below, I plan to do the following tasks:

- [x] Implement JSON encoders for `JsonSchemas` algebra
- [x] Use these JSON encoders in `openapi.JsonSchemas` interpreter, and leverage the fact that we already have a dependency to a JSON library to implement an `Encoder[OpenApi]` (as opposed to the current situation where we use `OpenApiSchemas` with a `JsonSchemas` codec interpreter)
- [x] Fix #185
- [x] Fix #121
- [x] Implement JSON decoders for `JsonSchemas` algebra
- [x] Remove the need to mix a JSON codec interpreter with `JsonSchemaEntities` since we can use the built-in one

----

It started as an anecdotical question and eventually pulled significant design reconsiderations. I’d like to request your feedback.

I was looking at #185 (generalizing enumerations to types other than `String`) and realized that solving this issue would require a significant change. Eventually, we’d need the `openapi` interpreter to not only be able to produce documentation for `JsonSchema[A]` values, but also to be able to **encode `A` values in JSON**. So, it would have to provide an `Encoder[A, Json]`. So far, such encoders were provided by the `playjson` and `circe` interpreters, which delegate to Play-json and Circe the actual implementation of the codecs. The idea was that people could pick their preferred JSON backend between those two.

So, when I realized that solving #185 would require including a JSON encoder, I was tempted to introduce a way for users to provide a `circe` or `playjson` interpreter to the `openapi` interpreter, so that the `openapi` could benefit from their JSON encoders. This would require little work in _endpoints_ but would be a bit complicated/confusing for end-users, I’m afraid.

To keep things simple, I’d rather **implement a JSON encoder in the `openapi` interpreter**. End-users wouldn’t have to do anything, and we could solve #185, as well as #121. And we could also simplify the way we generate the OpenAPI document itself (which is currently defined as a JSON schema, in the `OpenApiSchemas` trait, in which the user has to mix a codec interpreter to produce a concrete `OpenApiEncoder`, as explained [here](http://julienrf.github.io/endpoints/interpreters/openapi.html#endpoints)). So, this would be super cool.

[...]

PR #410, opened 2019-11-22
Embed JSON encoders in openapi interpreter

julienrf:
This addresses the second point of #399


documentation/manual/src/doc/interpreters/openapi.md
```diff
@@ -6,6 +6,9 @@
```
julienrf: The code examples of this page have been extended to include more context. This is how it looks like now:

Screenshot from 2019-11-25 09-25-37

(you can compare with the [current version](http://julienrf.github.io/endpoints/interpreters/openapi.html))

PR #414, opened 2019-11-24
sbt-scalajs 1.0.0-RC1

sullis:


julienrf:
Thanks @sullis! I believe we’ll have to support both 0.6 and 1.0.0 for quite some time. Would you be interested in doing that? Do you need any guidance?

harpocrates:
We'd like to use the JS clients of _endpoints_ with scala.js 1.0.0. How does scala.js handle building with multiple versions?

julienrf:
@harpocrates We need to pass the Scala.js version as an environment variable to select the correct Scala.js plugin version.

But the real impact is on the deployment process. Currently, we just do a `publishSigned` for each supported Scala version. If we support multiple Scala.js version we will have to also invoke `publishSigned` on the Scala.js modules only for each Scala.js version.

Also, note that we might have to tweak the deployment process anyway soon because we will probably stop using the same release lifecycle for all the components of _endpoints_ (the algebra and interpreters may have distinct release lifecycles).

I would like to avoid using fragile bash scripts, if possible, but I didn’t think of a good solution yet.

harpocrates:
Thanks for clarifying. I did see "Cross-building for Scala.js 0.6.x and 1.x" in [the 1.0.0 announcement](https://www.scala-js.org/news/2020/02/25/announcing-scalajs-1.0.0/) but was hoping there was some better solution. I'm glad our internal project only needs to support 1.0.0 for now 😄 - it seems a bit painful to deal with this.

harpocrates:
FWIW: for the time being we just made a fork of `endpoints` to use 1.0.0. In the process I found:

  * `faithful` and `play-json` also don't yet support 1.0.0, so we disabled everything that depends on those
  * `libraryDependencies += "org.scala-js" %% "scalajs-env-jsdom-nodejs" % "1.0.0"` is needed

Those two caveats aside, everything has been working great (through the `endpoints-xhr-client` client)!

julienrf:
Superseded by #582

PR #415, opened 2019-11-24
Travis CI: add openjdk11

sullis:


julienrf:
Hello @sullis, can you explain why we need this change? I would like to keep the matrix as small as possible because the CI takes a lot of time to complete :(

Maybe we can just replace jdk8 by jdk11?

sullis:
Perhaps there is another solution. 

I recommend replacing Travis CI with GitHub Actions CI/CD.

With GitHub Actions workflows, builds will execute in parallel. Here is a Scala CI workflow that I am using on another project:

https://github.com/sullis/jms-testkit/blob/master/.github/workflows/ci.yml

sullis:
Closing this PR.  I'll open a new PR that adds a GitHub CI workflow.

julienrf:
> With GitHub Actions workflows, builds will execute in parallel.

Note that this is also the case currently, the builds for the three supported Scala versions are run in parallel.

julienrf:
I'm happy to move to GH actions though, if it has advantages!

PR #417, opened 2019-11-26
add GitHub Actions CI

sullis:
This is a starting point.  This can be enhanced later in a second pull request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: Scala CI
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '1.8.0', '11.0.x', '13.0.x' ]
+        scala: [ '2.13.1', '2.12.10', '2.11.12' ]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK
+      uses: actions/setup-java@v1.2.0
+      with:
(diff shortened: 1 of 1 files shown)


julienrf:
Thanks @sullis!

Before I can consider switching to GH actions, I need the following features to be supported:
- test coverage must be run only if the tests did pass, we don’t need to run test coverage on all Scala versions (2.12 is enough for now),
- when a tag is pushed to master, a release should be published,
- we also want to test that the documentation builds without errors, for some Scala version (currently, 2.12),
- the resolved dependencies should be cached between CI runs.

Also, is there a way to test the workflow of the PR?

sullis:
This GitHub Actions workflow cannot execute until after the Pull Request is merged.

("Only repository collaborators may create workflows")

I recommend merging this PR.  It will activate the workflow and you can observe the results.

Both TravisCI and GitHub Actions will execute a build.

After we confirm this workflow (ci.yml) passes, I can open a follow-up PR that adds additional behavior

julienrf:
OK, let’s try that!

sullis:
the Workflow has started:
https://github.com/julienrf/endpoints/actions

sullis:
the build failed:  
https://github.com/julienrf/endpoints/commit/1fc97ef128e932064760fa890b042b63290e3663/checks?check_suite_id=329725041

I can work on a fix tomorrow.

julienrf:
Hello @sullis. The failing workflow marks all new PRs as failing. I have nothing against GitHub actions but I can’t keep the current state. Could you please make your experiments with GitHub Actions on your fork first? I’m gonna revert this PR in the meantime.

PR #447, opened 2019-12-20
Add more structure and links between API documentation pages

julienrf:
The architecture of _endpoints_ does not render well in the API documentation. Consider for instance the type `Response[A]` introduced by the `Responses` algebra. This type has a method `orElse`, but this method is not shown in the documentation because the type `Response[A]` is an abstract type member, and the method is provided via an implicit conversion. Similarly, types that have a `PartialInvariantFunctor` instance have the methods `xmap` and `xmapPartial` but those don’t show up in the documentation. Last, some types and operations do show up in the documentation but they exist only for technical reasons (e.g., `ResponseSyntax` is the implicit class that provides additional operations to the `Response` type), and introduce noise in the API documentation.

This commit meticulously classifies types and operations introduced by algebras so that “technical” irrelevant members (e.g., implicit classes) are separated from relevant members. It also systematically adds a note to each type that has implicit methods, with links to those methods.

### DIFF
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/BuiltInErrors.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/BuiltInErrors.scala
@@ -2,6 +2,9 @@ package endpoints.akkahttp.client
 
 import endpoints.{Invalid, algebra}
 
+/**
+  * @group interpreters
+  */
 trait BuiltInErrors extends algebra.BuiltInErrors { this: EndpointsWithCustomErrors =>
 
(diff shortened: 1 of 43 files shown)


julienrf:
All the relevant types introduced by an algebra are grouped together. You can see in the detail of the `Path[A]` docs the links to its implicit methods:

Screenshot from 2019-12-20 16-35-16

Relevant operations are also grouped together. Implicit classes, which bring operations to type members, are also part of this group:

Screenshot from 2019-12-20 16-37-08

PR #465, opened 2020-02-02
Add streaming support

### LINKED ISSUE #6: Add support for streamed entities

julienrf:
Fixes #6

This is an alternative to #326 that is more flexible and introduces fewer new types.

Introduces a `ChunkedEntities` algebra defining request and response entities that use the chunked transfer-encoding.

Streamed entities are represented by a type `Chunks[A]` in the algebra. For instance, a streamed response entity containing chunks of type `Event` has type `ResponseEntity[Chunks[Event]]`, and an endpoint returning such an entity has type `Endpoint[Unit, Chunks[Event]]`.

The type `Chunks[A]` is an abstract type, which is fixed by interpreters. Both client and server interpreters currently fix it to Akka Stream’s `Source[A, _]`.

Since Wiremock server does not support streaming, we use Akka-HTTP in tests to implement either a server (when testing client interpreters) or a client (when testing a server interpreter).

The type `ResponseEntity` of the Play server interpreter had to be adapted to support chunked transfer-encoding.

The chunked transfer-encoding works nice for long-running endpoints that intermittently return response items. It is a bit less adapted to file upload/download because the response `Content-Length` is unknown, which means that clients can not show a progress bar, for instance. The fact that we rely on a one to one mapping between low-level chunks and high-level Akka Stream’s `Source` items means that we need the underlying HTTP library to do no reframing process. This doesn’t seem to be the case of the Play-WS library, therefore the `endpoints.play.client` interpreters don’t support the `ChunkedEntities` algebra. The PR provides interpreters for `endpoints.play.server`, `endpoints.akkahttp.client`, `endpoints.akkahttp.server`, and `endpoints.openapi`.


nightscape:
@julienrf sorry for necro-bumping, but does this PR already allow to specify/use WebSockets, or would additional functionality be required?

julienrf:
This PR only allows us to describe (and interpret) HTTP endpoints that consume or return entities encoded using the Chunked Transfer-Encoding. So, no, this does not cover WebSockets.

Nevertheless, I believe that we could easily build WebSockets support on top of that. This would be a welcome contribution :)

PR #476, opened 2020-02-13
Give lazy constructors more precise return types

bmeesters:
This is a first version to check whether this is the right design to make lazy constructors return types more precise. I will check locally if this solves our problems, and if so will finish up this PR by adding unit tests & documentation.

**Edit:**
I added some unit tests and verified this code works in our services locally.

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
@@ -43,11 +43,7 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
       * principles apply to [[DocumentedCoProd]].
       */
     sealed trait DocumentedRecord extends DocumentedJsonSchema {
-      def value: StrictDocumentedRecord
-      def fields: List[Field]
-      def additionalProperties: Option[DocumentedJsonSchema]
-      def name: Option[String]
-      def example: Option[ujson.Value]
+      def toStrict: StrictDocumentedRecord
       def withName(name: String): DocumentedRecord
     }
 
@@ -57,29 +53,21 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
       name: Option[String] = None,
       example: Option[ujson.Value] = None
     ) extends DocumentedRecord {
-      val value: StrictDocumentedRecord = this
+      val toStrict: StrictDocumentedRecord = this
       def withName(name: String): DocumentedRecord = copy(name = Some(name))
     }
 
     class LazyDocumentedRecord(_record: => StrictDocumentedRecord) extends DocumentedRecord {
-      lazy val value: StrictDocumentedRecord = _record
-
-      def fields: List[DocumentedJsonSchema.Field] = value.fields
-      def additionalProperties: Option[DocumentedJsonSchema] = value.additionalProperties
-      def name: Option[String] = value.name
-      def example: Option[Value] = value.example
-      def withName(name: String): DocumentedRecord = new LazyDocumentedRecord(value.copy(name = Some(name)))
+      lazy val toStrict: StrictDocumentedRecord = _record
+      def example: Option[Value] = toStrict.example
+      def withName(name: String): DocumentedRecord = new LazyDocumentedRecord(toStrict.copy(name = Some(name)))
     }
 
     case class Field(name: String, tpe: DocumentedJsonSchema, isOptional: Boolean, documentation: Option[String])
 
     /** @see [[DocumentedRecord]]. */
     sealed trait DocumentedCoProd extends DocumentedJsonSchema {
-      def alternatives: List[(String, DocumentedRecord)]
-      def name: Option[String]
-      def discriminatorName: String
-      def example: Option[ujson.Value]
-      def value: StrictDocumentedCoProd
+      def toStrict: StrictDocumentedCoProd
       def withName(name: String): DocumentedCoProd
       def withDiscriminatorName(name: String): DocumentedCoProd
     }
@@ -90,22 +78,18 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
       discriminatorName: String = defaultDiscriminatorName,
       example: Option[ujson.Value] = None
     ) extends DocumentedCoProd {
-      val value: StrictDocumentedCoProd = this
+      val toStrict: StrictDocumentedCoProd = this
       def withName(name: String): DocumentedCoProd = copy(name = Some(name))
       def withAlternatives(alternatives: List[(String, DocumentedRecord)]): DocumentedCoProd = copy(alternatives = alternatives)
       def withDiscriminatorName(name: String): DocumentedCoProd = copy(discriminatorName = name)
     }
 
     class LazyDocumentedCoProd(_coProd: => StrictDocumentedCoProd) extends DocumentedCoProd {
-      lazy val value: StrictDocumentedCoProd = _coProd
+      lazy val toStrict: StrictDocumentedCoProd = _coProd
+      def example: Option[Value] = toStrict.example
 
-      def alternatives: List[(String, DocumentedRecord)] = value.alternatives
-      def name: Option[String] = value.name
-      def discriminatorName: String = value.discriminatorName
-      def example: Option[Value] = value.example
-
-      def withName(name: String): DocumentedCoProd = new LazyDocumentedCoProd(value.copy(name = Some(name)))
-      def withDiscriminatorName(name: String): DocumentedCoProd = new LazyDocumentedCoProd(value.copy(discriminatorName = name))
+      def withName(name: String): DocumentedCoProd = new LazyDocumentedCoProd(toStrict.copy(name = Some(name)))
+      def withDiscriminatorName(name: String): DocumentedCoProd = new LazyDocumentedCoProd(toStrict.copy(discriminatorName = name))
     }
 
     class LazySchema(_schema: => DocumentedJsonSchema) extends DocumentedJsonSchema {
@@ -179,14 +163,14 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
   def lazyRecord[A](schema: => Record[A], name: String): Record[A] = {
     new Record[A](
       ujsonSchemas.lazyRecord(schema.ujsonSchema, name),
-      new LazyDocumentedRecord(namedRecord(schema, name).docs.value)
+      new LazyDocumentedRecord(namedRecord(schema, name).docs.toStrict)
     )
   }
 
   def lazyTagged[A](schema: => Tagged[A], name: String): Tagged[A] =
     new Tagged(
       ujsonSchemas.lazyTagged(schema.ujsonSchema, name),
-      new LazyDocumentedCoProd(namedTagged(schema, name).docs.value)
+      new LazyDocumentedCoProd(namedTagged(schema, name).docs.toStrict)
     )
 
   def emptyRecord: Record[Unit] =
@@ -219,13 +203,13 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
   def choiceTagged[A, B](taggedA: Tagged[A], taggedB: Tagged[B]): Tagged[Either[A, B]] =
     new Tagged(
       ujsonSchemas.choiceTagged(taggedA.ujsonSchema, taggedB.ujsonSchema),
-      StrictDocumentedCoProd(taggedA.docs.alternatives ++ taggedB.docs.alternatives)
+      StrictDocumentedCoProd(taggedA.docs.toStrict.alternatives ++ taggedB.docs.toStrict.alternatives)
     )
 
   def zipRecords[A, B](recordA: Record[A], recordB: Record[B])(implicit t: Tupler[A, B]): Record[t.Out] =
     new Record(
       ujsonSchemas.zipRecords(recordA.ujsonSchema, recordB.ujsonSchema),
-      StrictDocumentedRecord(recordA.docs.fields ++ recordB.docs.fields)
+      StrictDocumentedRecord(recordA.docs.toStrict.fields ++ recordB.docs.toStrict.fields)
     )
 
   def withExampleJsonSchema[A](schema: JsonSchema[A], example: A): JsonSchema[A] = {
@@ -234,8 +218,8 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
       djs match {
         case s: StrictDocumentedRecord => s.copy(example = Some(exampleJson))
         case s: StrictDocumentedCoProd => s.copy(example = Some(exampleJson))
-        case s: LazyDocumentedRecord => new LazyDocumentedRecord(s.value.copy(example = Some(exampleJson)))
-        case s: LazyDocumentedCoProd => new LazyDocumentedCoProd(s.value.copy(example = Some(exampleJson)))
+        case s: LazyDocumentedRecord => new LazyDocumentedRecord(s.toStrict.copy(example = Some(exampleJson)))
+        case s: LazyDocumentedCoProd => new LazyDocumentedCoProd(s.toStrict.copy(example = Some(exampleJson)))
         case s: LazySchema       => new LazySchema(updatedDocs(s.value))
         case s: Primitive        => s.copy(example = Some(exampleJson))
         case s: Array            => s.copy(example = Some(exampleJson))
@@ -317,13 +301,13 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
           else Schema.Reference(name, Some(expandRecordSchema(record, coprodBase, referencedSchemas + name)), None)
         case record @ StrictDocumentedRecord(_, _, None, _) =>
           expandRecordSchema(record, coprodBase, referencedSchemas)
-        case record: LazyDocumentedRecord => inner(record.value)
+        case record: LazyDocumentedRecord => inner(record.toStrict)
         case coprod @ StrictDocumentedCoProd(_, Some(name), _, _) =>
           if (referencedSchemas(name)) Schema.Reference(name, None, None)
           else Schema.Reference(name, Some(expandCoproductSchema(coprod, referencedSchemas + name)), None)
         case coprod @ StrictDocumentedCoProd(_, None, _, _) =>
           expandCoproductSchema(coprod, referencedSchemas)
-        case coprod: LazyDocumentedCoProd => inner(coprod.value)
+        case coprod: LazyDocumentedCoProd => inner(coprod.toStrict)
         case lzy: LazySchema => inner(lzy.value)
         case Primitive(name, format, example) =>
           Schema.Primitive(name, format, None, example)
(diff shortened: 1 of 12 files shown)


openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -119,74 +170,79 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
...
     val exampleJson = schema.ujsonSchema.codec.encode(example)
     def updatedDocs(djs: DocumentedJsonSchema): DocumentedJsonSchema =
       djs match {
-        case s: DocumentedRecord => s.copy(example = Some(exampleJson))
-        case s: DocumentedCoProd => s.copy(example = Some(exampleJson))
+        case s: StrictDocumentedRecord => s.copy(example = Some(exampleJson))
+        case s: StrictDocumentedCoProd => s.copy(example = Some(exampleJson))
+        case s: LazyDocumentedRecord => new LazyDocumentedRecord(s.value.copy(example = Some(exampleJson)))
+        case s: LazyDocumentedCoProd => new LazyDocumentedCoProd(s.value.copy(example = Some(exampleJson)))
+        case s: LazySchema       => new LazySchema(updatedDocs(s.value))
```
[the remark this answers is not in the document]
bmeesters: I initially removed LazySchema, however I moved it back (line 111) since it actually might still be useful when you want to abstract over what LazySchema you want. (We actually have an extension on this that still relies on LazySchema).

The implementation changed though, I made it a class instead of a sealed trait, so it is just three lines of code.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -245,42 +301,48 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
...
-        else Schema.Reference(name, Some(Schema.Enum(toSchema(elementType, coprodBase, referencedSchemas + name), values, None, example)), None)
-      case DocumentedEnum(elementType, values, None, example) =>
-        Schema.Enum(toSchema(elementType, coprodBase, referencedSchemas), values, None, example)
-      case lzy: LazySchema =>
-        toSchema(lzy.value, coprodBase, referencedSchemas)
-      case OneOf(alternatives, example) =>
-        val alternativeSchemas = alternatives.map(alternative => toSchema(alternative, coprodBase, referencedSchemas))
-        Schema.OneOf(EnumeratedAlternatives(alternativeSchemas), None, example)
-    }
+    @tailrec
```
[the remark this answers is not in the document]
bmeesters: No, I just added it since Intellij IDEA said I should.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -35,20 +37,81 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
...
+      * are present as abstract `def`s to make it easier in use, regardless of which version you use. The same
+      * principles apply to [[DocumentedCoProd]].
+      */
+    sealed trait DocumentedRecord extends DocumentedJsonSchema {
+      def value: StrictDocumentedRecord
+      def fields: List[Field]
+      def additionalProperties: Option[DocumentedJsonSchema]
+      def name: Option[String]
+      def example: Option[ujson.Value]
+      def withName(name: String): DocumentedRecord
```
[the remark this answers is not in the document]
bmeesters: Actually these are all the properties of the strict version, which you also need when working with the lazy version. (If you check out this branch and delete these you will see the rest of code in this file relies on these fields quite a lot.)

You could also pattern match in all these places, but that would result in a lot more code. Alternatively I could create something like a fold with two functions, one for the lazy one, and one for the strict one. But ultimately I think this is the simplest to understand, even if a bit boilerplate-y.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -35,20 +37,81 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
...
-    case class DocumentedCoProd(
+    /** @see [[DocumentedRecord]]. */
+    sealed trait DocumentedCoProd extends DocumentedJsonSchema {
+      def alternatives: List[(String, DocumentedRecord)]
+      def name: Option[String]
+      def discriminatorName: String
+      def example: Option[ujson.Value]
+      def value: StrictDocumentedCoProd
+      def withName(name: String): DocumentedCoProd
+      def withDiscriminatorName(name: String): DocumentedCoProd
```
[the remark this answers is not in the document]
bmeesters: Idem, see for example `expandRecordSchema` that relies on accessing properties (regardless if it is lazy or strict)

openapi/openapi/src/test/scala/endpoints/openapi/EndpointsTest.scala
```diff
@@ -63,26 +63,75 @@ class EndpointsTest extends AnyWordSpec with Matchers with OptionValues {
...
-        "Rec",
-        Some(Schema.Object(
-          Schema.Property("next", Schema.Reference("Rec", None, None), isRequired = false, description = None) :: Nil,
-          additionalProperties = None,
-          description = None,
-          example = None
-        )),
-        None
-      )
+  val recSchema =
```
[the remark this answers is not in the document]
bmeesters: Ah sorry for that, I have a tendency to hit auto-format a lot. I'll revert it.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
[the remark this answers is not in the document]
bmeesters: However, now that I think about it, if you use the `toStrict` you don't need these anymore. Should I do that?

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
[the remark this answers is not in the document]
bmeesters: I could remove it if you want?

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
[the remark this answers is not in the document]
bmeesters: Again, if I use `toStrict` they may be removed if you want?

openapi/openapi/src/test/scala/endpoints/openapi/EndpointsTest.scala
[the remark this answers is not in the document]
bmeesters: Now that I look more into it, the indentation has changes because I moved it out of the test, because I want to reuse it for both the lazy tagged and lazy record version

openapi/openapi/src/main/scala/endpoints/ujson/JsonSchemas.scala
```diff
@@ -74,7 +76,8 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
   implicit def taggedPartialInvFunctor: PartialInvariantFunctor[Tagged] =
     new PartialInvariantFunctor[Tagged] {
       def xmapPartial[A, B](fa: Tagged[A], f: A => Validated[B], g: B => A): Tagged[B] =
-        new Tagged[B](fa.discriminator) {
+        new Tagged[B] {
+          val discriminator: String = fa.discriminator
```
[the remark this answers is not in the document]
bmeesters: Do you want me to only change the base to a `def`, or also the children? It seems to make sense to use a val since it only needs to evaluated once?

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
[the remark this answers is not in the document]
julienrf: Yes, at least going through `toStrict` would make it explicit that we are forcing the evaluation of the underlying schema.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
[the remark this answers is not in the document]
julienrf: No, this is fine.

openapi/openapi/src/main/scala/endpoints/ujson/JsonSchemas.scala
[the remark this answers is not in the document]
julienrf: I would use `def` everywhere, for simplicity, since the right-hand-side is super simple (the JIT might be able to inline it if needed).

openapi/openapi/src/test/scala/endpoints/openapi/EndpointsTest.scala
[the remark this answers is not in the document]
julienrf: OK, I missed that :)

bmeesters:
Thanks @julienrf for the quick review. I have processed the comments. I did remove the accessor methods from the `Documented` traits. Though I left the `example` as that is required by the `DocumentedJsonSchema` interface.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -35,20 +37,65 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas { openapiJsonSc
...
+      def example: Option[Value] = toStrict.example
+
+      def withName(name: String): DocumentedCoProd = new LazyDocumentedCoProd(toStrict.copy(name = Some(name)))
+      def withDiscriminatorName(name: String): DocumentedCoProd = new LazyDocumentedCoProd(toStrict.copy(discriminatorName = name))
+    }
+
+    class LazySchema(_schema: => DocumentedJsonSchema) extends DocumentedJsonSchema {
+      lazy val value: DocumentedJsonSchema = _schema
+      def example: Option[ujson.Value] = value.example
+    }
```
julienrf: Sorry to be nitpicking, but I would rather not have `example` as a method on `LazySchema`, `LazyDocumentedCoProd` and `LazyDocumentedRecord`. I think it’s clearer to write `.toStrict.example` instead.

Also, do you think we could make `LazySchema` a (sealed) parent trait of both `LazyDocumentedCoProd` and `LazyDocumentedRecord`? This would remove the need for pattern matching on `LazySchema` if we already pattern match on one of those.

bmeesters: Sure, NP. Better to take a bit more time than to rush it in. I'll remove `example` from `LazySchema` and check whether I can make it a sealed (parent) trait.

bmeesters: So, we actually have an extension on _endpoints_ where we define 

```
def multiLazyJsonSchema[A, C[X] <: Seq[X]](schema: => JsonSchema[A]): JsonSchema[C[A]]
```

I am not sure how we could define it without `LazySchema` since we do not know if it is a tagged or a record. Which is why I brought it back in the first place. Any ideas?

julienrf: May I ask what is your use case for this function?

Also, wouldn’t it also work if `LazySchema` was a parent type of both `LazyDocumentedCoProd` and `LazyDocumentedRecord`?

julienrf: > Sorry to be nitpicking, but I would rather not have `example` as a method on `LazySchema`, `LazyDocumentedCoProd` and `LazyDocumentedRecord`. I think it’s clearer to write `.toStrict.example` instead.

Actually, I’m thinking a bit more about this one, and the `example` member would not be easy to remove because, as you said in another comment, it is required by the `DocumentedJsonSchema` trait.

Also, it was already in `LazySchema` before your PR, so it’s fine to keep it as-is.

bmeesters: > May I ask what is your use case for this function?

We have a model that could look roughly like this: `Trip -> Actions -> Shipments -> Actions -> Trip`. Basically the model is a cyclic graph, however various use cases are trees with a different root walking this graph (without cycles). For example, we have trips that have load and unload actions to deliver various shipments on various locations (Trip -> Actions -> Shipments). However from a shipment point of view you could also be part of multiple trips (e.g. from China to some harbor to big town to a small town) (Shipment -> Actions -> Trips).

Technically we could make this a `Tagged` instead of a `JsonSchema` since Actions is a Tagged, so it could probably work out in the end. However it feels less expressive if you have to copy every method for `Tagged` and `Record`, especially if you do not care which it is. Similar to how the non lazy array variant currently works.

> Also, wouldn’t it also work if LazySchema was a parent type of both LazyDocumentedCoProd and LazyDocumentedRecord

Not sure. You'd still need to know whether it is a record or a tagged if you cannot create a `LazySchema` directly. I also don't think there is much harm in keeping it like this. Though I am happy to explore other alternatives.

julienrf: I’m realizing that I should have started with that but, can you show me some code that needs `lazyRecord` to return a `Record` instead of a `JsonSchema`? The idea of `lazyRecord` or `lazyTagged` is that it allows you to refer to a schema that is being defined. However, the specific type of this schema should not matter.

bmeesters:
```
case class Trip(
  entityInfo: EntityInfo,
  vehicle: Option[Association[Vehicle]] = None,
  actors: Option[Seq[Association[Actor]]] = None,
  actions: Option[Seq[Association[Action]]] = None,
  constraint: Option[Association[Constraint]] = None,
) extends Entity

lazy val tripJsonSchema: Record[Trip] = (
      jsonSchema[EntityInfo] and // jsonSchema is a custom summoner (implicitly)
      lazyOptField[Association[A]]("vehicle", associationJsonSchema(vehicleJsonSchema))) zip
      optField[Seq[Association[Actor]]]("actors")(multiLazyJsonSchema(associationJsonSchema(actorFormat))) zip
      optField[Seq[Association[Action]]]("actions")(multiLazyJsonSchema(associationJsonSchema(actionJsonSchemas)) zip
      lazyOptField[Association[Constraint]]("constraint", associationJsonSchema(constraintJsonSchema))
  ).xmap((Trip.apply _).tupled)(unlift(Trip.unapply))
```

julienrf: I believe you could use `optField` instead of `lazyOptField`, and `arrayJsonSchema` instead of `multiLazyJsonSchema` like so:

~~~ scala
def associationVehicleSchema(tripJsonSchemaRef: JsonSchema[Trip]): JsonSchema[Assocation[Vehicle]] = ...
def associationActorSchema(tripJsonSchemaRef: JsonSchema[Trip]): JsonSchema[Association[Actor]] = ...

val tripJsonSchema: Record[Trip] = {
  def selfRef = lazyRecord(tripJsonSchema, "Trip")

  optField("vehicle")(associationVehicleSchema(selfRef)) zip
  optField("actors")(arrayJsonSchema(associationActorSchema(selfRef)))
}.xmap(…)(…)
~~~

But I agree that it’s not very convenient to write (ie., you can not just let the compiler summon the schema of the optional field `actors`).

What do you think?

bmeesters: But these calls will ensure that the schema will be evaluated right? I don't expect that to work with recursive calls. Or am I missing something?

bmeesters: Also, thinking about it more. I don't think it makes sense to have lazy constructors explicitly called `lazyRecord` and `lazyTagged` and not return a `Record` or `Tagged` respectively. They won't compose well with how tagged and record are used in other places. I think the encoding does this better than the current solution. Even though we might have to think a bit more on how to deal with other `LazySchema`s.

julienrf: > But these calls will ensure that the schema will be evaluated right? Since `optField` is strict in its argument.

`optField` is strict but it get passed a `lazyRecord`, which does not evaluate the underlying record schema until needed. So, I think this will work.

> I don't think it makes sense to have lazy constructors explicitly called `lazyRecord` and `lazyTagged` and not return a `Record` or `Tagged` respectively.

True. Actually, I don’t remember why I had to introduce this distinction from `lazySchema` to `lazyRecord` and `lazyTagged`. But now I think a single `lazySchema[A](toStrict: => JsonSchema[A]): JsonSchema[A]` should be enough.

bmeesters: > optField is strict but it gets passed a lazyRecord, which does not evaluate the underlying record schema until needed. So, I think this will work.

That makes sense. Though I am not sure I like the ergonomics. Because it is not just one schema I now have to pass explicitly, but all of them (every entity points to actions and every action can point back). I also don't really understand what would be against extending the current API. The code doesn't get much more complicated, and the way to use it seems IMO simpler. Maybe you can elaborate a little bit on that? Because it seems you are leaning towards closing this PR?

bmeesters: Also I want to point out that I am a little disappointed that after the work I put in this might just be thrown away.

_However_, I also don't think that is a good enough reason to merge it. It should have value to make these changes. So if you don't think it will, I will accept that. So, it's not necessary to feel guilty about just closing it.

julienrf: Hey @bmeesters, I’m sorry this is my fault, it’s only after seeing the PR that I realized this is _maybe_ not the best way to go.

Whatever we do with regards to this PR (closing it or merging it), we should make sure that we find a satisfactory solution for your original issue. In particular:

>  I am not sure I like the ergonomics. Because it is not just one schema I now have to pass explicitly, but all of them (every entity points to actions and every action can point back)

I agree with this. Since `field`, `optField` or `arrayJsonSchema` are type-directed (they take an *implicit* `JsonSchema`), one solution could be to provide a type-level `Lazy` operator:

~~~ scala
type Lazy[A]

def lazySchema[A](toStrict: => JsonSchema[A]): JsonSchema[Lazy[A]]
~~~

And then:

~~~ scala
val tripJsonSchema: Record[Trip] = {
  implicit def selfRef: JsonSchema[Lazy[Trip]] = lazySchema(tripJsonSchema, "Trip")

  optField[Lazy[Trip]]("nextTrip") zip
  optField[List[Lazy[Trip]]]("nextTrips")
}.xmap(…)(…)
~~~

I need to think a bit more about it, to see whether this is a viable solution or not. Do you have any opinions?

bmeesters: No problem @julienrf. This is always the benefit of hindsight. We are not our code and I am definitely in favor of finding the best solution. The `Lazy` type approach sounds promising. Though I can't oversee how that would impact for instance all the Json implementations.

I think for now, I'll try to make it work with whatever is currently available (we have some customers waiting on an OpenAPI specification). And its fine by me to let it rest for a bit so you have time to think about it. I would also be willing to work on the Lazy approach. Though I expect I need some pointers along the way.

bmeesters:
@julienrf I have experimented with it some more and you were right. The current API of endpoints is sufficient to model everything we need. Actually our internal API is almost the same compared with endpoints version in this PR. I didn't realize you could work with laziness as described by you above:

> optField is strict but it gets passed a lazyRecord, which does not evaluate the underlying record schema until needed. So, I think this will work.

So maybe the solution is just simply creating a lazyJsonSchema (generalize over tagged and record) and be done with it? For me, you can close this PR, as it turns out doesn't provide enough value.

julienrf:
Hey @bmeesters, thanks for the follow-up!

> So maybe the solution is just simply creating a lazyJsonSchema (generalize over tagged and record) and be done with it?

Can you please remind me why would this be necessary/useful?

bmeesters:
> Can you please remind me why would this be necessary/useful?

It is not strictly necessary, but after the previous discussion it doesn't seem necessary to have `lazyRecord` and `lazyTagged` and just have one `lazyJsonSchema`, in the same way there is just one `arrayJsonSchema` that works with all `JsonSchema`. You could then also implement both the lazy tagged and record functions in terms of this more general one if you want to stay backwards compatible.

julienrf:
Yeah sorry I was confused but I completely understand now and I agree!

bmeesters:
NP :smile: I don't really have time to pick it up now, but since it is more a 'nice to have' than a must I would say this can be closed, do you agree?

julienrf:
Yes I agree!

PR #480, opened 2020-02-27
Discard response entity if decoding response headers failed

julienrf:


Krever:
Out of curiosity, what is the motivation for this?

julienrf:
That's how the Akka HTTP client works, you always have to consume response entities, which I forgot to do in case the response headers were not correctly decoded. This caused a resource leak.

PR #489, opened 2020-03-04
http4s server interpreter

### LINKED ISSUE #283: Add http4s interpreter

julienrf:
Fixes #283.

Following up on the initial work done by @jkobejs and with the help of @hygt here is an http4s server interpreter!

Only the basic modules are supported (`Endpoints`, `BasicAuthentication` and `JsonEntities`). Support for other modules like `ChunkedEntities` could be added in a separate PR.

I plan to squash all the commits into a single one, if that’s fine with you, @jkobejs and @hygt?


jkobejs:
@julienrf I’m fine with that.
Tnx for taking care of this 🙂

PR #499, opened 2020-03-17
Create http4s client interpreter

LukaJCB:
So far everything compiles, but I haven't tested anything yet. Will add tests and documentation as I progress. I'd appreciate any comments on what I have so far :)


http4s/client/src/main/scala/endpoints/http4s/client/Urls.scala
```diff
@@ -0,0 +1,140 @@
...
+import endpoints.Validated
+import org.http4s.Query
+import endpoints.algebra.Documentation
+import org.http4s.ParseResult
+
+trait Urls extends endpoints.algebra.Urls with StatusCodes {
+
+  type QueryString[A] = A => Query
+
+  type QueryStringParam[A] = A => List[String]
```
julienrf: Since `QueryStringParam` definitions are implicit, they will have the undesired effect of being eligible as implicit conversions with this type definition. I suggest using a type definition that is unrelated with `Function`

http4s/client/src/main/scala/endpoints/http4s/client/Urls.scala
```diff
@@ -0,0 +1,140 @@
...
+          f: A => Validated[B],
+          g: B => A
+      ): B => List[String] =
+        b => fa(g(b))
+    }
+
+  override def stringQueryString: QueryStringParam[String] =
+    s => Uri.encode(s) :: Nil
+
+  type Segment[A] = A => Uri.Path
```
julienrf: Same here, I suggest using a type that is unrelated with `Function`

LukaJCB:
I think everything works now except for this one test: https://github.com/julienrf/endpoints/blob/master/algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala#L284

http4s' `Uri` will Url encode the `+` when rendering as a string and not sure what I can do to mitigate this.

julienrf:
IIUC correctly [this SO thread](https://stackoverflow.com/questions/1634271/url-encoding-the-space-character-or-20), both encodings are valid. I suggest that we update the test assertion to accept both forms. What do you think?

http4s/build.sbt
```diff
@@ -3,6 +3,7 @@ import EndpointsSettings._
 val `algebra-jvm` = LocalProject("algebraJVM")
 val `algebra-circe-jvm` = LocalProject("algebra-circeJVM")
 val `json-schema-circe-jvm` = LocalProject("json-schema-circeJVM")
+val `json-schema-generic-jvm` = LocalProject("json-schema-genericJVM")
```
julienrf: This seems unused

http4s/build.sbt
```diff
@@ -20,3 +21,19 @@ val `http4s-server` =
...
+
+val `http4s-client` =
+  project
+    .in(file("client"))
+    .settings(
+      publishSettings,
+      `scala 2.12 to latest`,
+      name := "endpoints-http4s-client",
+      libraryDependencies ++= Seq(
+        "org.http4s" %%% "http4s-client" % http4sVersion,
```
julienrf:
```suggestion
        "org.http4s" %% "http4s-client" % http4sVersion,
```

http4s/client/src/main/scala/endpoints/http4s/client/ChunkedEntities.scala
```diff
@@ -0,0 +1,31 @@
+package endpoints.http4s.client
+
+import org.http4s.{Request => Http4sRequest, Response => Http4sResponse}
+import fs2.Chunk
+
+trait ChunkedEntities
```
julienrf: Awesome! :heart_eyes: 
(we need to add it on the server side then!)

http4s/client/src/test/scala/endpoints/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
```diff
@@ -0,0 +1,58 @@
...
+    with algebra.BasicAuthenticationTestApi
+    with algebra.EndpointsTestApi
+    with algebra.JsonFromCodecTestApi
+    with circe.JsonFromCirceCodecTestApi
+    with circe.JsonEntitiesFromCodecs {}
+
+class Http4sClientEndpointsJsonSchemaTest
+    extends client.EndpointsTestSuite[TestJsonSchemaClient[IO]]
+    with client.BasicAuthTestSuite[TestJsonSchemaClient[IO]]
+    with client.JsonFromCodecTestSuite[TestJsonSchemaClient[IO]] {
```
julienrf: It would be nice to add `client.ChunkedJsonEntitiesTestSuite` as well. You will have to make `TestJsonSchemaClient` extend `ChunkedJsonEntities` and `algebra.ChunkedJsonEntitiesTestApi`

http4s/client/src/main/scala/endpoints/http4s/client/ChunkedEntities.scala
LukaJCB: I'll probably make a follow up for this soon :)

http4s/client/src/test/scala/endpoints/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
LukaJCB: Done

LukaJCB:
I'll get to the docs as soon as I can :)

LukaJCB:
Hmm, I can't seem to reproduce the CI failure locally :/

julienrf:
I see non-deterministic failures on the tests. I’m not sure what’s the cause.

http4s/client/src/main/scala/endpoints/http4s/client/Endpoints.scala
```diff
@@ -0,0 +1,268 @@
...
+        (to, req) => fa(g(to), req)
+
+    }
+
+  override def emptyRequest: RequestEntity[Unit] = (_, req) => req
+
+  override def textRequest: RequestEntity[String] =
+    (value, req) => req.withEntity(value)
+
+  type Request[A] = A => Effect[Http4sResponse[Effect]]
```
julienrf: The fact that a `Request[A]` is interpreted as “submitting a request” seems weird. Do you think something like `A => Http4sRequest[Effect]` would work? (and you would do the `submit` in the implementation of `endpoint` only.)

LukaJCB: I can definitely do this and it makes sense to me, though given that everything is lazy, I don't expect this to make a difference and therefore I can't imagine this being the source of the non-determinism

LukaJCB: Fixed this in the most recent commit :)

http4s/client/src/main/scala/endpoints/http4s/client/ChunkedEntities.scala
LukaJCB: #510

LukaJCB:
For some reason the build seems to be failing for 2.12.10, but I can't make out why. It works fine locally and seems to have worked fine on CI before.

LukaJCB:
I added some documentation, but something seems to be really off with Travis, I'm getting this compilation error that to me just doesn't make any sense at all! https://travis-ci.com/github/julienrf/endpoints/jobs/321226997#L637

It's all green on my machine and that error to me seems like some sort of cache problem maybe? No idea how to fix that, and I can't seem to restart the travis jobs either :/

julienrf:
@LukaJCB I think that’s because we have added a bunch of new operations on master. When the CI runs on the result of merging your branch on master it fails. You should be able to reproduce the problem by rebasing to master (or merging master to your branch). I thnik most of the errors you see were caused by the following PR: https://github.com/julienrf/endpoints/pull/525
I’m sorry about these breaking changes. We should try to get your work merged as soon as possible Unfortunately, I’ve been unable to stabilize the tests implementation so fa (#533 doesn’t work, we need something that actually fixes the problem…)

LukaJCB:
I merged the changes from master into my branch and updated the code :)

Have you been able to reproduce these failures locally? I haven't been able to so far.

LukaJCB:
And just now after I posted this I seem to get some actual failures on my side locally: `java.lang.IllegalStateException: This publisher only supports one subscriber`

algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
```diff
@@ -281,10 +281,6 @@ trait EndpointsTestSuite[T <: EndpointsTestApi] extends ClientTestBase[T] {
           ) shouldEqual "/foo?id=f4b9defa-1ad8-453f-9a06-2683b8564b8d"
         }
 
-        "escaping" in {
-          encodeUrl(path /? qs[String]("q"))("foo bar/baz") shouldEqual "?q=foo+bar%2Fbaz"
```
julienrf, line 285: Would it be possible to keep this test but change the assertion to accept both `"?q=foo+bar%2Fbaz"` and `?q=foo%20bar%2Fbaz` as discussed in https://github.com/julienrf/endpoints/pull/499#issuecomment-602168809 ?

LukaJCB:
All green! 😍

algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
LukaJCB: Oh yeah, sorry about that.

algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala
```diff
@@ -282,7 +282,9 @@ trait EndpointsTestSuite[T <: EndpointsTestApi] extends ClientTestBase[T] {
         }
 
         "escaping" in {
-          encodeUrl(path /? qs[String]("q"))("foo bar/baz") shouldEqual "?q=foo+bar%2Fbaz"
+          val encoded = encodeUrl(path /? qs[String]("q"))("foo bar/baz")
+          println(encoded)
+          (encoded == "?q=foo+bar%2Fbaz" || encoded == "?q=foo%20bar/baz") shouldEqual true
```
julienrf, line 287:
```suggestion
          assert(encoded == "?q=foo+bar%2Fbaz" || encoded == "?q=foo%20bar/baz")
```

PR #502, opened 2020-03-23
Support `title` and `description` on `JsonSchema`

### LINKED ISSUE #500: Support OpenAPI's `title` and `description`
In OpenAPI, one can attach a `title` and `description` to any schema. It would be useful if `endpoints` also allowed these to be specified for any schema. Note about `title`: unlike the "name" that `endpoints` allows users to customize (see `namedTagged`, `namedRecord`, etc.), `title` does not need to be a valid uri, which means it can have spaces and interesting characters in it.

AFAICT, this would be a pretty straightforward change. At a glance, probably not much more than:

  - [ ] add `title: Option[String]` and `description: Option[String]` fields to every variant of `DocumentedJsonSchema` (and to the trait itself, like `example: Option[ujson.Value]`)
  - [ ] add a `title: Option[String]` field to every variant of `openapi.Schema`
  - [ ] update `OpenApi.schemaJson` to properly propagate these fields into the generated OpenAPI schemas
  - [ ] add `withTitleJsonSchema[A](schema: JsonSchema[A], title: String): JsonSchema[A]` and `withDescriptionJsonSchema[A](schema: JsonSchema[A], description: String): JsonSchema[A]`, and implement it for all relevant interpreters

Is this is a feature `endpoints` would like to support and is the plan I've laid out above reasonable? If so, I'd be happy to open a PR for this.

PS. Thank you for this great project. The extensible trait-based approach is really nice!

harpocrates:
Extend the core `JsonSchemas` algebra to support setting a title and
description on schemas. This is implemented in the same way as schema
examples.

  * The main user facing change is that `JsonSchemas` now has two more
    methods: `withDescriptionJsonSchema` and `withTitleJsonSchema`

  * The OpenAPI model supports titles and descriptions on every schema.
    The OpenAPI interpreter fills these titles and descriptions in.

  * Other encoder/decoder intepreters ignore any descriptions or titles.

This new functionality is tested in `DescriptionsTest` and `TitlesTest`.

Fixes #500

### DIFF
--- a/json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
@@ -288,6 +288,18 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
   /** Include an example value within the given schema */
   def withExampleJsonSchema[A](schema: JsonSchema[A], example: A): JsonSchema[A]
 
+  /** Add a description to the given schema */
+  def withDescriptionJsonSchema[A](
+      schema: JsonSchema[A],
+      description: String
+  ): JsonSchema[A]
+
+  /** Add a title to the given schema */
+  def withTitleJsonSchema[A](
+      schema: JsonSchema[A],
+      title: String
+  ): JsonSchema[A]
+
   /**
     * A schema that can be either `schemaA` or `schemaB`.
     *
@@ -326,6 +338,23 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
       withExampleJsonSchema(schemaA, example)
 
     /**
+      * Include a description of what this schema represents. Documentation
+      * interpreters can show this description. Encoder and decoder interpreters
+      * ignore this description.
+      * @param description information about the values described by the schema
+      */
+    def withDescription(description: String): JsonSchema[A] =
+      withDescriptionJsonSchema(schemaA, description)
+
+    /**
+      * Include a title for the schema. Documentation interpreters can show
+      * this title. Encoder and decoder interpreters ignore the title.
+      * @param title short title to attach to the schema
+      */
+    def withTitle(title: String): JsonSchema[A] =
+      withTitleJsonSchema(schemaA, title)
+
+    /**
       * A schema that can be either `schemaA` or `schemaB`.
       *
       * Documentation interpreter produce a `oneOf` JSON schema.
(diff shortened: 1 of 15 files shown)


json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -362,6 +391,11 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
       * Give a name to the schema.
       * Documentation interpreters use that name to refer to this schema.
       * Encoder and decoder interpreters ignore the name.
+      *
+      * @note Names are used by documentation interpreters to construct
+      *       references and that the JSON schema specification requires these
```
julienrf:
```suggestion
      *       references and the JSON schema specification requires these
```

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
```diff
@@ -282,19 +338,28 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
         case (
-            OneOf(alternatives1, maybeExample1),
-            OneOf(alternatives2, maybeExample2)
+            OneOf(alternatives1, _, maybeExample1, _),
+            OneOf(alternatives2, _, maybeExample2, _)
             ) =>
           OneOf(
             alternatives1 ++ alternatives2,
-            maybeExample1.orElse(maybeExample2)
+            example = maybeExample1.orElse(maybeExample2)
```
julienrf: I see that we don’t treat `title` and `description` in the same way as we treat `example`.

Now I’m thinking that maybe it’s better to not move the example from the alternative to the `OneOf` (like you do with `title` and `description`). What do you think?

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -338,6 +341,7 @@ case class MediaType(schema: Option[Schema])
 sealed trait Schema {
   def description: Option[String]
```
julienrf: BTW, did you find any place where we used to set a `description`? I couldn’t find one.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
harpocrates: I think the behaviour for `example` is fine: an example for a variant is always going to also be an example for the `oneOf`. In fact, for `oneOf`, any example is always going to have to choose one specific alternative.

That's obviously not the case for a title or description (in fact, I'm struggling to imagine a case where a variant's title/description could generalize...).

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
harpocrates: Yeah, I think it used to be set in `withDefinedDescription`, which was used for setting field property docs and for enumeration descriptions.

openapi/openapi/src/main/scala/endpoints/openapi/JsonSchemas.scala
julienrf: OK. Let’s keep these things as they are for now!

julienrf:
Thanks @harpocrates !

PR #506, opened 2020-03-24
WIP: Improve annotation-driven generic derivation

### LINKED ISSUE #501: Support un-naming a schema
When making a `Record`, `Tagged`, or `Enum`, we can name it with `namedRecord`, `namedTagged`, and `namedEnum` (or the corresponding methods on the implicit classes). Naming a schema means that it will get its own section in the `schemas` part of the generated OpenAPI docs. Furthermore, schemas derived via `genericRecord`, `genericTagged`, and `genericJsonSchema` are _always_ named.

Having named types is sometimes undesirable (especially when a derived schema is used as a part of a larger schema). It would be nice to support _un-naming_ a schema:

```scala
trait JsonSchemas {

  /** Removes any name from the record JSON schema */
  def unnamedRecord[A](schema: Record[A]): Record[A]

  /** Removes any name from the tagged JSON schema */
  def unnamedTagged[A](schema: Tagged[A]): Tagged[A]

  /** Removes any name from the enum JSON schema */
  def unnamedEnum[A](schema: Enum[A]): Enum[A]

}
```

This would effectively allow users to control schema inlining in the generated documentation.

Thoughts? I'm trying to figure out how to make some schemas more palatable, but I suspect I'm missing some other constraints here...

### LINKED ISSUE #505: `withExampleJsonSchema` loses schema specificity
If I have a `Record[A]`, there is no way to attach an example to this record and still have a `Record[A]` afterwards. The reason for this is that the type of `withExampleJsonSchema` only talks about `JsonSchema`:

```scala
case class Foo(i: Int)
val fooSchema: Record[Foo] = genericRecord[Foo].withExample(Foo(4)) // compile error
```

This issue also holds for `withDescriptionJsonSchema` and `withTitleJsonSchema`.

harpocrates:
The end goal of this PR is to support generating schemas from annotated types like the following:

```scala
@docs("a hard or fluffy pet")
@name("Pet")
sealed trait Pet

@docs("felines")
@unnamed
case class Cat(
  @docs("the cat's name") name: String
) extends Pet

@name("Lizard")
@title("A lizard")
@docs("these pets often rest on shoulders")
case class Lizard(
  lovesRocks: Boolean
) extends Pet
```

This involved:

  * extending `withDescriptionJsonSchema`, `withTitleJsonSchema`, and `withExampleJsonSchema` to have variants operating specifically on `Record`, `Tagged`, and `Enum`

  * tweaking generic derivation to allow the `docs` annotation on classes and traits

  * tweaking generic derivation to make the name optional. Although the default is still to generate a name based on the class tag, there is now an `unnamed` annotation that can be used to prevent a schema from being named.

  * adding a `title` annotation, which is used to set the title of schemas for classes or traits

Fixes #501.
Fixes #505.

### DIFF
--- a/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
@@ -286,15 +286,60 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
     Record(encoder, decoder)
   }
 
+  def withExampleRecord[A](
+      schema: Record[A],
+      example: A
+  ): Record[A] = schema
+
+  def withExampleTagged[A](
+      schema: Tagged[A],
+      example: A
+  ): Tagged[A] = schema
+
+  def withExampleEnum[A](
+      schema: Enum[A],
+      example: A
+  ): Enum[A] = schema
+
   def withExampleJsonSchema[A](
       schema: JsonSchema[A],
       example: A
   ): JsonSchema[A] = schema
 
+  def withDescriptionRecord[A](
+      schema: Record[A],
+      description: String
+  ): Record[A] = schema
+
+  def withDescriptionTagged[A](
+      schema: Tagged[A],
+      description: String
+  ): Tagged[A] = schema
+
+  def withDescriptionEnum[A](
+      schema: Enum[A],
+      description: String
+  ): Enum[A] = schema
+
   def withDescriptionJsonSchema[A](
       schema: JsonSchema[A],
       description: String
   ): JsonSchema[A] = schema
+
+  def withTitleRecord[A](
+      schema: Record[A],
+      title: String
+  ): Record[A] = schema
+
+  def withTitleTagged[A](
+      schema: Tagged[A],
+      title: String
+  ): Tagged[A] = schema
+
+  def withTitleEnum[A](
+      schema: Enum[A],
+      title: String
+  ): Enum[A] = schema
 
   def withTitleJsonSchema[A](
       schema: JsonSchema[A],
--- a/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints/generic/JsonSchemas.scala
@@ -71,7 +71,8 @@ trait JsonSchemas extends algebra.JsonSchemas {
   object GenericJsonSchema
       extends GenericJsonSchemaLowPriority
       with GenericDiscriminatorNames
-      with GenericSchemaNames {
+      with GenericSchemaNames
+      with GenericDescriptions {
 
     implicit def emptyRecordCase: DocumentedGenericRecord[HNil, HNil] =
       (docs: HNil) => emptyRecord.xmap[HNil](_ => HNil)(_ => ())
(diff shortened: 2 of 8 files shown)


harpocrates:
This is WIP because I've yet to add any tests or documentation. @julienrf before I finish this PR, I wanted to double check that this is an approach you support. If not, I won't sink any more time into it. 😄 

I may also want to add an `example` annotation. I'm not yet sure that the annotation, derivation,  type inference, and failure modes there would all work out (the annotation would be generic).

harpocrates:
Swagger UI representation for the example from https://github.com/julienrf/endpoints/pull/506#issue-393273602:

Screen Shot 2020-03-24 at 5 55 34 PM

julienrf:
> before I finish this PR, I wanted to double check that this is an approach you support.

In general, I’m more in favor of manually defined schemas, but I understand that it requires a bit more boilerplate, that’s why I eventually added support for the `@docs` annotation and others.

I agree on the principle that extending the support to more than just the description of fields is useful. That’s quite a lot of additional code, though, and we will have to find a solution to make it compatible with Scala 3, eventually…

harpocrates:
Besides extending `withDescriptionJsonSchema`, `withTitleJsonSchema`, and `withExampleJsonSchema` to have variants operating specifically on `Record`, `Tagged`, and `Enum`, the rest of these changes can just as easily live outside _endpoints_. If the generic derivation tweaks aren't worth the maintenance burden, I don't mind rolling my own generic derivation in our project that uses _endpoints_.

For the time being, we've found that writing manual schemas is a bit too verbose. The duplicated field names and long chains of `orElse` followed by an `xmap` (with matches on things like `case Left(Left(Left(Right(s)))) => s`) are the most annoying bits.

Do you think you could make a final decision as to which subset of this PR (if any) is worth merging? That way I can finalize this PR and then decide on our internal strategy for deriving schemas.

julienrf:
> Besides extending `withDescriptionJsonSchema`, `withTitleJsonSchema`, and `withExampleJsonSchema` to have variants operating specifically on `Record`, `Tagged`, and `Enum`, the rest of these changes can just as easily live outside _endpoints_. If the generic derivation tweaks aren't worth the maintenance burden, I don't mind rolling my own generic derivation in our project that uses _endpoints_.

OK then let’s start with a PR that only refines the return type of `withDescription`, `withTitle` and `withExample`. We can see later to improve the generic derivation process to include all the features you proposed.

> For the time being, we've found that writing manual schemas is a bit too verbose. The duplicated field names and long chains of `orElse` followed by an `xmap` (with matches on things like `case Left(Left(Left(Right(s)))) => s`) are the most annoying bits.

Right. I remember that @bmeesters had an idea to improve this part. I’m happy to discuss about it :)

bmeesters:
Hello @julienrf and @harpocrates. We are using a solution that can get rid of all the unpacking of `Either`s. The downside is that it relies on `ClassTag`s, not sure if we can get rid of those. I do think it is worth it though, and we haven't had any problems with this solution yet.

The signature is:

```
// Yes, the name can be better..
def orTagged[A: ClassTag, C >: A, B <: C: ClassTag](taggedA: Tagged[A], taggedB: Tagged[B]): Tagged[C]
```

You can use it like this (where `or` is put in the implicit class and `Location` is the supertype):

```
  implicit lazy val locationSchema: Tagged[Location] =
    fromIdSchema or
      fromNameSchema

  val fromIdSchema: Tagged[Location.FromId] =
    field[Int]("id", Some("The unique ID of the location"))
      .xmap(Location.FromId)(Function.unlift(Location.FromId.unapply))
      .tagged("fromId")

  val fromNameSchema: Tagged[Location.FromName] =
    field[String]("name", Some("The name by which the location can be identified"))
      .xmap(Location.FromName)(Function.unlift(Location.FromName.unapply))
      .tagged("fromName")
```

Not that this will work with any arbitrary amount of sub-types.

Currently I do not have too much time to create a PR in endpoints. Though I think I can get some time next sprint (starting April 6th). If that is on time enough I can do it. Otherwise I would also be willing to just drop the most essential code here if that can help others now.

julienrf:
@bmeesters I think it would be very useful to have. We should properly document how it works and what are the limitations.

For instance, it can distinguish types based on their runtime type information only, not on what was erased. This means that a `List[Foo]` and a `List[Bar]` can’t be distinguished:

~~~ scala
val listFoo: JsonSchema[List[Foo]] = ...
val listBar: JsonSchema[List[Bar]] = ...
val schema = listFoo or listBar // warning you won’t be able to decode `List[Bar]` values
~~~

(that being said, that use case should not happen much, I believe…)

About the name, IIRC we mentioned `orElseMerge`?

bmeesters:
@julienrf, I can do the implementation and also the documentation (from 6th of April) though I need some pointers on where this should go.

Yes, `orElseMerge` was indeed the better suggestion, so I also think we should go with that.

harpocrates:
The proposed type signature of `orElseMerge` sounds like it could also lead to more efficient encoder/decoder implementations. @bmeesters let me know if there is anything I can do to help here - having this is something we would very much like!

julienrf:
@bmeesters I’ve created the issue #515. Let’s continue the discussion there :)

julienrf:
I think you managed to get all the features of this PR merged in other small PRs, didn’t you?

harpocrates:
Yep! Thank you for dealing with all of those small PRs so promptly. 😄

PR #517, opened 2020-03-27
`docs` annotation works on classes/traits

harpocrates:
The `docs` annotation will now work on classes and traits as well as
fields. When specified, it will set the description of the matching
schema.

Also tweaked the `generic.JsonSchemas` test case to check the new
annotation locations too.

### DIFF
--- a/json-schema/json-schema-generic/src/main/scala/endpoints/generic/annotations.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints/generic/annotations.scala
@@ -1,12 +1,12 @@
 package endpoints.generic
 
 /**
-  * Documents a case class field.
+  * Adds a description to a case class field, a case class, or a sealed trait.
   *
-  * Annotate a case class field with this annotation to define its
-  * documentation.
+  * Annotate a case class field, case class, or sealed trait with this
+  * annotation to define its the description of the schema or record field.
   *
-  * @param text Description of the annotated field
+  * @param text Description of the annotated schema or field
   */
 case class docs(text: String) extends scala.annotation.Annotation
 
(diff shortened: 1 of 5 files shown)


harpocrates:
If this is not functionality that _endpoints_ wants to maintain, that's perfectly fine too.

This particular annotation is probably the most useful one for us: we have a good number of small sealed traits and we'd like to document the variants, but we don't want to have to write repetitive boilerplate sealed trait schemas explicitly just so we can attach descriptions to the schemas.

json-schema/json-schema-generic/src/main/scala/endpoints/generic/annotations.scala
```diff
@@ -1,12 +1,12 @@
 package endpoints.generic
 
 /**
-  * Documents a case class field.
+  * Adds a description to a case class field, a case class, or a sealed trait.
   *
-  * Annotate a case class field with this annotation to define its
-  * documentation.
+  * Annotate a case class field, case class, or sealed trait with this
+  * annotation to define its the description of the schema or record field.
```
julienrf:
```suggestion
  * annotation to define its description.
```

julienrf:
```suggestion
  * annotation to define the description of the schema or the description of the record field.
```
? Sounds heavy, no?

harpocrates: Oof yeah - my docstring does not read well. How about:

```
Annotate a case class field, case class, or sealed trait with this
annotation to set a description for the schema or the record field.
```

julienrf: Yes, that one is great!

harpocrates: I've updated this doc (as well as tweaking the mention of `@docs` in `documentation/`). Let me know if there is anything else you think we should change.

PR #518, opened 2020-03-27
Add `title` annotation

harpocrates:
The new `title` annotation sets the title of a schema in the generic
derivation process. This is likely to be more useful than the `name`
annotation, since the latter is actually the internal OpenAPI name,
which has to be a valid URI suffix.

The `name` annotation's scaladoc now includes a note about OpenAPI's
use and a reference to `title`.


harpocrates:
As with #517, I'll understand if this is not functionality _endpoints_ wants to maintain.

Personally, I do think this is worth it though - it is hardly any actual code, and setting the title is something I want to do on pretty much all schemas (end users should not have to slog through the internal package hierarchies when browsing generated documentation).

PR #521, opened 2020-03-31
Add `unnamed` annotation

### LINKED ISSUE #501: Support un-naming a schema

harpocrates:
Tweak generic derivation to make the name optional. Although the default
is still to generate a name based on the class tag, there is now an
`unnamed` annotation that can be used to prevent a schema from getting a
name at all.

Fixes #501

### DIFF
--- a/documentation/manual/src/paradox/algebras/json-schemas.md
+++ b/documentation/manual/src/paradox/algebras/json-schemas.md
@@ -227,7 +227,8 @@ type. The rules for deriving the schema are the following:
   annotation, otherwise the `defaultDiscriminatorName` value is used,
 - the schema is named by the `@name` annotation, if present, or by invoking the
   `classTagToSchemaName` operation with the `ClassTag` of the type for which the schema
-  is derived.
+  is derived. If you wish to avoid naming the schema, use the `@unnamed` annotation
+  (unnamed schemas get inlined in their OpenAPI documentation).
 - the schema title is set with the `@title` annotation, if present
 
 Here is an example that illustrates how to configure the generic schema derivation process:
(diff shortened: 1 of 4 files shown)


harpocrates:
Summarizing some thoughts from elsewhere: right now, there is no way to avoid a generic schema being named (you either specify your own name with `@name`, or you get a derived name from the class tag).

The benefit of unnamed schemas is that they get inlined in OpenAPI clients like Swagger UI. This is particularly useful for:

  - ADTs where the case classes make more sense being inlined into the sealed trait (than each having their own top-level schema that the sealed trait refers to).
  - highly nested records, where some of the inner records don't warrant top-level schema descriptions

PR #533, opened 2020-04-11
Fix CI flakiness

julienrf:
With the help of @LukaJCB, this will hopefully fix the flakiness of the tests.


julienrf:
Only one build out of 18 failed. That’s better than the current state, but still not stable enough :-/

I’m really curious what could be the cause. Looking at [the code](https://github.com/http4s/http4s/blob/3bf89ded091ee55511c409ef83ebc63715fedef1/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServerBuilder.scala#L342) my guess is that the `ServerChannel` is not fully closed when the corresponding cats-effect resource is closed.

PR #535, opened 2020-04-12
Add `xmapPartialAlong` variant of `xmapPartial`

### LINKED ISSUE #528: Variant of `xmapPartial` that takes a `Codec` argument
Might be useful to have an `xmapPartialAlong` method on `PartialInvariantFunctorSyntax` that takes just a `Codec` as an argument: `fa.xmapPartialAlong(codec)`. The name `xmapPartialAlong` is not very good though.

harpocrates:
This variant takes a `Codec` instead of two functions.

Fixes #528

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Codec.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Codec.scala
@@ -1,6 +1,6 @@
 package endpoints.algebra
 
-import endpoints.Validated
+import endpoints.{PartialInvariantFunctor, PartialInvariantFunctorSyntax, Validated}
 
 /**
   * A way to decode a `From` value into a `To` value.
@@ -23,6 +23,8 @@ object Decoder {
   )(bc: Decoder[B, C]): Decoder[A, C] =
     from => ab.decode(from).flatMap(bc.decode)
 
+  /** Produce a decoder that just outputs its input value */
+  def identity[A]: Decoder[A, A] = identity[A]
 }
 
 /**
@@ -42,6 +44,8 @@ object Encoder {
   )(bc: Encoder[B, C]): Encoder[A, C] =
     from => bc.encode(ab.encode(from))
 
+  /** Produce an encoder that just outputs its input value */
+  def identity[A]: Encoder[A, A] = identity[A]
 }
 
 /**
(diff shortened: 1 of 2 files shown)


harpocrates:
@julienrf I'm still not clear on exactly how to best test this. Could you advise:

  * should the testsuite test codec-related operations?
  * should the testsuite test functor-related operations?
  * in order to call `xmapPartialAlong`, one needs a functor. Should the test suite make such a functor instance (eg. for `Option`)? It doesn't seem like the `algebra` library defines any actual instances.

julienrf:
What about implementing `xmapPartialAlong` in `PartialInvariantFunctor` (and also have it in `PartialInvariantFunctorSyntax`)? Something like:

~~~ scala
final def xmapPartialAlong[A, B](fa: F[A], codec: Codec[A, B]): F[B] =
  xmapPartial(fa, codec.decode, codec.encode)
~~~

You can test it in a similar way we already test `xmapPartial`, with one specific instance of `PartialInvariantFunctor` such as the one for `QueryString`.

By the way, the name `xmapPartialAlong` is not very meaningful to me. I would rather use the same name (`xmapPartial`) or maybe something like `xmapWithCodec`?

harpocrates:
> What about implementing xmapPartialAlong in PartialInvariantFunctor (and also have it in PartialInvariantFunctorSyntax)? Something like:

`Codec` is only defined in `algebra`, whereas `PartialInvariantFunctor` is in `json-schema`. Unless we move `Codec` to `json-schema`, I see no way to define `xmapPartialAlong` inside `PartialInvariantFunctor`/`PartialInvariantFunctorSyntax`. I think moving `Codec` to `json-schema` is fine, but what do you think?

> You can test it in a similar way we already test xmapPartial, with one specific instance of PartialInvariantFunctor such as the one for QueryString.

👍 

> By the way, the name xmapPartialAlong is not very meaningful to me. I would rather use the same name (xmapPartial) or maybe something like xmapWithCodec?

Woops! Sorry about that - I prefer your name too, just forgot to update my code.

julienrf:
> I think moving `Codec` to `json-schema` is fine, but what do you think?

Sure, no problem.

harpocrates:
Superseded by #538

PR #536, opened 2020-04-12
Add `NoDocsJsonSchemas` for decoder `JsonSchemas`

### LINKED ISSUE #514: Trait to mix into `algebra.JsonSchemas` to no-op documentation operations
Interpreters for `algebra.JsonSchemas` which are just for encoding/decoding share a bunch of boilerplate code to no-op implement operations that are documentation specific. It would be cleaner to add some trait to implement these no-op operations only once (and then mix that trait into all encoder/decoder implementations).

Example duplication:

  * https://github.com/julienrf/endpoints/blob/5c9c60ce74db13ee3c0703a7de68f5ae96739906/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala#L289-L302

  * https://github.com/julienrf/endpoints/blob/5c9c60ce74db13ee3c0703a7de68f5ae96739906/json-schema/json-schema-playjson/src/main/scala/endpoints/playjson/JsonSchemas.scala#L160-L173

  * https://github.com/julienrf/endpoints/blob/5c9c60ce74db13ee3c0703a7de68f5ae96739906/openapi/openapi/src/main/scala/endpoints/ujson/JsonSchemas.scala#L241-L254

Something that looks like:

```scala
trait NoDocsJsonSchemas extends algebra.JsonSchemas {
  def namedRecord[A](schema: Record[A], name: String) = schema
  def namedTagged[A](schema: Tagged[A], name: String) = schema
  def namedEnum[A](schema: Enum[A], name: String) = schema
  def withExampleJsonSchema[A](schema: JsonSchema[A], e: A) = schema
  def withDescriptionJsonSchema[A](schema: JsonSchema[A], d: String) = schema
  def withTitleJsonSchema[A](schema: JsonSchema[A], t: String) = schema
}
```

If #511 gets merged, we could throw in those 9 new methods too.

harpocrates:
The new `NoDocsJsonSchemas` trait implements all the documentation
related methods of `JsonSchemas` as no-ops. This reduces the boilerplate
in decoder inteprereters: they can mix in `algebra.NoDocsJsonSchemas`
instead of having to implement a bunch of pass-through documentation
methods.

This factors out a good chunk of common code in the Circe, Play, and
Ujson interpreters.

Fixes #514.

### DIFF
--- a/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints/circe/JsonSchemas.scala
@@ -9,7 +9,7 @@ import scala.collection.compat._
 /**
   * An interpreter for [[endpoints.algebra.JsonSchemas]] that produces a circe codec.
   */
-trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
+trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
 
   trait JsonSchema[A] {
     def encoder: Encoder[A]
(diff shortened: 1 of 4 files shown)


harpocrates:
Should I keep force-pushing to re-trigger CI until it passes? It seems like the 2.12 CI regularly has spurious failures around the `http4s-server` tests (I tried and failed to reproduce those locally).

julienrf:
No, we should fix the actual issue. We don’t need the CI to be green to merge the PR.

PR #537, opened 2020-04-12
Expand documentation

julienrf:
- Add a `release-and-compatibility-notes` page,
- Polish the documentation of the main algebras and their code examples,
- Improve the API documentation by adding more links between types and operations and by adding a specification of the expected behavior for client, server, and documentation interpreters, when relevant.

Also: make a couple of internal operations private.


algebras/algebra/src/main/scala/endpoints/package.scala
```diff
@@ -0,0 +1,7 @@
+/**
+  *   - '''algebras''' are in the [[algebra]] package,
+  *   - '''interpreters''' are in the other packages (e.g., [[akkahttp]],
```
harpocrates, line 3: Not sure if this is an issue, but `sbt doc` does not manage to generate links for `[[akkahttp]]`, `[[http4s]]`, and `[[openapi]]`.

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
```diff
@@ -486,18 +553,45 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
...
     def orElse[B](taggedB: Tagged[B]): Tagged[Either[A, B]] =
       choiceTagged(taggedA, taggedB)
 
+    /** Define a schema that alternatively accepts `taggedA` or `taggedB` and merges the result
+      *
+      * Similar to `orElse` but instead of returning a `Tagged[Either[A, B]]`, it returns
+      * a `Tagged[C]`, where `C` is a super type of both `A` and `B`.
+      *
+      *   - Encoder interpreters forward to `taggedA` or `taggedB` based on the runtime type
+      *     information of the value to encode,
```
harpocrates, line 574: I think it is necessary to specify that `B` is tried before `A` during encoding, since the natural chained-method use

```
barSchema.tagged("Bar") orElseMerge
  bazSchema.tagged("Baz") orElseMerge
  baxSchema.tagged("Bax") orElseMerge
  quxSchema.tagged("Qux") orElseMerge
  quuxSchema.tagged("Quux")
``` 

parses as

```
(((barSchema.tagged("Bar").orElseMerge(
  bazSchema.tagged("Baz"))).orElseMerge(
  baxSchema.tagged("Bax"))).orElseMerge(
  quxSchema.tagged("Qux"))).orElseMerge(
  quuxSchema.tagged("Quux"))
``` 

which means every single `orElseMerge` fixes `C =:= Foo`. This would be quite problematic if `orElseMerge` tried `A` before `B`.

algebras/algebra/src/main/scala/endpoints/package.scala
```diff
@@ -0,0 +1,7 @@
...
+  *     [[http4s]], [[openapi]], etc.),
+  *   - Types defined in this package are utility classes (e.g., [[Validated]], [[Tupler]]).
```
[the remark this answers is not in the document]
julienrf: Definitely. I was thinking of doing that at some point.

algebras/algebra/src/main/scala/endpoints/package.scala
julienrf: Thanks for checking that! It works with sbt-unidoc (which is what is used when you run the task `manual/previewSite`)

harpocrates: Oh nice! Good to know - I'll check that in the future.

json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala
julienrf: Interesting, I didn’t realize that until now!

However, if I swap the [two lines](https://github.com/julienrf/endpoints/blob/7fa43f59d7210c31ab192b71d36d1107ceb1ff46/json-schema/json-schema/src/main/scala/endpoints/algebra/JsonSchemas.scala#L293-L294) the tests still pass… Now I’m confused!

harpocrates: > However, if I swap the two lines the tests still pass… Now I’m confused!

That's because the tests aren't thorough enough 😠. Applying this diff to the tests and then swapping `case a: A`/`case b: B`, you should be able to trigger a failure.

```diff
diff --git a/json-schema/json-schema-circe/src/test/scala/endpoints/circe/JsonSchemasTest.scala b/json-schema/json-schema-circe/src/test/scala/endpoints/circe/JsonSchemasTest.scala
index b0910628..1ea23d3c 100644
--- a/json-schema/json-schema-circe/src/test/scala/endpoints/circe/JsonSchemasTest.scala
+++ b/json-schema/json-schema-circe/src/test/scala/endpoints/circe/JsonSchemasTest.scala
@@ -10,7 +10,7 @@ class JsonSchemasTest extends AnyFreeSpec {
       extends algebra.JsonSchemasFixtures
       with circe.JsonSchemas
 
-  import JsonSchemasCodec.{User, Foo, Bar}
+  import JsonSchemasCodec.{User, Foo, Bar, Qux}
 
   "case class" in {
     val userJson =
@@ -69,6 +69,14 @@ class JsonSchemasTest extends AnyFreeSpec {
     assert(Foo.alternativeSchemaForMerge.decoder.decodeJson(barJson).exists(_ == bar))
     assert(Foo.alternativeSchemaForMerge.encoder.apply(bar) == barJson)
 
+
+    val quxJson =
+      Json.obj(
+        "type" -> Json.fromString("Qux")
+      )
+    assert(Foo.alternativeSchemaForMerge.decoder.decodeJson(quxJson).exists(_ == Qux))
+    assert(Foo.alternativeSchemaForMerge.encoder.apply(Qux) == quxJson)
+
     assert(
       Foo.alternativeSchemaForMerge.decoder
         .decodeJson(Json.obj())
```

julienrf: Thanks!

I’ve added more details to the note and I’ve also included your test.

Let me know what you think.

harpocrates: Looks great! Thanks for also adding the test case. I think this is the last suggestion I had - everything looks merge ready to me.

julienrf:
Thanks for the detailed review @harpocrates

PR #538, opened 2020-04-12
Add `xmapWithCodec` variant of `xmapPartial`

### LINKED ISSUE #528: Variant of `xmapPartial` that takes a `Codec` argument

harpocrates:
This variant takes a `Codec` instead of two functions. This is
accompanied with a couple other smaller changes that seemed natural:

  * `Codec` is moved into `json-schema` (so that `xmapWithCodec` can be
    defined in the `PartialInvariantFunctor` trait)

  * the UUID/Int/Long/Double/Boolean for query strings and segments are
    now defined using `xmapWithCodec` and a shared set of codecs

  * added a `tryParseString` method to the `Codec` companion object for
    easily producing a `Codec[String, A]` given parsing/printing
    functions and use this to create UUID/Int/Long/Double/Boolean
    codecs.

Supersedes #535 
Fixes #528

### DIFF
--- a/json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
@@ -67,20 +67,14 @@ object Codec {
       Decoder.sequentially(ab)(bc)
     )
 
-  /** Produce a codec that passes values through without touching them */
-  def identity[A]: Codec[A, A] = new Codec[A, A] {
-    def decode(from: A): Validated[A] = Valid(from)
-    def encode(from: A): A = from
-  }
-
-  /** Produce a codec to/from a string. An [[Invalid]] message mentioning the
-    * type name is produced when parsing fails.
+  /** Produce a codec to/from a string. If the parsing function fails, the
+    * decoding output is an [[Invalid]] with a message mentioning the type name.
     *
     * @param type name of the type being decoded
-    * @param parse parsing function to use
-    * @param print printing function to use
+    * @param parse parsing function to use, with exceptions turned into [[Invalid]]
+    * @param print printing function to use, not supposed to throw exceptions
     */
-  def tryParseString[A](
+  def parseStringCatchingExceptions[A](
       `type`: String,
       parse: String => A,
       print: A => String = (x: A) => x.toString()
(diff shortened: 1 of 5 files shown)


json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
```diff
@@ -68,4 +67,40 @@ object Codec {
...
+  }
+
+  /** Produce a codec to/from a string. An [[Invalid]] message mentioning the
+    * type name is produced when parsing fails.
+    *
+    * @param type name of the type being decoded
+    * @param parse parsing function to use
+    * @param print printing function to use
+    */
+  def tryParseString[A](
```
julienrf, line 83: I was considering making this method private. (See https://github.com/julienrf/endpoints/pull/537/files#diff-26b01f5a906ec9633108d28cfd1aabb1L161-R175)

What do you think?

json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
```diff
@@ -68,4 +67,40 @@ object Codec {
       Decoder.sequentially(ab)(bc)
     )
 
+  /** Produce a codec that passes values through without touching them */
+  def identity[A]: Codec[A, A] = new Codec[A, A] {
```
julienrf, line 71: Is this one used somewhere?

json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
harpocrates: That is reasonable - it does seem a bit ad-hoc.

Somewhat ironically, my company _is_ using `tryParseString` in our codebase, so we may end up needing to duplicate some part of this function if it becomes private.

Let me know what you prefer and I'll adjust accordingly.

julienrf: We could leave it public but I would like to rename it to something that includes “catching”. For instance `parseStringCatching[A]`? And we should document the fact that exceptions thrown during parsing are caught and turn into an `Invalid` result.

json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
harpocrates: No, it isn't. I figured it might start being useful now that `Codec` can be used as an argument in more places and that the cost of including it is relatively low.

Motivating example: in my companies codebase, we have a trait somewhat like:

```scala
trait CompExt extends algebra.Endpoints {

  /** The decoded type of binary data */
  type BStr

  /** Since binary data gets encoded as Base64 strings, it is necessary to
    * define the serialization/deserialization to/from byte arrays
    */
  def byteStringCodec: endpoints.algebra.Codec[Array[Byte], BStr]

  /** Schema for binary data encoding (uses base64 encoded strings) */
  implicit lazy val byteStringSchema: JsonSchema[BStr] = {
    val enc = java.util.Base64.getEncoder
    val dec = java.util.Base64.getDecoder

    stringJsonSchema(format = Some("byte"))
      .withExample("abcdef")
      .xmapPartial(tryParseString("base64 string")(dec.decode(_)))(enc.encodeToString(_))
      .xmapWithCodec(byteStringCodec)
  }
}
```

Then, the Scala backend may implement this with `type BStr = akka.util.ByteString` while the Scala.js front-end (which doesn't have an Akka dependency) sticks with `type BStr = Array[Byte]`. Then it is handy to be able to do `val byteStringCodec = Codec.identity[Array[Byte]]` in the client.

Shall I take it out for now?

json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
```diff
@@ -68,4 +67,34 @@ object Codec {
...
     )
 
+  /** Produce a codec to/from a string. If the parsing function fails, the
+    * decoding output is an [[Invalid]] with a message mentioning the type name.
+    *
+    * @param type name of the type being decoded
+    * @param parse parsing function to use, with exceptions turned into [[Invalid]]
+    * @param print printing function to use, not supposed to throw exceptions
+    */
+  def parseStringCatching[A](
```
julienrf, line 77: Sorry to be nitpicking, but what about `parseStringCatchingExceptions` instead?

json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
julienrf: I see that you already removed it. Fine, we can re-introduce it if you feel it should be there :)

json-schema/json-schema/src/main/scala/endpoints/algebra/Codec.scala
harpocrates: That's actually a good name 👍. Done.

julienrf:
Thanks!

Issue #529, opened 2020-04-07
Multiple possible request types for an endpoint


harpocrates:
It would be useful to allow multiple different types of request types. For instance `jsonRequest[Foo].orElse(textRequest): RequestEntity[Either[Foo, String]]`. However, I'm not sure how to statically rule out cases where the two request types are actually the same (ex: `jsonRequest[Foo].orElse(jsonRequest[Bar])`). Maybe it would suffice to document this outside the type system?

I know OpenAPI supports documenting multiple request types for an endpoint.

julienrf:
That makes sense.

> However, I'm not sure how to statically rule out cases where the two request types are actually the same (ex: `jsonRequest[Foo].orElse(jsonRequest[Bar])`)

In that specific case, maybe `jsonRequest(fooSchema orElse barSchema)`?

harpocrates:
Any chance this could make it into 1.0.0? @ethanb29 has an implementation with a couple of the interpreters already done.

julienrf:
Yes.

julienrf:
Hey @harpocrates, any update on this?

harpocrates:
I think #544 is the only difficult part. After that, this should be quick straightforward to implement.

harpocrates:
I may be missing something, but I think this will be impossible to implement for Play's `BodyParser` (there doesn't seem to be a way to disentangle the parsing from the validation). I've got all the other cases implemented though. Assuming this really is impossible for Play, I think that leaves two options:

  * make some trait extending `Requests` and implement it in all clients/servers `Endpoints` implementation, except for Play
  
    ```scala
    trait EitherRequestEntities extends Requests {
      def choiceRequestEntity[A, B](
        requestEntityA: RequestEntity[A],
        requestEntityB: RequestEntity[B]
      ): RequestEntity[Either[A, B]]

      implicit class RequestEntitySyntax[A](requestEntity: RequestEntity[A]) {
        def orElse[B](otherRequestEntity: RequestEntity[B]): RequestEntity[Either[A, B]] =
          choiceRequestEntity(requestEntity, otherRequestEntity)
      }
    }
    ```

  * wontfix this issue

@julienrf are you OK with the first option? If so, do you have opinions for the names?

julienrf:
[...]

PR #539, opened 2020-04-13
Add `plainTextRequest` and clarify `textRequest`

### LINKED ISSUE #534: Akka-HTTP textRequest matches any Content-Type
[akkahttp.server.EndpointsWithCustomErrors#textRequest  ](https://github.com/julienrf/endpoints/blob/master/akka-http/server/src/main/scala/endpoints/akkahttp/server/Endpoints.scala#L117) matches any `Content-Type` header, while [openapi.Requests#textRequest](https://github.com/julienrf/endpoints/blob/master/openapi/openapi/src/main/scala/endpoints/openapi/Requests.scala#L45) documents the algebra as only matching `text/plain`.

Is this intended behavior? If so, is the rationale documented anywhere?

harpocrates:
The newly added `plainTextRequest` is a request that must have type
`text/plain`. The semantics of `textRequest` have also been clarfied.
The important bits are:

  * it can have any content type
  * if the content-type specifies some charset, that charset gets used

The server implementations of `textRequest` have been adjusted to all
match these semantics - it turned out that the servers were all doing
different things!

  * Akka-HTTP was doing the "right" thing
  * http4s was always decoding the body as UTF-8, regardless of charset
  * Play was not accepting content-types different that `text/plain`

The server semantics of `testRequest`/`plainTextRequest` are tested in
the new `TextEntities` test suite.

Fixes #534

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
@@ -112,9 +112,35 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
 
   /**
     * Request with a `String` body.
+    *
+    *   - Server interpreters are expected to be permissive and accept multiple
+    *     content types (not just `text/plain`) using the charset defined in the
+    *     `Content-Type` header to decide how to decode the body. If no charset
+    *     is defined (or if there is no `Content-Type` header) the interpreters
+    *     are free to define their own defaults for decoding.
+    *
+    *   - Client interpreters produce an HTTP request with a `text/plain` content type.
+    *
+    * @see [[plainTextRequest]] if you want to avoid accepting any content type
     * @group operations
     */
   def textRequest: RequestEntity[String]
+
+  /**
+    * Request with content-type `text/plain`
+    *
+    *   - Server interpreters raise an error if the incoming request entity does
+    *     not have the right content type. By default, they produce a Bad Request (400)
+    *     response with a list of error messages in a JSON array. Refer to the documentation
+    *     of your server interpreter to customize this behavior.
+    *
+    *   - Client interpreters produce an HTTP request with a `text/plain` content type.
+    *
+    * @see [[textRequest]] for a more permissive text request entity.
+    *
+    * @group operations
+    */
+  def plainTextRequest: RequestEntity[String]
 
   /**
     * Request for given parameters
--- a/akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
+++ b/akka-http/client/src/main/scala/endpoints/akkahttp/client/Endpoints.scala
@@ -134,6 +134,8 @@ trait EndpointsWithCustomErrors
   lazy val textRequest: (String, HttpRequest) => HttpRequest =
     (body, request) => request.copy(entity = HttpEntity(body))
 
+  lazy val plainTextRequest = textRequest
+
   def request[A, B, C, AB, Out](
       method: Method,
       url: Url[A],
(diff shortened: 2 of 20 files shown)


algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
```diff
@@ -112,10 +112,36 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
...
+    *     response with a list of error messages in a JSON array. Refer to the documentation
+    *     of your server interpreter to customize this behavior.
+    *
+    *   - Client interpreters produce an HTTP request with a `text/plain` content type.
+    *
+    * @see [[textRequest]] for a more permissive text request entity.
+    *
+    * @group operations
+    */
+  def plainTextRequest: RequestEntity[String]
```
julienrf, line 143: Thanks a lot for clarifying the behavior of both operations.

Do we really need to have both `textRequest` and `plainTextRequest` in the library? It seems to me that we should pick one, and let library users add the other on their side, if they want to.

harpocrates: If we were to pick one, I'd pick `plainTextRequest`. The deciding factor that convinced me to keep `textRequest` too is: I think there should still be a way to set up an endpoint that  accepts test requests that don't specify any content-type. `plainTextRequest` _requires_ `text/plain`, and that requirement is necessary for accurate OpenAPI documentation and `.orElse` for request entities.

If you agree that we want a way to accept test requests that don't specify content-type, the rest of the specification for `textRequest` follows pretty naturally.

harpocrates:
Alternate design:

```
def textRequest(isStrict: Boolean = false): RequestEntity[String]
def jsonRequest(isStrict: Boolean = false): RequestEntity[String]
```

if `isStrict = true`, then 415 is returned whenever a request does not pass in a request with exactly the right content-type header. Otherwise, we are lenient (maybe `jsonRequest` always at least asserts that, when specified, the content-type is not _different_ than `application/json`).

algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
julienrf: What do you think of just changing the existing behavior of `textRequest` to be more strict (and do the same for `jsonRequest`)? We can see later if it becomes really important to also have a more forgiving behavior for servers. I initially thought it could be a good default, but I think supporting #529 is more important so maybe it’s better to just change the default behavior.

harpocrates: I'm in favour of such a change. Forcing content-types everywhere will mean that it is much harder to use #529 in a way that doesn't make sense.

But just to re-iterate: taking this path means requests without a content-type will always fail. Is that OK with you?

julienrf: Would it be possible to require a `Content-Type` _only_ if there are several possible request entity types, but not if there is only one?

harpocrates: Yeah, I like that. It is probably going to complicate the servers a bit (they'll internally need to keep track of whether a `RequestEntity` is strict due to being part of an `orElse`), but that's fine.

Shall we do this? If so, I'm going to close this PR and open another to focus on `Content-Type` and the character encoding issues.

julienrf: > Shall we do this? If so, I'm going to close this PR and open another to focus on `Content-Type` and the character encoding issues.

Yes, that would be awesome! Thank you for the fruitful discussion.

harpocrates: > Thank you for the fruitful discussion.

Likewise 😄

Issue #352, opened 2019-09-05
Akka-http server interpreter does not embrace rejections


julienrf:
Errors are sometimes reported as rejections (e.g. when a marshaller fails), but sometimes reported as [completions](https://github.com/julienrf/endpoints/blob/dd97d792a1d9626e88b8f5403526eb47bda45b58/akka-http/server/src/main/scala/endpoints/akkahttp/server/Urls.scala#L192).

We should report errors as rejections as much as possible, so that users can handle them in their rejection handlers.

julienrf:
I have been thinking a bit about this and my problem with the concept of rejections in Akka HTTP is that when a route is rejected, that rejection might be silently ignored if another route completes:

~~~ scala
val route1 =
  path("foo") {
    parameter("x".as[Int]) { x =>
      complete("")
    }
  }
val route2 =
  path("foo") {
    complete("")
  }
val routes = route1 ~ route2
~~~

In my opinion, if a client requests `/foo?x=forty-two`, the `route1` route should complete with a Bad Request response, rather than rejecting the request, letting the remaining part of the routes tree (here, just `route2`) handle it.

You can see it in action here: https://scastie.scala-lang.org/NIz6ANVMRuKtjE4Ycz5CZA

In the current state of #370, routes immediately complete with a response in such cases, which makes more sense, I believe. Said otherwise, that PR gives *less* chance to work with rejection handlers.

Letting rejections flow can lead to two situations:

1. one of the subsequent routes does complete, silently swallowing the rejection error (which is not desirable),
2. or, none of the subsequent routes does match the incoming request, but in that case, why spend time trying the remaning routes?

In both cases, this makes no sense to me.

I think the problem is that Akka HTTP uses rejections to signify **both** that a request does not match an endpoint (eventually leading to returning a 404 response) and that a request is invalid (in that case we want to immediately return a 400 response). I believe those two situations should be handled by different abstractions.

So, my current position is to still **not** embrace rejections. But I’m happy to be proven wrong ;-)

julienrf:
With #382, the equivalent of rejection handlers is the following:

~~~ scala
  // This is the _endpoints_ equivalent to Akka HTTP exception handlers
  override def handleServerError(throwable: Throwable): StandardRoute =
    StandardRoute {
      Directives.extractUri { uri =>
        val message = s"Request to $uri could not be handled normally ${throwable.getMessage}"
        logger.error(message, throwable)
        // The provided implementation of `serverErrorResponse` completes
[...]

PR #544, opened 2020-04-19
Use and set `Content-Type` header uniformly

harpocrates:
`textRequest`/`jsonRequest` use the content-type header for more validation:

  - the expected content type is checked against requests (and specified in responses)
  - the character encoding of `textRequest` is determined from the content-type header (`jsonRequest` is always UTF-8 encoded)

Tests are added to check this is the case. In the process the following behaviour changed:

  * Akka HTTP server checks for `text/plain` content-type in `textRequest`
  * Http4s checks for `text/plain` and infers encoding from content-type in `textRequest`
  * Http4s checks for `application/json` from content-type in `jsonRequest`
  * Play checks for `application/json` from content-type in `jsonRequest`
  * Scalaj sets `text/plain` and encoding in `textRequest`

### DIFF
--- a/akka-http/server/src/test/scala/endpoints/akkahttp/server/EndpointsJsonSchemaTest.scala
+++ b/akka-http/server/src/test/scala/endpoints/akkahttp/server/EndpointsJsonSchemaTest.scala
@@ -2,6 +2,7 @@ package endpoints.akkahttp.server
 
 import akka.http.scaladsl.model.ContentTypes
 import akka.http.scaladsl.model.StatusCodes.{BadRequest, OK}
+import akka.http.scaladsl.server.UnsupportedRequestContentTypeRejection
 import akka.http.scaladsl.testkit.ScalatestRouteTest
 import endpoints.algebra.User
 import endpoints.generic
@@ -86,10 +87,23 @@ class EndpointsJsonSchemaTest
         )
       }
 
+      // Valid URL and invalid entity (3)
+      Put("/user/42").withEntity(
+        ContentTypes.`text/plain(UTF-8)`,
+        "{\"name\":\"Alice\",\"age\":true}"
+      ) ~> testRoutes.updateUser ~> check {
+        handled shouldBe false
+        rejection shouldBe UnsupportedRequestContentTypeRejection(
+          supported = Set(ContentTypes.`application/json`),
+          contentType = Some(ContentTypes.`text/plain(UTF-8)`)
+        )
+      }
+
       // Valid URL and entity
       request("/user/42", "{\"name\":\"Alice\",\"age\":55}") ~> testRoutes.updateUser ~> check {
         handled shouldBe true
         status shouldBe OK
+        contentType shouldBe ContentTypes.`application/json`
         ujson.read(responseAs[String]) shouldBe ujson.Obj(
           "name" -> ujson.Str("Alice"),
           "age" -> ujson.Num(55)
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
@@ -113,6 +113,14 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
 
   /**
     * Request with a `String` body.
+    *
+    *   - Server interpreters accept requests with content-type `text/plain` and
+    *     reject requests with an incorrect content-type.
+    *   - Server interpreters will use the character encoding set in the
+    *     content-type header to determine how the text is decoded.
+    *   - Client interpreters supply content-type `text/plain` with an explicit
+    *     character encoding
+    *
     * @group operations
     */
   def textRequest: RequestEntity[String]
--- a/algebras/algebra/src/test/scala/endpoints/algebra/client/JsonTestSuite.scala
+++ b/algebras/algebra/src/test/scala/endpoints/algebra/client/JsonTestSuite.scala
@@ -17,6 +17,7 @@ trait JsonTestSuite[T <: JsonTestApi] extends ClientTestBase[T] {
         val addressStr = """{"street":"avenue1","city":"NY"}"""
         wireMockServer.stubFor(
           post(urlEqualTo("/user"))
+            .withHeader("Content-Type", containing("application/json"))
             .withRequestBody(equalToJson(userStr))
             .willReturn(
               aResponse()
--- a/http4s/server/src/main/scala/endpoints/http4s/server/Endpoints.scala
+++ b/http4s/server/src/main/scala/endpoints/http4s/server/Endpoints.scala
@@ -12,9 +12,8 @@ import endpoints.{
   Validated,
   algebra
 }
-import fs2._
 import org.http4s
-import org.http4s.{EntityEncoder, Header, Headers}
+import org.http4s.{EntityEncoder, EntityDecoder, Header, Headers}
 
 import scala.util.control.NonFatal
 
@@ -271,13 +270,20 @@ trait EndpointsWithCustomErrors
 
   def emptyRequest: RequestEntity[Unit] = _ => Effect.pure(Right(()))
 
+  /* Setting `strict = true` means that this won't accept requests that are
+   * missing their Content-Type header. However, if we use `strict = false`,
+   * requests with incorrect specified `Content-Type` still get accepted.
+   */
   def textRequest: RequestEntity[String] =
     req =>
-      req.body
-        .through(text.utf8Decode)
-        .compile
-        .toList
-        .map(chunks => Right(chunks.mkString))
+      EntityDecoder
+        .decodeBy(http4s.MediaType.text.plain) { msg: http4s.Media[Effect] =>
+          http4s.DecodeResult.success(EntityDecoder.decodeString(msg))
+        }
+        .decode(req, strict = true)
+        .leftWiden[Throwable]
+        .rethrowT
+        .map(Right(_))
 
   def request[UrlP, BodyP, HeadersP, UrlAndBodyPTupled, Out](
       method: Method,
--- a/http4s/server/src/main/scala/endpoints/http4s/server/Endpoints.scala
+++ b/http4s/server/src/main/scala/endpoints/http4s/server/Endpoints.scala
@@ -277,7 +277,7 @@ trait EndpointsWithCustomErrors
   def textRequest: RequestEntity[String] =
     req =>
       EntityDecoder
-        .decodeBy(http4s.MediaType.text.plain) { msg: http4s.Media[Effect] =>
+        .decodeBy(http4s.MediaType.text.plain) { (msg: http4s.Media[Effect]) =>
           http4s.DecodeResult.success(EntityDecoder.decodeString(msg))
         }
         .decode(req, strict = true)
--- a/http4s/server/src/main/scala/endpoints/http4s/server/JsonEntities.scala
+++ b/http4s/server/src/main/scala/endpoints/http4s/server/JsonEntities.scala
@@ -6,7 +6,7 @@ import endpoints.{Invalid, Valid, algebra}
 import fs2.Chunk
 import org.http4s
 import org.http4s.headers.`Content-Type`
-import org.http4s.{EntityEncoder, MediaType}
+import org.http4s.{EntityEncoder, EntityDecoder, MediaType}
 
 /**
   * Interpreter for [[algebra.JsonEntitiesFromCodecs]] that decodes JSON requests
@@ -18,10 +18,19 @@ trait JsonEntitiesFromCodecs
     extends algebra.JsonEntitiesFromCodecs
     with EndpointsWithCustomErrors {
 
+  /* Setting `strict = true` means that this won't accept requests that are
+   * missing their Content-Type header. However, if we use `strict = false`,
+   * requests with incorrect specified `Content-Type` still get accepted.
+   */
   def jsonRequest[A](implicit codec: JsonCodec[A]): RequestEntity[A] =
-    req => {
+    req =>
       http4s.EntityDecoder
-        .decodeString(req)
+        .decodeBy(MediaType.application.json) { msg: http4s.Media[Effect] =>
+          http4s.DecodeResult.success(EntityDecoder.decodeString(msg))
+        }
+        .decode(req, strict = true)
+        .leftWiden[Throwable]
+        .rethrowT
         .map { value =>
           stringCodec(codec)
             .decode(value) match {
@@ -29,8 +38,6 @@ trait JsonEntitiesFromCodecs
             case inv: Invalid => Left(handleClientErrors(inv))
           }
         }
-
-    }
 
   def jsonResponse[A](implicit codec: JsonCodec[A]): ResponseEntity[A] =
     EntityEncoder[Effect, Chunk[Byte]]
--- a/http4s/server/src/main/scala/endpoints/http4s/server/JsonEntities.scala
+++ b/http4s/server/src/main/scala/endpoints/http4s/server/JsonEntities.scala
@@ -25,7 +25,7 @@ trait JsonEntitiesFromCodecs
   def jsonRequest[A](implicit codec: JsonCodec[A]): RequestEntity[A] =
     req =>
       http4s.EntityDecoder
-        .decodeBy(MediaType.application.json) { msg: http4s.Media[Effect] =>
+        .decodeBy(MediaType.application.json) { (msg: http4s.Media[Effect]) =>
           http4s.DecodeResult.success(EntityDecoder.decodeString(msg))
         }
         .decode(req, strict = true)
--- a/play/server/src/main/scala/endpoints/play/server/JsonEntities.scala
+++ b/play/server/src/main/scala/endpoints/play/server/JsonEntities.scala
@@ -4,8 +4,9 @@ import akka.util.ByteString
 import endpoints.algebra
 import endpoints.algebra.{Decoder, Encoder, Codec}
 import endpoints.Invalid
-import play.api.mvc.BodyParser
-import play.api.http.{ContentTypes, Writeable}
+import play.api.mvc.{BodyParser, RequestHeader}
+import play.api.http.{ContentTypes, Status, Writeable}
+import play.api.libs.streams.Accumulator
 
 /**
   * Interpreter for [[algebra.JsonEntitiesFromCodecs]] that decodes JSON requests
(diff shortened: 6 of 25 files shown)


harpocrates:
This is still a draft because Play's handling of JSON still doesn't check the content type. I'm struggling to see how this can be done. Play seems less extensible that the other two servers.

algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
```diff
@@ -112,6 +112,14 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
...
   /**
     * Request with a `String` body.
+    *
+    *   - Server interpreters accept requests with content-type `text/plain` and
+    *     reject requests with an incorrect content-type.
+    *   - Server interpreters will use the character encoding set in the
+    *     content-type header to determine how the text is decoded.
+    *   - Client interpreters supply content-type `text/plain` with an explicit
+    *     character encoding
+    *
```
harpocrates, line 122: This is a slightly weaker variant than https://github.com/julienrf/endpoints/pull/539#discussion_r409044736,  since otherwise I'm not sure how to implement for Play or Http4s. The key difference is that instead of specifying that server interpreters will only require content-type when there are multiple request entity types, we leave it up to particular server implementations.

@julienrf is that a reasonable position?

julienrf: Does that mean that requests with no content-type might be rejected by some interpreters?

harpocrates: Correct.

Right now, those server interpreters are Http4s and Play. Here's why:

  * Http4s [`EntityDecoder.decode`](https://github.com/http4s/http4s/blob/b8a6d9d2a5ddbf1652a78db30c3a613cec649688/core/src/main/scala/org/http4s/EntityDecoder.scala#L25) only looks at the content-type if `strict = true`, but then it also rejects requests that don't specify content type

  * Play's logic is a bit more spread out, but whenever a `BodyParser` looks at a content-type, it fails the request if the content-type is absent (see [the `text` parser](https://github.com/playframework/playframework/blob/67972edaa52057108eb1f0d108ffa1c204df4296/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L525), for instance)

We probably could re-implement the desired content-type logic all in `endpoints` (so that no content-type is still allowed). However, I see the value _endpoints_ as providing some re-usable minimal interface over existing well-tested servers/clients. Diving too deep into entity parsing would defeat that.

julienrf: > We probably could re-implement the desired content-type logic all in `endpoints` (so that no content-type is still allowed). However, I see the value _endpoints_ as providing some re-usable minimal interface over existing well-tested servers/clients. Diving too deep into entity parsing would defeat that.

I understand, but I hope this wouldn’t make our server implementations too strict :-/

I don’t know how much this can be a problem in practice, though. I just remember having been a bit annoyed by having to add a content-type because my request was otherwise rejected by a server. But this is not a big deal… I believe most HTTP clients automatically set the request content-type?

Otherwise I can surely help getting a more “tolerant” entity parsing in the Play server interpreter. For http4s I’m not sure but I can try.

harpocrates: Yeah, most HTTP clients should set content-type (ours do for sure now).

> Otherwise I can surely help getting a more “tolerant” entity parsing in the Play server interpreter. For http4s I’m not sure but I can try.

Play has [`tolerantText`](https://github.com/playframework/playframework/blob/67972edaa52057108eb1f0d108ffa1c204df4296/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L463). However, it takes the position that if you are going to check content-type at all, you should also assert that it is present (ex: `tolerantText` doesn't assert content type `text/plain`, but it doesn't guard against `application/json`).

julienrf: I think that for both http4s and Play it should be feasible (and not too complex) to implement our own `EntityDecoder` and `BodyParser` (respectively), to behave such that they reject content-types that don’t match but accept requests with no content-type set. I can give it a try if you want!

julienrf: Here is an example of implementation of `textRequest` for http4s and Play that rejects incompatible content-types but accepts requests with no Content-Type header: https://github.com/julienrf/endpoints/commit/0ac804848e8ee9795aef5621c23ebd916f279250

But I’m still not sure what we should do.

BTW, I think you discovered some bugs: what happens currently if one sends a request with an incompatible Content-Type? It gets rejected by the Play server interpreter, but not with a “good” response. The default behavior of Play is to return an HTML response.

harpocrates: > BTW, I think you discovered some bugs: what happens currently if one sends a request with an incompatible Content-Type? It gets rejected by the Play server interpreter, but not with a “good” response. The default behavior of Play is to return an HTML response.

Not sure about what the expected body type of the response type is, but I think the status 415 is correct (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415). I don't think this should be a 400 (I'm looking at the test case output that changed in the linked commit).

> Here is an example of implementation of `textRequest` for http4s and Play that rejects incompatible content-types but accepts requests with no Content-Type header: 0ac8048
>
> But I’m still not sure what we should do.

Aside from the 415 status code, these implementations generally make sense. My personal preference is still against this direction though.

As I see it, _endpoints_ interpreters should try to produce outputs that are fairly idiomatic for whatever framework/library they are targeting. Re-implementing logic in _endpoints_  makes it harder for the interpreted outputs (routes in this case) to fit neatly into the rest of the framework.

For instance, one of the cool reasons to use Play is to dynamically adjust the sort of errors that you want all routes to return (see https://www.playframework.com/documentation/2.5.x/ScalaErrorHandling). With the implementation you propose, routes created by _endpoints_ with `textRequest` would no longer obey this.

julienrf: Yes, 415 is a correct status code, I agree. The motivation for not letting Play handle errors is that an important property when using _endpoints_ is to have client, server, and documentation consistent together. If we let the underlying HTTP library handle errors, then this one is out of our control, and clients might receive error responses they can’t decode (or, which are not documented). For instance, an error response with an unknown status code, or worse, with an unexpected content-type (such HTML instead of JSON).

This is what led me to the current design that bypasses Akka-HTTP rejection handlers, and Play’s HttpErrorHandler.

harpocrates: I still don't agree with content-type check, but I understand the motivation. Thanks for explaining! 👍

julienrf: > I still don't agree with content-type check

By “content-type check”, you mean the fact that in the commit I pushed I only check that the request content-type is not different than expected?

harpocrates: Yeah sort of, although I don't disapprove of the code itself 😄.

I just disagree with the idea of trying to re-build content-type checking for servers with logic inside of _endpoints_. My opinion is we should defer to the server's own logic whenever possible.

julienrf: > I just disagree with the idea of trying to re-build content-type checking for servers with logic inside of _endpoints_. My opinion is we should defer to the server's own logic whenever possible.

I understand this point and I’m flexible to remove this custom logic. However, I still strongly believe that we should handle errors ourselves (not let the underlying library do it) so that we can guarantee that clients, servers, and documentation are consistent with the actual implementation. I’m talking here about the error that happens if a request is sent with an incompatible content-type. In such a case, we might want to return our own 415 error response (which needs to be properly documented), or we can just reuse the 400 error response (I know it’s not RESTful, but I sincerely doubt this can cause any real problem).

harpocrates: Yeah, this makes sense. I do think respecting HTTP conventions like 415 is important, especially when building servers that are going to be consumed by third parties (who may not be relying on _endpoints_).

It would indeed be better for endpoints to uniformly manage all possible errors. This is a bit tricky in practice though. Some errors, such as 404, 405, (maybe even 415) require a more global view of the set of routes defined. Other errors, such as 413, 414, 429, 431 are status codes that I generally expect any server may return (depending on their operating constraints), but I'm not sure it would be useful for _endpoints_ to track them all.

In summary: I agree with your goal, but I worry doing it right will require very deep integration with servers. Personally I am satisfied if _endpoints_ focuses on clearly specifying the correct way to call REST endpoints and providing only a best-effort description/handling of possible errors (maybe with better guarantees around status codes 400 and 500).

julienrf: > I do think respecting HTTP conventions like 415 is important, especially when building servers that are going to be consumed by third parties (who may not be relying on _endpoints_).

I’ve always thought that respecting such conventions was be useful for humans only, but I doubt programs do have a special treatment when they receive a 415 error to e.g. try with a different content type, do they?

harpocrates: In the case of 415, I agree there isn't much a program can do, but a status code can still inform how to present the error message. I can imagine a smart client getting a 413 (Request entity too large) and automatically trying to apply some tricks to retry with a compressed query. 401 is definitely useful for machines too.

julienrf: If we return a 415 status this should be included as possible response type in the documentation (like we do with statuses 500 and 400).

julienrf: Let’s move forward with this PR, though: what do you think of taking the path of requiring the `Content-Type` to be mandatory? (so that we can then implement #529)

harpocrates: Yeah, I can do that. Before I do though:

Are you sure we should? We can move forward with #529 even without mandatory `Content-Type` (ie, the PR as it is right now). I was envisioning that `Content-Type` would only become absolutely mandatory for `RequestEntities` produced by `.orElse`.

julienrf: Right! OK, let’s live with the few differences of behaviors between each interpreter then.

akka-http/server/src/test/scala/endpoints/akkahttp/server/EndpointsJsonSchemaTest.scala
```diff
@@ -86,10 +87,23 @@ class EndpointsJsonSchemaTest
         )
       }
 
+      // Valid URL and invalid entity (3)
+      Put("/user/42").withEntity(
+        ContentTypes.`text/plain(UTF-8)`,
+        "{\"name\":\"Alice\",\"age\":true}"
+      ) ~> testRoutes.updateUser ~> check {
+        handled shouldBe false
+        rejection shouldBe UnsupportedRequestContentTypeRejection(
```
julienrf, line 96: I’m not in favor of using Akka HTTP rejections, instead I would rather handle the request and return a 415 (or 400) response.

algebras/algebra/src/test/scala/endpoints/algebra/client/JsonTestSuite.scala
```diff
@@ -17,6 +17,7 @@ trait JsonTestSuite[T <: JsonTestApi] extends ClientTestBase[T] {
         val addressStr = """{"street":"avenue1","city":"NY"}"""
         wireMockServer.stubFor(
           post(urlEqualTo("/user"))
+            .withHeader("Content-Type", containing("application/json"))
```
julienrf, line 20: Does that make the test fail if the client doesn’t supply the header?

play/server/src/main/scala/endpoints/play/server/JsonEntities.scala
```diff
@@ -71,14 +72,32 @@ private object JsonEntities {
...
+      } else {
+        Accumulator.done {
+          endpoints.playComponents.httpErrorHandler
+            .onClientError(
+              request,
+              Status.UNSUPPORTED_MEDIA_TYPE,
+              "Expecting application/json body"
+            )
+            .map(Left(_))(endpoints.playComponents.executionContext)
+        }
```
julienrf, line 97: We should follow the same pattern as with bad request errors and return a response ourselves rather than forwarding to the `httpErrorHandler`.

akka-http/server/src/test/scala/endpoints/akkahttp/server/EndpointsJsonSchemaTest.scala
harpocrates: Sorry for the long period of no response! I've been pondering what a good way forward for this sort of problem is. Although I understand your concern, returning directly a 400/415 would mean that routes produced by _endpoints_ would no longer interact well with other Akka HTTP routes.

Right now, we mingle routes from `endpoint.implementBy(...)` and routes written in Akka HTTP's DSL directly (to handle routes that don't have representations in _endpoints_). If the routes implemented by _endpoints_ start manually returning 400/415, operators like `route1 ~ route2` won't work the same anymore (`route2` won't be tried if `route1` bypasses Akka HTTP's rejection mechanism). Besides, this will make it virtually impossible to ensure that our 415 responses are uniform.

algebras/algebra/src/test/scala/endpoints/algebra/client/JsonTestSuite.scala
harpocrates: Yeah, I think so.

play/server/src/main/scala/endpoints/play/server/JsonEntities.scala
harpocrates: See the comment about the Akka HTTP case. Since I don't use Play, I don't care much about the decision here, but I expect that Play users would at least care about uniformity/customizability of 415 responses (especially since customizing 415's looks like a feature - https://www.playframework.com/documentation/2.8.x/ScalaErrorHandling#Extending-the-default-error-handler looks like something I would want to be able to use).

akka-http/server/src/test/scala/endpoints/akkahttp/server/EndpointsJsonSchemaTest.scala
julienrf: I’ve described my reasoning for not using rejections in #352. I’m happy to reconsider my decision if you think this is not a viable approach.

> Right now, we mingle routes from `endpoint.implementBy(...)` and routes written in Akka HTTP's DSL directly (to handle routes that don't have representations in _endpoints_). If the routes implemented by _endpoints_ start manually returning 400/415, operators like `route1 ~ route2` won't work the same anymore (`route2` won't be tried if `route1` bypasses Akka HTTP's rejection mechanism).

The reason why routes implemented by _endpoints_ manually return 400/500 responses, is because that’s the only way I found to ensure clients, servers, and documentation are consistent together. If a route implemented by _endpoints_ returns a rejection, and then the rejection handler returns a response that _endpoints_ is unaware about, then clients would not be able to handle this response, and documentation may be inconsistent with the actual response.

> Besides, this will make it virtually impossible to ensure that our 415 responses are uniform.

I understand that this is maybe not the most desirable solution, but what would you think of reusing the responses returned by _endpoints_ to ensure that your Akka HTTP endpoints return responses that are uniform with the responses returned by _endpoints_? (this is what we do at work)

harpocrates: Thanks for the link to the issue - I'm glad to see this is something you considered! Let me try to convince you this should change:

Having the right content type is like having the right method (GET/PUT/etc.): it is part of the definition of the endpoint (documented in the OpenAPI spec, specified by clients, used by servers). The wrong content type just means the endpoint shouldn't match.

> The reason why routes implemented by endpoints manually return 400/500 responses, is because that’s the only way I found to ensure clients, servers, and documentation are consistent together.

### Consistency between documentation and client/server

I have the following expectation from the documentation interpreters:

  1. tell me how to correctly call an endpoint (aka: right URI, parameters, JSON schema, content type, etc.)
  2. list special ways in which the endpoint may fail even if I did everything in 1. This should cover:
        - additional validation errors that the JSON schema may not cover (400)
        - business logic errors (500)

I think this is the best that _endpoints_ can really hope for. Trying to cover all the ways in which a web-server can fail is going to be undesirably noisy and depends on the web-server backend; I don't expect 414 or 431 to be documented in OpenAPI, even if they can probably be reliably triggered by pathological inputs on virtually any web-server.

I view a 415 sort of like a souped up 404 - none of the endpoints matched, but the server is helpfully trying to inform the user that one endpoint came close to matching. Consequently, it has no business being documented or handled by `endpoints`.

### Consistency between client and server

 I don't think this causes any issues around consistency between client/server. The clients will specify the right content type and the servers expect the right content type. No different than what is already being done for specifying the right method.

julienrf: > I don't think this causes any issues around consistency between client/server. The clients will specify the right content type and the servers expect the right content type. No different than what is already being done for specifying the right method.

Indeed, _endpoints_ client interpreters would not be impacted. The problem is _ad-hoc_ clients implemented based on the OpenAPI documentation. If the documentation does not mention that 415 is a possible response type, and what would be the type of the entity in such a case (JSON, text, ?), an _ad-hoc_ client might find the situation surprising.

But maybe the infrastructure required to properly document such response types would be too heavy compared to its benefits…? So far, we have something that is exhaustive: all possible response types appear in the documentation (including 400 and 500 errors), and this is useful because returning 400 or 500 errors is common in practice, so we want to properly document them.

julienrf: So, here are my thoughts: let’s start with the status code 415 (and not 400, as I was maybe suggesting before), and not include such 415 responses in the documentation (like you do in this PR). If this becomes an issue (people eventually need to have these responses documented in OpenAPI) we can change later.

I’m also OK to use rejection handlers (and `HttpErrorHandler` in Play interpreter) for now, we can change later to align with the way we handle 400 and 500 responses, if needed.

harpocrates: Thanks for reporting back! Are there any outstanding changes needed for this PR then? I think everything is as you describe already:

  * status code 415 is being returned
  * 415 responses are provisionally omitted from OpenAPI documentation (and from the scaladocs too)
  * we provisionally use rejection handlers / `HttpErrorHandler` for these errors

julienrf: Indeed, I’ve had another look at the changes and I think this is good to go! Thanks for your work and your patience :)

julienrf:
Do you mind rebasing on top of master?

harpocrates:
> Do you mind rebasing on top of master?

Rebased now. Thanks again for patiently going over the design considerations involved here. 😄

http4s/server/src/main/scala/endpoints/http4s/server/JsonEntities.scala
```diff
@@ -18,10 +18,19 @@ trait JsonEntitiesFromCodecs
...
+  /* Setting `strict = true` means that this won't accept requests that are
+   * missing their Content-Type header. However, if we use `strict = false`,
+   * requests with incorrect specified `Content-Type` still get accepted.
+   */
   def jsonRequest[A](implicit codec: JsonCodec[A]): RequestEntity[A] =
-    req => {
+    req =>
       http4s.EntityDecoder
-        .decodeString(req)
+        .decodeBy(MediaType.application.json) { msg: http4s.Media[Effect] =>
```
julienrf, line 28: Dotty requires that we put parens here:

```suggestion
        .decodeBy(MediaType.application.json) { (msg: http4s.Media[Effect]) =>
```

http4s/server/src/main/scala/endpoints/http4s/server/Endpoints.scala
```diff
@@ -271,13 +270,20 @@ trait EndpointsWithCustomErrors
...
+   */
   def textRequest: RequestEntity[String] =
     req =>
-      req.body
-        .through(text.utf8Decode)
-        .compile
-        .toList
-        .map(chunks => Right(chunks.mkString))
+      EntityDecoder
+        .decodeBy(http4s.MediaType.text.plain) { msg: http4s.Media[Effect] =>
```
julienrf, line 280: And here:


```suggestion
        .decodeBy(http4s.MediaType.text.plain) { (msg: http4s.Media[Effect]) =>
```

Issue #531, opened 2020-04-08
Extend OpenApi model


bmeesters:
The official OpenApi model offers some operations which are currently not supported by Endpoints generation. For example you can create a top level description whereas the current definition only supports a version and a title:

```scala
case class Info(
    title: String,
    version: String
    // TODO: add description field as well
)
```

Another, more involved change would be to add support for information on tags, see https://swagger.io/docs/specification/grouping-operations-with-tags/. Not only are you allowed to tag operations, you can also add additional information to these tags like you would to components. In a more complex domain, this would allow you to describe your data types in semantics, not only in structure.

bmeesters:
@julienrf I added this issue since we need something like this in our documentation. The first is probably very simple. The latter would require to reshape tags, as they are currently only strings.

I would be willing to work on this, though it might be good to sketch a solution before coding. WDYT?

julienrf:
I agree that we should support as many things that are supported by OpenAPI. As you said, the current model provides only partial support.

One important point (since I would like to release a 1.x version soon) is to either make the OpenAPI model complete soon or to redesign it in a way that supports adding more stuff to it in a backward-compatible way. This means using plain classes instead of case classes (we can see an example of a design that supports backward-compatible evolutions in [sbt-contraband](https://www.scala-sbt.org/contraband/code-generation.html#Record+types), for instance).

About tags, the idea would to use [this model](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#tagObject) instead of just strings or is there anything else that I missed?

To ease migration, it would be better if we could still provide the existing API for adding tags and constructing the `Info`. ie, code like this should still compile:

~~~ scala
val someResource =
  endpoint(
    get(path / "some-resource"),
    ok(textResponse),
    docs = EndpointDocs(description = Some("The contents of some resource"), tags = List("foo"))
  )

val docs = openApi(Info("Title", "0.0.0"))(someResource)
~~~

Does that make sense?

I know that redesigning the OpenAPI model would require a significant effort comparing to just adding a  `description` field, but would you be interested in doing that?

bmeesters:
Yes I can understand your concerns. IMO it makes sense to change de OpenApi model so that it remains backwards compatible once new things are added. Have you thought about also doing something with abstract types and the OpenApi model? Not sure if that is feasible, but it is one of the strengths of _endpoints_ that you can also update your data model.

[...]

PR #546, opened 2020-04-21
Rewrite open api case classes to normal classes for backwards compatibility

bmeesters:
Partially solves #531

### DIFF
--- a/json-schema/json-schema/src/test/scala/endpoints/HashingTest.scala
+++ b/json-schema/json-schema/src/test/scala/endpoints/HashingTest.scala
@@ -9,7 +9,7 @@ object HashingTest extends Properties("Hashing") {
     val oneValue =
       (x == y) == (Hashing.hash(x) == Hashing.hash(y))
     val twoValues =
-      ((x, s) == (y, t)) == (Hashing.hash(x, s) == Hashing.hash(y, t))
+      ((x, s) == ((y, t))) == (Hashing.hash(x, s) == Hashing.hash(y, t))
     oneValue && twoValues
   }
 
--- a/json-schema/json-schema/src/test/scala/endpoints/HashingTest.scala
+++ b/json-schema/json-schema/src/test/scala/endpoints/HashingTest.scala
@@ -1,0 +1,16 @@
+package endpoints
+
+import org.scalacheck.Properties
+import org.scalacheck.Prop.forAll
+
+object HashingTest extends Properties("Hashing") {
+
+  property("hash") = forAll { (x: Int, y: Int, s: String, t: String) =>
+    val oneValue =
+      (x == y) == (Hashing.hash(x) == Hashing.hash(y))
+    val twoValues =
+      ((x, s) == (y, t)) == (Hashing.hash(x, s) == Hashing.hash(y, t))
+    oneValue && twoValues
+  }
+
+}
--- a/openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
@@ -525,24 +525,134 @@ object Operation {
 
 }
 
-case class SecurityRequirement(
-    name: String,
-    scheme: SecurityScheme,
-    scopes: List[String] = Nil
-)
-
-case class RequestBody(
-    description: Option[String],
-    content: Map[String, MediaType]
+class SecurityRequirement(
+    val name: String,
+    val scheme: SecurityScheme,
+    val scopes: List[String] = Nil
 ) {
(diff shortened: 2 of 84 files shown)


bmeesters:
Hello @julienrf I have started on id:531 as discussed. I think this is all that has to happen right (but then for every class)? I suppose all case classes in the `OpenApi.scala` file should be done? Just checking, before I made the same mistake over and over for every domain object.

julienrf:
Thanks @bmeesters for starting the discussion :)

>  I think this is all that has to happen

We should also override `hashCode` and `equals`.

> I suppose all case classes in the OpenApi.scala file should be done?

Ideally, yes… We should be able to support any minor evolution of the OpenAPI schema.

There is also the class `EndpointDocs`. If that’s too much work we can split it, let me know what you want to do.

bmeesters:
No problem :slightly_smiling_face: 

How should the `hashCode` look like? `equals` I can imagine. I will probably start with all case classes in the OpenApi file. If it is just these two extra classes I will also do those. Though any more I would like to separate them. Its not complex work, but it is a bit tedious.

julienrf:
We can just follow the pattern used by sbt-contraband, I believe: https://www.scala-sbt.org/contraband/code-generation.html#Record+types

bmeesters:
> We can just follow the pattern used by sbt-contraband, I believe: https://www.scala-sbt.org/contraband/code-generation.html#Record+types

Including the numbers (37, 37, 17) or just adding the hashcodes of the individual fields? I don't really understand how this came to be, which is why I am asking.

julienrf:
Yes, the factors are important (and the parenthesis as well). You can find more information here: https://stackoverflow.com/questions/7370925/what-is-the-standard-idiom-for-implementing-equals-and-hashcode-in-scala

In case you use IntelliJ, it can generate those methods for you.

harpocrates:
Why not just use `MurmurHash3` directly? I think this is anyways what case classes do nowadays, right?

julienrf:
[Indeed](https://github.com/scala/scala/blob/51233da2bc6fb9e4315c8dc9620f213a5e13c888/src/library/scala/runtime/ScalaRunTime.scala#L161), I didn’t know. Yes, we should use that too!

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -269,65 +304,467 @@ object OpenApi {
...
+}
+
+object Info {
+
+  def apply(title: String, version: String): Info =
+    new Info(title, version)
+
+}
+
+class PathItem(
```
julienrf, line 345: Should we make the constructor `private`?

bmeesters: Yes, I think that makes sense.

bmeesters:
Hey @julienrf, it took a while, but I think this is ready for review now. I didn't do any of the changes regarding `Tag` since that would require logic changes and this is only rewrites of existing classes. I plan to start on that once this is merged. Can you have a look?

julienrf:
Thanks a lot @bmeesters, I’ll review it today.

json-schema/json-schema/src/test/scala/endpoints/HashingTest.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.scalacheck.Properties
+import org.scalacheck.Prop.forAll
+
+object HashingTest extends Properties("Hashing") {
+
+  property("hash") = forAll { (x: Int, y: Int, s: String, t: String) =>
+    val oneValue =
+      (x == y) == (Hashing.hash(x) == Hashing.hash(y))
+    val twoValues =
+      ((x, s) == (y, t)) == (Hashing.hash(x, s) == Hashing.hash(y, t))
```
julienrf, line 12: It seems that this syntax doesn’t work with Scala 2.

```suggestion
      ((x, s) == ((y, t))) == (Hashing.hash(x, s) == Hashing.hash(y, t))
```

bmeesters:
> I just have one open question. For types EndpointDocs and CallbackDocs, I’ve defined a apply methods that only take required parameters, and optional parameters (or those for which we have a sensible default) can be set by using withXxx methods. Maybe it would be more regular to follow this scheme everywhere… what do you think? The drawback is that this might create incompatibilities with the current API (see the @deprecated methods I’ve added).

I think that makes sense. The breakage is so minor that I don't expect anyone to have major troubles moving to the new scheme. I'll do one more commit to add that as well.

julienrf:
There is one last thing: we should make the classes `final` as well.

bmeesters:
Right, forgot about that. Done now.

PR #551, opened 2020-04-26
Remove some warnings on Dotty

julienrf:
- Remove occurrences of unreducible application of higher-kinded types
- Rename `enum` identifier to `enumeration`

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Endpoints.scala
@@ -80,11 +80,50 @@ trait EndpointsWithCustomErrors extends Requests with Responses with Errors {
     * @param entity   Contents of the callback message
     * @param response Expected response
     */
-  case class CallbackDocs(
+  case class CallbackDocs private (
       method: Method,
-      entity: RequestEntity[_],
-      response: Response[_],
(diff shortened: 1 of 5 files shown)


julienrf:
@bmeesters, This might overlap or clash with the work you are currently doing on `CallbackDocs`. I’m sorry about it if that’s the case! If you prefer, I can take care of migrating these two data types.

bmeesters:
Hello @julienrf, thanks for the warning! I have to admit I did not make much progress yet, so you can safely make some changes without overlap. I think I won't have time to continue until next week. Some unexpected things came up that have priority now.

julienrf:
Allright, thanks for your response!

julienrf:
Don’t be sorry, I am very grateful to you for the time you took to review my PR!

> I spent some time trying to actually get this to cross-build with Dotty. Is the latter possible yet?

It indeed requires a couple of changes in the build definition (essentially, adding `.withDottyCompat` to our Scala 2.13 library dependencies). Thankfully, a [major improvement](https://github.com/lampepfl/dotty/pull/8824) has been merged in Dotty a few days ago, so we should be able to compile our main artifacts without many changes. The two remaining “problems” are the `generic` module, which depends on shapeless, which is not available in Scala 3, and the tests, which depend on Scalatest, which does not work with all Dotty versions…

PR #556, opened 2020-05-01
Clear some warnings and remove unused code

julienrf:


julienrf:
There is a failure that happens often on the CI but that’s difficult to reproduce.

https://travis-ci.com/github/julienrf/endpoints/jobs/326290812#L1475

It’s always at the same place: http4s-client test suite, same test.

julienrf:
I’ve pushed a workaround in 62f71d4 but it would be nice to have a proper solution. Any cats-effect or http4s expert knows what’s wrong in our code?

PR #577, opened 2020-05-27
Extend Info with description in OpenApi model

bmeesters:


openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -313,39 +322,45 @@ object OpenApi {
...
     other match {
-      case that: Info => title == that.title && version == that.version
-      case _          => false
+      case that: Info =>
+        title == that.title && version == that.version && description == that.description
+      case _ => false
     }
 
   override def hashCode(): Int =
     Hashing.hash(title, version)
```
julienrf, line 340: You should add `description` here as well.

bmeesters: Missed, I'll update.

bmeesters:
Hello @julienrf, I was also wanted to include the tags here, but maybe that is better as a separate PR. WDYT?

julienrf:
Yes, I prefer smaller PRs :) It’s easier to review, thanks!

bmeesters:
Alright, I think this is ready for review. I'll do another PR later for adding support for richer tags.

julienrf:
Thank you!

Issue #574, opened 2020-05-25
Akka HTTP client doesn't properly encode segments with spaces


mleclercq:
If a URL segment contains a space, it is encoded as `+` instead of `%20` (`+` is OK for query-string, but in the path of the URL, `%20` should be used).

I did the following, using Akka `Uri` to workaround the issue:

```scala
    override implicit lazy val stringSegment: Segment[String] = (s: String) =>
      Uri.Path.Segment(s, Uri.Path.Empty).toString()
```

julienrf:
Good catch! Would you be interested in contributing a fix? You need to fix the Akka-HTTP interpreter and also add tests [here](https://github.com/julienrf/endpoints/blob/e52dbb17da69ac5488201b860382abfc1369eb24/algebras/algebra/src/test/scala/endpoints/algebra/client/EndpointsTestSuite.scala#L318-L333)

PR #579, opened 2020-05-30
Fix encoding of Urls Segment

mleclercq:
Addresses #574

For akka and http4s, I used existing encoder from these frameworks to encode `Segment`. For the others, I re-implemented the encoding inspired by the http4s implementation.


julienrf:
Thank you!

PR #581, opened 2020-06-03
Add Tag to OpenAPI model

bmeesters:
Resolves https://github.com/julienrf/endpoints/issues/531

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
@@ -2,11 +2,14 @@ package endpoints.algebra
 
 import endpoints.Hashing
 
-class Tag private (
+/**
+  * @see [[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#tagObject]]
+  */
+final class Tag private (
     val name: String,
     val description: Option[String],
     val externalDocs: Option[ExternalDocumentationObject]
-) {
+) extends Serializable {
 
   override def toString =
     s"Tag($name, $description, $externalDocs)"
@@ -40,16 +43,21 @@ class Tag private (
 
 object Tag {
 
-  def apply(
-      name: String
-  ) = new Tag(name, None, None)
+  /**
+    * Creates a new Tag without a description and external documentation object.  Use
+    * Tag(..).withDescription(..).withExternalDocs(..) to add those attributes.
+    */
+  def apply(name: String) = new Tag(name, None, None)
 
 }
 
-class ExternalDocumentationObject private (
+/**
+  * @see [[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#externalDocumentationObject]]
+  */
+final class ExternalDocumentationObject private (
     val url: String,
     val description: Option[String]
-) {
+) extends Serializable {
 
   override def toString: String =
     s"ExternalDocumentationObject($url, $description)"
@@ -75,7 +83,9 @@ class ExternalDocumentationObject private (
 
 object ExternalDocumentationObject {
 
-  def apply(
-      url: String
-  ) = new ExternalDocumentationObject(url, None)
+  /**
+    * This creates an ExternalDocumentationObject without a description. Use
+    * ExternalDocumentationObject(..).withDescription(..) to add a description.
+    */
+  def apply(url: String) = new ExternalDocumentationObject(url, None)
 }
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
@@ -1,0 +1,81 @@
+package endpoints.algebra
+
+import endpoints.Hashing
+
+class Tag private (
+    val name: String,
+    val description: Option[String],
+    val externalDocs: Option[ExternalDocumentationObject]
+) {
+
+  override def toString =
+    s"Tag($name, $description, $externalDocs)"
+
+  override def equals(other: Any): Boolean =
+    other match {
+      case that: Tag =>
+        name == that.name && description == that.description && externalDocs == that.externalDocs
+      case _ => false
+    }
+
+  override def hashCode(): Int =
+    Hashing.hash(name, description, externalDocs)
+
+  private[this] def copy(
+      name: String = name,
+      description: Option[String] = description,
+      externalDocs: Option[ExternalDocumentationObject] = externalDocs
+  ): Tag = new Tag(name, description, externalDocs)
+
+  def withName(name: String): Tag =
+    copy(name = name)
+
+  def withDescription(description: Option[String]): Tag =
+    copy(description = description)
+
+  def withExternalDocs(externalDocs: Option[ExternalDocumentationObject]): Tag =
+    copy(externalDocs = externalDocs)
+
+}
+
+object Tag {
+
+  def apply(
+      name: String
+  ) = new Tag(name, None, None)
+
+}
+
+class ExternalDocumentationObject private (
+    val url: String,
+    val description: Option[String]
+) {
+
+  override def toString: String =
+    s"ExternalDocumentationObject($url, $description)"
+
+  override def equals(other: Any): Boolean =
+    other match {
+      case that: ExternalDocumentationObject =>
+        url == that.url && description == that.description
+      case _ => false
+    }
+
+  override def hashCode(): Int =
+    Hashing.hash(url, description)
+
+  def withUrl(url: String): ExternalDocumentationObject =
+    new ExternalDocumentationObject(url, description)
+
+  def withDescription(
+      description: Option[String]
+  ): ExternalDocumentationObject =
+    new ExternalDocumentationObject(url, description)
+}
+
+object ExternalDocumentationObject {
+
+  def apply(
+      url: String
+  ) = new ExternalDocumentationObject(url, None)
+}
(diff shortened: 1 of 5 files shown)


bmeesters:
Heey @julienrf, I did the minimum effort to support richer tags. There are no unit tests for the extended model yet (just made the current ones pass) & I do not check on whether there are tags with the same name but different content.

Before I get started on that I wanted to check the location of `Tag.scala`. It doesn't feel completely right to put it in algebra (since it is part of OpenAPI), but otherwise you cannot create tags from an algebra. Any thoughts?

julienrf:
Thanks @bmeesters!

> I wanted to check the location of `Tag.scala`

As you said, it has to be in the algebra. It’s like `EndpointDocs` or `CallbackDocs`. These classes carry information used by the OpenAPI interpreter only, but we want to introduce them at the algebra level.

bmeesters:
Alright, thanks for getting back to me so soon! I'll continue as is then.

bmeesters:
@julienrf I think this is good to go. Some decisions that might need a little attention:

- The deprecated way of creating EndpointDocs will automatically convert strings to tags.
- Tags only appear in the tags list whenever they have more than a name. In combination with the above this means that the exact same open-api will be generated whenever users upgrade to this version.
- The OpenAPI class _throws_ an exception on creation if tags are inconsistent. Technically the tags are only necessary when generating open-api JSON but I think it is better to fail right away. Also I generally prefer typed errors, but since this indicates a programming error I don't think that has much value now.
The tag extraction is probably not very fast with multiple collections involved. Though I think the code is simple and it shouldn't matter to much since the low amount of tags.

I tested the fixture JSON against the swagger UI and it looks as expected.

WDYT?

julienrf:
Hey @bmeesters I’ll have a look at this one this evening. There seems to be a compiler crash when the scaladoc is generated… See https://travis-ci.com/github/julienrf/endpoints/jobs/343077672#L1386
Would you have time to have a look?

algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
```diff
@@ -0,0 +1,81 @@
+package endpoints.algebra
+
+import endpoints.Hashing
+
+class Tag private (
+    val name: String,
+    val description: Option[String],
+    val externalDocs: Option[ExternalDocumentationObject]
+) {
```
julienrf, line 9: Maybe also extend `Serializable`?

algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
```diff
@@ -0,0 +1,81 @@
+package endpoints.algebra
+
+import endpoints.Hashing
+
+class Tag private (
```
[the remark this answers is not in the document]
julienrf: Also, could you make the class `final`?

algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
```diff
@@ -0,0 +1,81 @@
...
+    copy(description = description)
+
+  def withExternalDocs(externalDocs: Option[ExternalDocumentationObject]): Tag =
+    copy(externalDocs = externalDocs)
+
+}
+
+object Tag {
+
+  def apply(
```
julienrf, line 43: Could you add a scaladoc comment explaining that this builds a tag with the given name, no description and no external documentation object?

algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
```diff
@@ -0,0 +1,81 @@
...
+      name: String
+  ) = new Tag(name, None, None)
+
+}
+
+class ExternalDocumentationObject private (
```
julienrf, line 49: Same here, could you make it `final`, extend `Serializable`, and add a Scaladoc description?

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -3,12 +3,13 @@ package endpoints.openapi.model
...
 
 import endpoints.Hashing
-import endpoints.algebra.Encoder
+import endpoints.algebra.{Encoder, ExternalDocumentationObject, Tag}
 
 import scala.collection.mutable
 
 /**
   * @see [[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md]]
+  * @throws java.lang.IllegalArgumentException on creation if the tags are inconsistent with each other.
```
julienrf, line 12:
```suggestion
  * @throws java.lang.IllegalArgumentException on creation if several tags have the same name but not the same other attributes.
```

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -306,14 +339,34 @@ object OpenApi {
...
+    }.flatten
+
+    val tagsByName = allTags.groupBy(_.name)
+    tagsByName.foreach { case (_, listOfTags) =>
+      val set = listOfTags.toSet
+      if (set.size > 1) {
+        throw new IllegalArgumentException(s"Found tags with the same name but different values: $set")
+      }
+    }
+    allTags.filter(tag => tag.description.nonEmpty || tag.externalDocs.nonEmpty).toSet.toList
```
julienrf, line 367: Why do we keep only those whose description or external doc is defined?

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -306,14 +339,34 @@ object OpenApi {
...
+      if (openApi.tags.nonEmpty) {
+       fields += "tags" -> ujson.Arr(openApi.tags.map(tagJson): _*)
+      }
       if (openApi.components.schemas.nonEmpty || openApi.components.securitySchemes.nonEmpty) {
         fields += "components" -> componentsJson(openApi.components)
       }
       new ujson.Obj(fields)
     }
 
+  private def extractTags(paths: Map[String, PathItem]): List[Tag] = {
```
julienrf, line 353: Should we return a `Set[Tag]`?

algebras/algebra/src/main/scala/endpoints/algebra/Tag.scala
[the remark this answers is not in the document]
bmeesters: Yes, sorry, was again a bit sloppy with the conversion to a normal class. I'll update it.

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
bmeesters: Yes we could do that, I made it a list since that is what ujson expects, that conversion can also happen later.

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
bmeesters: The tags at the root are used only when there is more information in the tag than just the name. So, if a tag is just a name there is no value in adding it in the root anyway. It will still be displayed as before.

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
bmeesters: I expect that the compiler crash has something to do with the reference of the exception, though the error itself is quite unreadable. I try by restructuring this documentation a bit.

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
julienrf: OK, thanks for the clarification! Do we have a test for that? It seems that in `ReferencedSchemaTest` we only use a “rich” tag?

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
```diff
@@ -308,12 +344,45 @@ object OpenApi {
           "info" -> infoJson(openApi.info),
           "paths" -> pathsJson(openApi.paths)
         )
+      if (openApi.tags.nonEmpty) {
+        fields += "tags" -> new ujson.Arr(
+          openApi.tags.map(tag => tagJson(tag)).to(mutable.ArrayBuffer)
```
julienrf, line 349: This doesn’t compile on 2.12. Can you see if importing `scala.collection.compat._` fixes the issue?

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
bmeesters: Sure, I'll add another case t the `ReferencedSchemaTest`.

openapi/openapi/src/main/scala/endpoints/openapi/model/OpenApi.scala
bmeesters: When I add that it seems it is not used. I could also go back to the constructor as before (`tags: _*`)

julienrf: Yeah, that’s also a solution :)

It’s unused only on 2.13, though, so you should be able to remove the warning with a `@nowarn` annotation (I believe). But your alternative solution might be simpler :)

bmeesters:
No problem, happy to help! :slightly_smiling_face:

Btw, any chance there will be a release including these changes soon?

julienrf:
There is still one issue that I would like to see addressed before cutting a release: https://github.com/julienrf/endpoints/milestone/5

bmeesters:
Alright, makes sense. Does the current build publish anything (a snapshot version) publicly, or is that only done with releases? Would be nice if we would somehow already use this.

julienrf:
I’ve just published a snapshot version `0.15.0+90-108c999c` so that you can test it.

bmeesters:
Thanks @julienrf, much appreciated!

PR #582, opened 2020-06-04
chore: bump scala.js to 1.1.0

ngbinh:
note that I have to commented out examples-cqrs projects as it depends
on `monadic-html`. I can try to release a scala.js 1.1.0 version for
`monadic-html` but not sure if it worths it.


julienrf:
> it depends on `monadic-html`.

Would it be possible to replace it with something else?

.travis.yml
```diff
@@ -36,9 +36,6 @@ jdk:
 jobs:
   include:
 
-    - stage: test
-      scala: 2.12.11
-      script: sbt ++$TRAVIS_SCALA_VERSION test
```
julienrf, line 41: This is 2.12, not 2.11 :)

ngbinh: Wow,  will fix.

ngbinh:
> Would it be possible to replace it with something else?

Yes. I would like to use `https://github.com/raquo/Laminar`, a pure Scala UI library

julienrf:
It seems [org.scoverage:scalac-scoverage-runtime_sjs1.1_2.12:1.4.1](https://travis-ci.com/github/julienrf/endpoints/jobs/343486676#L1460) is not available?

ngbinh:
ok, I will work on fixing the build and Laminar.

ngbinh:
pushed a version with `Laminar` instead of `monadic-html`. Also, fixing issue with `scoverage-runtime` and sjs 1.X

documentation/examples/cqrs/infra/src/main/resources/application.conf
```diff
@@ -1 +1,2 @@
-play.crypto.secret=abcabcabcabcabcabc
\ No newline at end of file
+play.crypto.secret=abcabcabcabcabcabc
+play.http.secret.key=11111111000000
```
julienrf, line 2: I believe `play.crypto.secret` is not needed anymore?

```suggestion
play.http.secret.key=11111111000000
```

ngbinh: removed all of the references on `application,conf`

ngbinh:
Hmm, I have little idea on why Travis builds fail.

documentation/examples/cqrs/web-client/src/main/scala/cqrs/webclient/Main.scala
```diff
@@ -51,7 +48,7 @@ object Main {
       }
     }
 
-    def onAddValueClicked(meter: Meter): Event => Unit = _ => {
+    def onAddValueClicked: Observer[Meter] = Observer { meter =>
```
raquo, line 51: This can be just `def onAddValueClicked(meter: Meter): Unit = {` (usage would be exactly the same), or if you want to keep it as Observer it can be a `val` instead of `def`

documentation/examples/cqrs/web-client/src/main/scala/cqrs/webclient/Main.scala
```diff
@@ -74,53 +71,61 @@ object Main {
...
-                  } />
-                        <button onclick={onAddValueClicked(meter)}>Add</button>
-                      </p>
-                    </section>
-              }
+    val app = article(
+      h1("Meters"),
+      p(
+        input(
+          `type` := "text",
```
raquo, line 78: If you wanted to avoid backticks there are `typ` and `tpe` aliases for `type`.

documentation/examples/cqrs/web-client/src/main/scala/cqrs/webclient/Main.scala
```diff
@@ -74,53 +71,61 @@ object Main {
...
+          placeholder := "New meter name",
+          idAttr := newMeterId,
+          required := true
+        ),
+        button(onClick.map(_ => ()) --> onNewMeterClicked, "Create")
```
raquo, line 83: You can shorten `map(_ => ())` to `mapTo(())`

documentation/examples/cqrs/web-client/src/main/scala/cqrs/webclient/Main.scala
```diff
@@ -74,53 +71,61 @@ object Main {
...
+      ),
+      child <-- metersVar.signal.map { meters =>
+        if (meters.isEmpty) {
+          p("No meters yet!")
+        } else {
+          div(
+            meters.toSeq.map {
```
raquo, line 90: Keep in mind that metersVar contains a Map, which is by its nature unsorted, so it will be rendered in arbitrary order determined by toSeq. You could make metersVar contain a List[Meter] instead, or define some sorting order if you wanted to address that.

ngbinh:
Thanks @raquo, will address your comments soon

ngbinh:
@julienrf do you have any idea on when 1.0 will land? No pressure here

julienrf:
@ngbinh I’ll try to make it before the end of the month.

ngbinh:
I will try to resolve the conflicts soon

julienrf:
I think merging `master` should just work, except maybe for `EndpointsSettings.scala`.

julienrf:
I’ve tagged `v1.0.0` so the next step is to merge your PR :)

ngbinh:
rebased...

project/EndpointsSettings.scala
```diff
@@ -99,7 +98,7 @@ object EndpointsSettings {
   // --- Common dependencies
 
   val circeVersion = "0.13.0"
-  val playjsonVersion = "2.8.1"
+  val playjsonVersion = "2.9.0"
```
julienrf, line 101: Is it compatible with Play 2.8.2?

ngbinh: I think it is, also only `2.9.0` has published artifacts for scala.js 1.1.0 https://mvnrepository.com/artifact/com.typesafe.play/play-json

julienrf:
Thank you @ngbinh !

julienrf:
Artifacts for Scala.js are on their way to Maven Central.

PR #591, opened 2020-06-22
Add `orElse` for `RequestEntity` based on content-type

harpocrates:
Add a `choiceRequestEntity` combinator and `orElse` implicit ops method
for combining two request entities and branching based on content-type.
OpenAPI supports describing this sort of endpoint too.

Motivating use case: it is now possible to define an endpoint that
accepts either a plain `textRequest` or a richer `jsonRequest`.

Implements #529

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
@@ -125,6 +125,29 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
     */
   def textRequest: RequestEntity[String]
 
+  def choiceRequestEntity[A, B](
+    requestEntityA: RequestEntity[A],
+    requestEntityB: RequestEntity[B]
+  ): RequestEntity[Either[A, B]]
+
+  implicit class RequestEntitySyntax[A](requestEntity: RequestEntity[A]) {
+
+    /**
+      * If the entity does not seem enc/decodable as an [[A]], try decoding as a [[B]] using otherRequestEntity
+      * @note If the Entity may be enc/decodable as an [[A]], it will be. This is particularly important when
+      *       [[requestEntity]] is akka-server's [[textRequest]], which will match and decode any text-serializable
+      *       entity, regardless of Content-Type
+      * @note If [[A]] and [[B]] are both JSON-encoded and use disjoint schemas, use
+      *       [[endpoints.algebra.JsonSchemas.TaggedOps#orElse]] at the schema level instead
+      * @see [[SumTypedRequestEntity#choiceRequestEntity]]
+      * @param otherRequestEntity A [[RequestEntity]] to use to enc/decode a [[B]]
+      * @tparam B what to attempt as an enc/decoding target if not an [[A]]
+      * @return a [[RequestEntity]][Either[A, B]]
+      */
+    def orElse[B](otherRequestEntity: RequestEntity[B]): RequestEntity[Either[A, B]] =
+      choiceRequestEntity(requestEntity, otherRequestEntity)
+  }
+
   /**
     * Request for given parameters
     *
--- a/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
+++ b/algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
@@ -95,7 +95,8 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
     *     in a JSON array. Refer to the documentation of your server interpreter
     *     to customize this behavior.
     *
-    * @note  This type has implicit methods provided by the [[PartialInvariantFunctorSyntax]] class.
+    * @note This type has implicit methods provided by the [[PartialInvariantFunctorSyntax]] and
+    *       [[RequestEntitySyntax]] classes.
     * @group types */
   type RequestEntity[A]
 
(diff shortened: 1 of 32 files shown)


algebras/algebra/src/main/scala/endpoints/algebra/Requests.scala
```diff
@@ -125,6 +125,41 @@ trait Requests extends Urls with Methods with SemigroupalSyntax {
...
+    * Server interpreters accept either of the request entities
+    * Client interpreters provide one of the two request entities
+    * Documentation interpreters list all possible content types and their entities
+    */
+  def choiceRequestEntity[A, B](
+      requestEntityA: RequestEntity[A],
+      requestEntityB: RequestEntity[B]
+  ): RequestEntity[Either[A, B]]
+
+  implicit class RequestEntitySyntax[A](requestEntity: RequestEntity[A]) {
```
julienrf, line 144: Could you also add that the type `RequestEntity` has implicit methods provided by the `RequestEntitySyntax` class, on line 98?

algebras/algebra/src/test/scala/endpoints/algebra/server/SumTypedEntitiesTestSuite.scala
```diff
@@ -0,0 +1,69 @@
...
+              response.entity.contentType == ContentTypes.`application/json`
+            )
+            ujson.read(entity) shouldEqual ujson.Obj(
+              "name" -> ujson.Str("Alice"),
+              "age" -> ujson.Num(42)
+            )
+        }
+        ()
+      }
+    }
```
julienrf, line 50: Do you mind adding a test with a request with a content-type `application/json` but with an invalid JSON entity (e.g. omit the `age` field), and check that the error is 400, not 415?

harpocrates: Good idea! That is an important thing to test.

LeifW:
Is there something for using the content-type / accepts header to determine the deserialization / serialization of a given domain object - something that doesn't involve `Either`s? The more normal case w/ REST / HTTP is that a given URL represents a given domain object, and the e.g. Content-Type headers represent different serialization options (e.g. JSON, XML, YAML, HTML, etc) for the object...
E.g. in the Haskell equivalent Servant, you specify the type an endpoint corresponds to, and a list of allowed content types for interacting with that type: https://hackage.haskell.org/package/servant-0.19/docs/Servant-API-ContentTypes.html

julienrf:
Hello,

There are no utilities in endpoints4s that help with content negotiation, so far. Content negotiation is different from the usual communication scheme that happens in endpoints4s, where both the client and the server use the very same protocol (hence, no need to support multiple representations of a resource).

That being said, we could add content negotiation to endpoints4s (I think that would be an interesting addition). You can even work on it outside of endpoints4s itself, and we can consider merging it to the core library later.

I can guide you further if needed, but in summary, you need to create a new trait that extends `EndpointsWtihCustomErrors` to extend the algebra:

~~~ scala
trait ContentNegotiationAlgebra extends endpoints4s.algebra.EndpointsWithCustomErrors {

  // response status will be OK if the client accepts a content-type supported by the server,
  // or Not Acceptable otherwise
  def negotiated[A](supportedContentTypes: List[ResponseEntity[A]], documentation: String): Response[A]
  
}
~~~

Then, you would write a documentation interpreter:

~~~ scala
trait ContentNegotiationDocumentation extends ContentNegotiationAlgebra
    with endpoints4s.openapi.EndpointsWithCustomErrors {

  def negotiated[A](supportedContentTypes: List[ResponseEntity[A]], documentation: String): Response[A] =
    List(
      DocumentedResponse(status = OK, documentation, emptyHeaders, acceptedContentTypes.flatten.toMap),
      DocumentedResponse(status = 406, "Unsupported content-type requested by the client", emptyHeaders, Map.empty)
    )

}
~~~

And a server interpreter:

~~~ scala
trait ContentNegotiationServer extends ContentNegotiationAlgebra
    with endpoints4s.http4s.server.EndpointsWithCustomErrors {

  def negotiated[A](acceptedContentTypes: List[ResponseEntity[A]], documentation: String): Response[A] =
    ??? // TODO compare the accepted content types with what you support and return the appropriate response

}
~~~

While I am writing this, I realize that it might currently be impossible to implement the http4s interpreter (because when you build a response you don’t have access to the underlying request anymore, which is necessary to inspect the `Accept` header). So, this might require a change in the http4s interpreter. I think it would be possible to implement an akka-http interpreter, though.

That being said, to do something similar on the “request” side (as mentioned in your comment), I think you can do it already now with very little “boilerplate”:

~~~ scala
val entity1: RequestEntity[A] = … // e.g., JSON request entity
val entity2: RequestEntity[A] = ... // e.g., YAML request entity for the same domain object
val entities =
  entity1.orElse(entity2)
    .xmap {
      // on the server side, “forget” about the serialization option (the “Left” or “Right” information)
      case Left(a) => a
      case Right(a) => a
    } {
      // on the client side, pick an arbitrary serialization option
      a => Right(a)
    }
~~~

This could even be captured in a reusable utility method:

~~~ scala
trait MergedRequestEntities extends endpoints4s.algebra.Requests {
  def merged[A](entity: RequestEntity[Either[A, A]]): RequestEntity[A] =
    entity.xmap {
      // on the server side, “forget” about the serialization option (the “Left” or “Right” information)
      case Left(a) => a
      case Right(a) => a
    } {
      // on the client side, pick an arbitrary serialization option
      a => Right(a)
    }
}
~~~

LeifW:
Is there something like this for response entities, too (based on looking at the Accept header of the request)?
Found this tapir issue for adding that: https://github.com/softwaremill/tapir/issues/248
Was picturing perhaps e.g.
```scala
 def jsonOrYamlResponse[A: JsonResponse]: ResponseEntity[A]
 def jsonOrYamlRequest[A: JsonRequest]: RequestEntity[A]
 ```
 The latter can be implemented with this current `orElse`, but I don't know how to do the former.

PR #595, opened 2020-06-26
Rename endpoints to endpoints4s

### LINKED ISSUE #593: New organization name
As suggested on [gitter](https://gitter.im/julienrf/endpoints?at=5eb98a65fb95b64428d4bbc5), it could be interesting to move the current `endpoints` repository to a separate organization. I would like to use this opportunity to also change the organization name of the published artifacts (currently, we use `org.julienrf`).

I’m not very imaginative… I would like to ask you if you have ideas for an organization name. It could be `scala-endpoints` for the github organization, and `org.scala-endpoints` for the Maven organization.

julienrf:
- Organization name is now `org.endpoints4s` instead of `org.julienrf`
- Artifact names are not anymore prefixed by `endpoints-` (e.g., `endpoints-algebra` has been renamed to just `algebra`)
- Documentation is updated
- `scmInfo` points to `endpoints4s/endpoints4s`

Fixes #593


julienrf:
I should probably also rename the `endpoints` package to `endpoints4s` but I’m worried about the migration effort for users. I’ll see if I can come up with a scalafix migration rule for that before…

julienrf:
Actually, it seems to easy to implement a reliable rewrite rule that renames the package `endpoints` into `endpoints4s` in user code. So, I’m considering doing that change. Any objections?

PR #597, opened 2020-06-30
Add operationId to http4s endpoint

Fristi:
Such that it can be used for tracing purposes


julienrf:
Hmm, it looks like hijacking the `operationId` that exists for documentation purposes, no? Or do you really want your server operationId be the same as the documentation operationId?

Fristi:
Ideally this would be the same for me, as it would be tedious to redefine this. Also when someone uses your endpoint by consulting the openapi spec it would be nice to have the same operationId so talking about the same thing is easier :)

julienrf:
OK, thanks for the explanation.

PR #599, opened 2020-07-01
Add flag for using simpleName while deriving JsonSchema

### LINKED ISSUE #598: Make derived schemaName configurable
We now use  `classTagToSchemaName` which uses a `ClassTag.runtimeClass.getName`. 

When you generate code from OpenAPI definitions, it would be more elegant to drop the package name. The https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/typescript-axios.md generator generates for a package `foo.bar.Person` the following class name in TypeScript `FooBarPerson`.

For us it's not preferable, we would like to have `Person`, so having a strategy which uses `ClassTag.runtimeClass.getSimpleName` instead would be nice

Fristi:
Resolves #598

### DIFF
--- a/json-schema/json-schema-generic/src/main/scala/endpoints4s/generic/JsonSchemas.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints4s/generic/JsonSchemas.scala
@@ -284,6 +284,8 @@ trait JsonSchemas extends algebra.JsonSchemas {
     }
   }
 
+  val useSimpleName: Boolean = false
+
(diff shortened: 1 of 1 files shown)


Fristi:
Just found out that overriding is possible, maybe we can just drop this

julienrf:
Yes, the preferred way to customize this is by overriding. If that solutions works for that’s also simpler for endpoints4s :)

Fristi:
Let's stick with that!

PR #603, opened 2020-07-08
Add more status codes

### LINKED ISSUE #602: Add missing status codes
There only seems to be a barebones amount of status codes available right now, ideally we'd support all of the standardized ones.

LukaJCB:
Should fix #602


LukaJCB:
Should we add constructors for each of these as well, like we have for `ok` and `badRequest`?

tdroxler:
agree, `response(Conflict, emptyResponse)` looks good.

LukaJCB:
Yup agreed as well!

julienrf:
I’m not merging the PR now, I want to first have MiMa setup.

LukaJCB:
Good call!

julienrf:
MiMa is now enabled on `master`, do you mind rebasing your PR?

LukaJCB:
Done :)

LukaJCB:
Ah snap this is binary incompatible of course, because it's an open trait. Not sure how to move forward here, we could add default implementations that fallback to `200` and `400`, but not sure if that's a great idea.

julienrf:
> we could add default implementations that fallback to `200` and `400`, but not sure if that's a great idea.

No indeed this is not a good idea IMHO.

I’ll try to find out a solution.

julienrf:
I’m not sure we can “safely” introduce such a binary incompatibility between minor versions. I think that end-user will eventually face compilation errors such as “object creation impossible, missing implementation for…”, or runtime errors such as “NoSuchMethodError”. So, we need to consider solutions to not break bincompat.

I’m tempted to give a try to the solution that consists in providing a default implementation in the algebra (until the next major release). An issue is that if a project uses `algebra:1.1.0` to define endpoints and second project uses `client:1.0.0` to invoke the endpoints of the first project, the second project won’t see any warning or error that requires updating the client dependency to `client:1.1.0` to match the algebra version used by the first project, leading to the default implementation being used instead of the implementation provided by the client. That being said, the default implementation could consist in throwing an exception indicating the version of the interpreter required (this would help the developers to fix the issue).

This situation looks a little bit like Akka, where all the modules have to be aligned on the same patch version, and you see the error only at runtime… (and this is quite annoying)

Otherwise, I see two safe alternatives:

- bump the version number to 2.0.0. But that sucks, my goal is to reach a point where we don’t break the bincompat of the algebra so often (but maybe after merging the status codes the algebra will be complete enough and will finally stay stable…?)
- move the new status codes to a separate trait than `algebra.StatusCodes`. E.g., `trait algebra.AllStatusCodes extends algebra.StatusCodes`. This is probably the solution we should try to use as much as possible for new additions to the algebra, but in the case of the status codes, it’d be weird that half of them are in `algebra.StatusCodes` and the other half in a separate trait…

My conclusion is that, for this PR, I would go with a default implementation, in `algebra.StatusCodes` like `sys.error("version 1.1.0 of the endpoints4s algebra is not supported by this interpreter, please update the project dependencies to support this algebra version")`. 

In the future, when possible (e.g., to support multipart entities), we should prefer defining new algebra operations in new traits.

What do you think?

PR #608, opened 2020-07-24
Streamline the way servers decode request headers

julienrf:
The way server interpreters handle request headers was not consistent between interpreters.

This PR makes sure that all the servers:

- reject request whose headers are missing with a standard Bad Request response,
- return all the errors related to request headers validation (not just the first one).

### DIFF
--- a/akka-http/build.sbt
+++ b/akka-http/build.sbt
@@ -35,6 +35,7 @@ val `akka-http-server` =
       publishSettings,
       `scala 2.12 to dotty`,
       name := "akka-http-server",
+      version := "2.0.0",
       libraryDependencies ++= Seq(
         ("com.typesafe.akka" %% "akka-http" % akkaHttpVersion).withDottyCompat(scalaVersion.value),
         ("com.typesafe.akka" %% "akka-stream" % akkaActorVersion).withDottyCompat(scalaVersion.value),
--- a/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
+++ b/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
@@ -2,7 +2,7 @@ package endpoints4s.akkahttp.server
 
 import akka.http.scaladsl.marshalling.{Marshaller, ToEntityMarshaller, ToResponseMarshaller}
 import akka.http.scaladsl.model.headers.RawHeader
-import akka.http.scaladsl.model.{HttpEntity, HttpHeader, MediaTypes}
+import akka.http.scaladsl.model.{HttpEntity, HttpHeader, HttpRequest, MediaTypes}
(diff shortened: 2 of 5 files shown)


http4s/server/src/main/scala/endpoints4s/http4s/server/Endpoints.scala
```diff
@@ -361,8 +361,7 @@ trait EndpointsWithCustomErrors extends algebra.EndpointsWithCustomErrors with M
           tupler: Tupler[A, B]
       ): RequestHeaders[tupler.Out] =
         headers =>
-          fa(headers)
-            .flatMap(a => fb(headers).map(b => tupler(a, b)))
+          fa(headers).zip(fb(headers))
```
julienrf, line 364: I guess the part of this PR that deals with http4s will be a bit easier to review than the one that deals with akka-http.

I think this is a good example of how structured abstractions make it easier to reason about code… The primary abstraction of akka-http, namely `Directive`, is too powerful IMHO and this makes it hard to reason about code that uses directives.

akka-http/build.sbt
```diff
@@ -35,6 +35,7 @@ val `akka-http-server` =
       publishSettings,
       `scala 2.12 to dotty`,
       name := "akka-http-server",
+      version := "2.0.0",
```
Fristi, line 38: Why are you being specific here with the version?

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
```diff
@@ -224,7 +243,13 @@ trait EndpointsWithCustomErrors
       tuplerABC: Tupler.Aux[AB, C, Out]
   ): Request[Out] = {
     val methodDirective = convToDirective1(Directives.method(method))
-    val matchDirective = methodDirective & url.directive & headers
+    val headersDirective: Directive1[C] =
+      directive1InvFunctor.xmapPartial[Validated[C], C](
+        Directives.extractRequest.map(headers.decode),
+        validatedC => validatedC,
```
Fristi, line 249: `identity`

algebras/algebra/src/test/scala/endpoints4s/algebra/server/EndpointsTestSuite.scala
```diff
@@ -460,6 +460,35 @@ trait EndpointsTestSuite[T <: endpoints4s.algebra.EndpointsTestApi] extends Serv
       }
     }
 
+    "reject requests with missing headers" in {
```
Fristi, line 463: Wouldn't it be better to split the cases in separate tests? In case one fails the whole test fails

akka-http/build.sbt
julienrf: We need to bump the major version number because the changes on the akka-http server interpreter are not backward binary compatible with the previous release.

The algebra stays with a 1.x version because it is binary compatible.

PR #611, opened 2020-08-12
change http4s server to support effectful errorhandling

GrafBlutwurst:
I am not entirely sure if the way I handled `handleXError` with `handleXErrorEffect` is very smart. The idea was to retain compatibility.

### DIFF
--- a/http4s/server/src/main/scala/endpoints4s/http4s/server/Endpoints.scala
+++ b/http4s/server/src/main/scala/endpoints4s/http4s/server/Endpoints.scala
@@ -107,16 +107,16 @@ trait EndpointsWithCustomErrors extends algebra.EndpointsWithCustomErrors with M
                 try {
                   implementation(a)
                     .map(response)
-                    .recover {
-                      case NonFatal(t) => handleServerError(t)
+                    .recoverWith {
+                      case NonFatal(t) => handleServerErrorEffect(t)
                     }
                 } catch {
-                  case NonFatal(t) => handleServerError(t).pure[Effect]
+                  case NonFatal(t) => handleServerErrorEffect(t)
                 }
               case Left(errorResponse) => errorResponse.pure[Effect]
             })
         } catch {
-          case NonFatal(t) => Some(handleServerError(t).pure[Effect])
+          case NonFatal(t) => Some(handleServerErrorEffect(t))
         }
       }
       Function.unlift(handler)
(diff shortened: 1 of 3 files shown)


http4s/server/src/main/scala/endpoints4s/http4s/server/Endpoints.scala
```diff
@@ -374,6 +383,9 @@ trait EndpointsWithCustomErrors extends algebra.EndpointsWithCustomErrors with M
   def handleClientErrors(invalid: Invalid): http4s.Response[Effect] =
     clientErrorsResponse(invalidToClientErrors(invalid))
 
+  def handleClientErrorsEffect(invalid: Invalid): Effect[http4s.Response[Effect]] =
```
julienrf, line 386: We should document that this method is called when decoding a request failed. By the way, do we still call `handleClientErrors`? If not, we should update its documentation accordingly. Or, maybe better, we should probably only keep one method (the one that returns an `Effect[Response[Effect]]`), otherwise users may get confused: which method needs to be overridden?

I would be in favor of breaking the backward compatibility and to change the return type of `handleClientErrors` instead of introducing another operation. What do others think?

GrafBlutwurst: `handleClientErrors` shouldn't be called anywhere internally anymore if i didn't mess up. I also agree with removing the pure handle functions since the rewrite to accomodate the effectful version on use site should be fairly minimal. Should I hold on the changes until the discussion played out or already switch it around?

julienrf: > Should I hold on the changes until the discussion played out or already switch it around?

Let’s do the switch already :)

GrafBlutwurst: done

julienrf: I would rather keep the same name as before, `handleClientErrors`. Or do you think we should keep the suffix `Effect`? If that’s the case, we need to update the manual to use the new name.

GrafBlutwurst: I agree it makes sense to keep the name, I was just too quick with my previous reiteration.

julienrf:
Thank you Dominic! I’m really sorry for the delay…

PR #637, opened 2020-09-18
Improve support of case classes’ fields with default values

### LINKED ISSUE #635: JSON schemas of objects containing optional properties with default values are not well supported
It is common to evolve a schema by adding a new optional field for which we fallback to a default value if it is missing.

For instance, consider the following case class modeling a point:

~~~ scala
case class Point(latitude: Double, longitude: Double)
~~~

A corresponding JSON schema would be the following:

~~~ scala
val pointSchema: JsonSchema[Point] = (
  field[Double]("latitude") zip
  field[Double]("longitude")
).as[Point]
~~~

To evolve our data type to include the elevation above ground level (with a default value of `0`), in Scala we just add a field with a default value:

~~~ scala
case class Point(latitude: Double, longitude: Double, elevation: Double = 0)
~~~

A backward compatible JSON schema would have an optional `elevation` field. When the `elevation` is missing, we want to decode a `Point` with an elevation of `0`. Such a schema can be defined as follows:

~~~ scala
val pointSchema: JsonSchema[Point] = (
  field[Double]("latitude") zip
  field[Doubel]("longitude") zip
  optField[Double]("elevation").xmap(_.getOrElse(0))(Some(_))
).as[Point]
~~~

We might want to introduce a method for modeling a JSON schema with an optional field decoded to a default value:

~~~ scala
final def optFieldWithDefault[A: JsonSchema](name: String, defaultValue: A, docs: Option[String] = None): Record[A] =
  optField(name, docs).xmap(_.getOrElse(defaultValue))(Some(_))
~~~

So that the above snippet can be simplified to:

~~~ scala
val pointSchema: JsonSchema[Point] = (
  field[Double]("latitude") zip
  field[Doubel]("longitude") zip
  optFieldWithDefault[Double]("elevation", 0)
).as[Point]
~~~

Last, generic derivation could detect when a case class field has a default value and map it to an `optFieldWithDefault` in such a case. Libraries like Circe already derive such codecs.

julienrf:
Fixes #635 (see description in the issue)

### DIFF
--- a/json-schema/json-schema-generic/src/main/scala/endpoints4s/generic/JsonSchemas.scala
+++ b/json-schema/json-schema-generic/src/main/scala/endpoints4s/generic/JsonSchemas.scala
@@ -10,6 +10,7 @@ import shapeless.{
   Annotations,
   CNil,
   Coproduct,
+  Default,
   Generic,
   HList,
   HNil,
@@ -74,8 +75,12 @@ trait JsonSchemas extends algebra.JsonSchemas {
       with GenericDescriptions
       with GenericTitles {
 
-    implicit def emptyRecordCase: DocumentedGenericRecord[HNil, HNil] =
+    @deprecated("Replaced by `genericEmptyRecord`", "1.2.0")
+    def emptyRecordCase: DocumentedGenericRecord[HNil, HNil] =
       (_: HNil) => emptyRecord.xmap[HNil](_ => HNil)(_ => ())
+
+    implicit def genericEmptyRecord: DerivedGenericRecord[HNil, HNil, HNil] =
+      (_: HNil, _: HNil) => emptyRecord.xmap[HNil](_ => HNil)(_ => ())
 
     implicit def singletonCoproduct[L <: Symbol, A](implicit
         labelSingleton: Witness.Aux[L],
@@ -95,7 +100,8 @@ trait JsonSchemas extends algebra.JsonSchemas {
   trait GenericJsonSchemaLowPriority extends GenericJsonSchemaLowLowPriority {
     this: GenericJsonSchema.type =>
 
-    implicit def consRecord[L <: Symbol, H, T <: HList, DH <: Option[docs], DT <: HList](implicit
+    @deprecated("Replaced by `consGenericRecord`", "1.2.0")
+    def consRecord[L <: Symbol, H, T <: HList, DH <: Option[docs], DT <: HList](implicit
         labelHead: Witness.Aux[L],
         jsonSchemaHead: JsonSchema[H],
         jsonSchemaTail: DocumentedGenericRecord[T, DT]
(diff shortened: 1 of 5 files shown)


json-schema/json-schema-generic/src/main/scala/endpoints4s/generic/JsonSchemas.scala
```diff
@@ -109,7 +115,38 @@ trait JsonSchemas extends algebra.JsonSchemas {
...
+        jsonSchemaTail: DerivedGenericRecord[T, DT, VT]
+    ): DerivedGenericRecord[FieldType[L, H] :: T, DH :: DT, VH :: VT] =
+      new DerivedGenericRecord[FieldType[L, H] :: T, DH :: DT, VH :: VT] {
+        def record(docs: DH :: DT, defaultValues: VH :: VT) = {
+          val recordHead = defaultValues.head match {
+            case Some(defaultValue) =>
+              optFieldWithDefault[H](labelHead.value.name, defaultValue, docs.head.map(_.text))
+            case None =>
+              field[H](labelHead.value.name, docs.head.map(_.text))
+          }
```
julienrf, line 138: This is the logic that maps case class fields with a default value to an optional JSON field with default value, and regular case class fields to a mandatory JSON field.

json-schema/json-schema-generic/src/main/scala/endpoints4s/generic/JsonSchemas.scala
```diff
@@ -74,9 +75,13 @@ trait JsonSchemas extends algebra.JsonSchemas {
       with GenericDescriptions
       with GenericTitles {
 
-    implicit def emptyRecordCase: DocumentedGenericRecord[HNil, HNil] =
+    @deprecated("Replaced by `genericEmptyRecord`", "1.2.0")
+    def emptyRecordCase: DocumentedGenericRecord[HNil, HNil] =
```
julienrf, line 79: Existing methods involved in the generic derivation process are not anymore implicit so that there is no risk for the compiler to use them when deriving schemas.
We don’t remove them to keep backward binary compatibility, but we mark them as `@deprecated` to signal that we can remove them in the next major release.

gmarti (approved):

gmarti (approved):

gmarti:
lgtm

harpocrates:
This is a great change, but in case you weren't already aware: this will break backwards compatibility for classes like

```scala
case class Foo(a: Option[Int] = None)

genericJsonRecord[Foo] // no longer derives an instance - you must remove `= None`
```

I prefer the new derivation (most of the `Option` fields we had were anyways to work around this missing feature), and gladly pay the breakage price.

julienrf:
Thanks, I didn’t know this would break things like that.

We might be able to special case the derivation process to skip optional fields with a default value (so, only fields with a default value whose type is not `Option` would be handled by the derivation mechanism). But this would make the derivation process more complex at the same time, so I prefer if we don’t support optional fields with a default value (like now).

PR #641, opened 2020-09-18
Make Http header extraction case insensitive in http4s server interpreter

### LINKED ISSUE #638: The HTTP headers are compared case sensitive in the http4s server interpreter

rreckel:
Fixes #638

### DIFF
--- a/algebras/algebra/src/test/scala/endpoints4s/algebra/server/EndpointsTestSuite.scala
+++ b/algebras/algebra/src/test/scala/endpoints4s/algebra/server/EndpointsTestSuite.scala
@@ -488,6 +488,19 @@ trait EndpointsTestSuite[T <: endpoints4s.algebra.EndpointsTestApi] extends Serv
       }
     }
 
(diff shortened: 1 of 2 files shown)


julienrf:
Thank you @rreckel! The changes look good but do you mind adding a test in file `algebras/algebra/src/test/scala/endpoints4s/algebra/server/EndpointsTestSuite.scala` (you can see how the existing tests look like at line 478)? We might have to adapt the other interpreters to be consistent. I can help with that in case you’re not familiar with akka-http or Play.
Ideally, we should do the same for client interpreters (make sure they are insensitive the header’s case), but this could be done in a separate PR.

rreckel:
Ok, I added a very trivial test which is a copy/paste of your mentioned test with lower case headers. I hope that is enough. If not, or if you wish to add a different test case, just let me know.

julienrf:
Yes, thank you the test look good to me! It’s nice to see that the other interpreters already pass the test :) (the test suite is run against all the server interpreters)

julienrf (approved):

PR #664, opened 2020-10-29
Add JsonSchemas for date types

### LINKED ISSUE #233: Support for dates in JsonSchema
Should be pretty easy to implement `JsonSchema[java.time.LocalDate]`, but that's not everything.
In OpenAPI specifications I'd expect a schema like `{ "type": "string", "format": "date" }`.

It seems impossible for the user to define a schema with a custom format, but we could add schemas for well-known types. The comments in https://github.com/julienrf/endpoints/pull/136 suggest exactly that. I'd say `java.time` is not third-party and could be added right away.

Other types (like `OffsetDateTime`) could be added as well, but I'm not so sure which ones make sense. I only stumbled upon `LocalDate` because I use it in a REST service.

julienrf:
Add `JsonSchema` instances for `Instant`, `OffsetDateTime`, and `Duration`. The JSON representation is a string containing an ISO 8601 of the date or duration.

Fixes #233

### DIFF
--- a/json-schema/build.sbt
+++ b/json-schema/build.sbt
@@ -19,7 +19,11 @@ val `json-schema` =
       (Compile / boilerplateSource) := baseDirectory.value / ".." / "src" / "main" / "boilerplate"
     )
     .enablePlugins(spray.boilerplate.BoilerplatePlugin)
-    .jsConfigure(_.disablePlugins(ScoverageSbtPlugin))
+    .jsConfigure(
+      _.settings(
+        libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0" % Test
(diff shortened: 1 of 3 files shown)


julienrf:
@stefanholzmueller is this fine to you?

Note that I didn’t cover `LocalDate` as I’m not sure it should be the preferred data type to communicate between remote computers since the time zone is not embedded in the `LocalDate` (so it’s easy to misinterprete it).

stefanholzmueller (approved):
No objections about leaving out `LocalTime`. You're right, time-zone problems are terrible, and nowadays I predominantly use `OffsetDateTime`. If someone really needs another date type, they can write their own `JsonSchema`.

Anyways, it is great to support date-times out of the box now!

Issue #542, opened 2020-04-15
numeric JSON schemas don’t support a format (and other documentation attributes)


julienrf:
According to the OpenAPI specification, it is possible to set attributes such as `format`, `minimum`, etc. to numeric JSON schemas:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#schemaObject

We should provide a new `JsonSchema` constructor for each numeric type (`Int`, `Long`, `Double`, `Float`). These constructors should take as parameters the currently missing documentation attributes.

The default (implicit) `JsonSchema` for these types should not set these parameters. (Like we do with `stringJsonSchema`)

bmeesters:
I am not sure if I am understanding you correctly. But the format determines the type right?

![image](https://user-images.githubusercontent.com/32057771/99643750-5e6d4580-2a4d-11eb-93c6-5ab064225875.png)

So each constructor has optional properties like `minimum`, `'maxiumum`, etc, and the type `Float`, `Double`, `Int`, `Long` determines the possible format (if any). I.e. you cannot use JsonSchema[Int] with format `float`.

WDYT?

julienrf:
Indeed, `format` is already added to the schemas according to the OpenAPI Specification. I think we need only need to add support for the optional properties that can be attached to numeric schemas: https://json-schema.org/understanding-json-schema/reference/numeric.html.

bmeesters:
Yes, sorry, I was a little ask the question again (so I deleted it), without properly looking at what was already there. I think I have a working solution. Though it needs some fine tuning to ensure it is backwards compatible. I'll make a draft PR so we can look at the solution.

PR #684, opened 2020-11-19
Add numeric properties/constraints in JsonSchemas

bmeesters:
closes https://github.com/endpoints4s/endpoints4s/issues/542

### DIFF
--- a/json-schema/json-schema-playjson/src/test/scala/endpoints4s/playjson/JsonSchemasTest.scala
+++ b/json-schema/json-schema-playjson/src/test/scala/endpoints4s/playjson/JsonSchemasTest.scala
@@ -382,6 +382,13 @@ class JsonSchemasTest extends AnyFreeSpec {
     assertError(intOrBoolean, JsString("foo"), "Invalid value: \"foo\"")
   }
 
+  "constrained numeric value" in {
+    testRoundtrip(constraintNumericSchema, JsNumber(6), 6)
+    assertError(constraintNumericSchema, JsNumber(10), createNumericErrorMessage(10))
+    assertError(constraintNumericSchema, JsNumber(-1), createNumericErrorMessage(-1))
+    assertError(constraintNumericSchema, JsNumber(5), createNumericErrorMessage(-5))
+  }
+
   private def testRoundtrip[A](
       jsonSchema: JsonSchema[A],
       json: JsValue,
--- a/json-schema/json-schema-playjson/src/test/scala/endpoints4s/playjson/JsonSchemasTest.scala
+++ b/json-schema/json-schema-playjson/src/test/scala/endpoints4s/playjson/JsonSchemasTest.scala
@@ -386,7 +386,7 @@ class JsonSchemasTest extends AnyFreeSpec {
     testRoundtrip(constraintNumericSchema, JsNumber(6), 6)
     assertError(constraintNumericSchema, JsNumber(10), createNumericErrorMessage(10))
     assertError(constraintNumericSchema, JsNumber(-1), createNumericErrorMessage(-1))
-    assertError(constraintNumericSchema, JsNumber(5), createNumericErrorMessage(-5))
+    assertError(constraintNumericSchema, JsNumber(5), createNumericErrorMessage(5))
   }
 
   private def testRoundtrip[A](
--- a/json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
@@ -17,11 +17,12 @@ case class NumericConstraints[A](
     (p2s(minimum, "minimum") ++
       p2s(exclusiveMinimum, "exclusiveMinimum") ++
       p2s(maximum, "maximum") ++
-      p2s(exclusiveMaximum, "exclusiveMaximum")).mkString(", ")
+      p2s(exclusiveMaximum, "exclusiveMaximum") ++
+      p2s(multipleOf, "multipleOf")).mkString(", ")
 
   /** Check whether the value satisfies all the constraints */
   def satisfiedBy(value: A): Boolean = {
-    def checkMultipleOf(bound: A) = mult.multipleOf(bound, value)
+    def checkMultipleOf(bound: A) = mult.multipleOf(value, bound)
 
     def checkMinimum(bound: A) =
       if (exclusiveMinimum.getOrElse(false)) bound < value
@@ -31,9 +32,9 @@ case class NumericConstraints[A](
       if (exclusiveMaximum.getOrElse(false)) bound > value
       else bound >= value
 
-    minimum.forall(checkMinimum) && maximum.forall(checkMaximum) && multipleOf.forall(
-      checkMultipleOf
-    )
+    minimum.forall(checkMinimum) &&
+    maximum.forall(checkMaximum) &&
+    multipleOf.forall(checkMultipleOf)
   }
 
   private def p2s[B](opt: Option[B], name: String): List[String] =
--- a/json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
@@ -4,21 +4,59 @@ package endpoints4s
   * valid w.r.t. to the properties.
   * @group operations
   */
-case class NumericConstraints[A](
-    minimum: Option[A] = None,
-    exclusiveMinimum: Option[Boolean] = None,
-    maximum: Option[A] = None,
-    exclusiveMaximum: Option[Boolean] = None,
-    multipleOf: Option[A] = None
-)(implicit val ord: Ordering[A], mult: MultipleOf[A]) {
+final class NumericConstraints[A] private (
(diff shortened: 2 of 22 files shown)


bmeesters:
Hello @julienrf, I set up a solution for the properties on numeric values. It's not done yet (no new unit tests & the open-api swagger is not updated yet), but I do think the solution will work & is backwards compatible if cleaned up a bit. 

The backwards-compatibility was a bit tricky, since adding new abstract methods breaks it for interpreters defined outside endpoints4s, so it was not possible to let the old property-less methods depend on the new methods with properties. Now, the abstract algebra adds a few new methods that refer to the abstract existing methods so nothing breaks, and the interpreters override both to make the relationship vice-versa in the actual implementation. It's not very pretty, but it provides the new functionality as opt-in without breaking any interpreters defined outside of endpoints4s. Please let me know if you have any ideas on how to improve it.

julienrf:
> The backwards-compatibility was a bit tricky, since adding new abstract methods breaks it for interpreters defined outside endpoints4s, so it was not possible to let the old property-less methods depend on the new methods with properties. Now, the abstract algebra adds a few new methods that refer to the abstract existing methods so nothing breaks, and the interpreters override both to make the relationship vice-versa in the actual implementation. It's not very pretty, but it provides the new functionality as opt-in without breaking any interpreters defined outside of endpoints4s.

Nice trick :sweat_smile: I think it does what we want and I can’t find a case where it would cause an issue with existing code, so that’s good to me!

MiMa still flags an incompatibility about an abstract method introduced by the case class `NumericConstraints` in the trait `endpoints4s.algebra.JsonSchemas`. Would it be possible to move the case class definition out of the trait?

bmeesters:
Hey @julienrf, I think this is ready-for-review now. The build seems to fail on:

```
[error] /home/travis/build/endpoints4s/endpoints4s/openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala:496:69: value Numeric of type scala.math.Numeric.type does not take type parameters.
[error]   private def toD[A: Numeric](opt: Option[A]) = opt.map(a => Numeric[A].toDouble(a))
```

Which surprises me, since `Numeric` actually as an apply method that summons the implicit. Also I get no errors locally. Any idea what's going on?

json-schema/json-schema-playjson/src/test/scala/endpoints4s/playjson/JsonSchemasTest.scala
```diff
@@ -382,6 +382,13 @@ class JsonSchemasTest extends AnyFreeSpec {
     assertError(intOrBoolean, JsString("foo"), "Invalid value: \"foo\"")
   }
 
+  "constrained numeric value" in {
+    testRoundtrip(constraintNumericSchema, JsNumber(6), 6)
+    assertError(constraintNumericSchema, JsNumber(10), createNumericErrorMessage(10))
+    assertError(constraintNumericSchema, JsNumber(-1), createNumericErrorMessage(-1))
+    assertError(constraintNumericSchema, JsNumber(5), createNumericErrorMessage(-5))
```
julienrf, line 389:
```suggestion
    assertError(constraintNumericSchema, JsNumber(5), createNumericErrorMessage(5))
```

json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
```diff
@@ -0,0 +1,42 @@
+package endpoints4s
+
+/** Possible restrictions on the numeric value used. Needs an instance for [[Ordering]] to check whether values are
+  * valid w.r.t. to the properties.
+  * @group operations
+  */
+case class NumericConstraints[A](
```
julienrf, line 7: I know it’s a bit boring to write, but I think it would be better to not have case classes in the public algebra API. Would it be possible to make them plain classes, but to override `equals`, `hashCode`, and `toString`?

julienrf (commented):
Thank you @bmeesters, I think this is a great addition :)

I’ve only left a minor comment about a test that doesn’t pass, and another about the usage of case classes.

json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
bmeesters: Yes, no problem, think that would be the smart thing todo, it already paid off in the Open API model when adding these optional constraints. I'll do it after the weekend!

bmeesters:
@julienrf I have processed your review comments. Note that there is still a build failure, though I don't understand why this is happening.

json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
```diff
@@ -0,0 +1,87 @@
...
+  * valid w.r.t. to the properties.
+  * @group operations
+  */
+final class NumericConstraints[A] private (
+    val minimum: Option[A] = None,
+    val exclusiveMinimum: Option[Boolean] = None,
+    val maximum: Option[A] = None,
+    val exclusiveMaximum: Option[Boolean] = None,
+    val multipleOf: Option[A] = None
+)(implicit val ord: Ordering[A], mult: MultipleOf[A])
```
julienrf, line 13: Why do we need `ord` to be a `val`?

julienrf (approved):

json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
bmeesters: We don't, not sure why I did that. I'll remove it. I'll also use `implicitly` for `Numeric`, maybe that will either fix the build problem, or at least give a better error message.

julienrf: Great, yes I think removing the `val` here, and using `implicitly` in the test failure should help.

json-schema/json-schema/src/main/scala/endpoints4s/NumericConstraints.scala
```diff
@@ -0,0 +1,87 @@
+package endpoints4s
+
+/** Possible restrictions on the numeric value used. Needs an instance for [[Ordering]] to check whether values are
```
julienrf, line 3:
```suggestion
/** Possible restrictions on the numeric value used. Needs an instance for `Ordering` to check whether values are
```

Seems like scaladoc does not know how to link to `Ordering`…

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -83,7 +88,12 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
         format: Option[String] = None,
         description: Option[String] = None,
         example: Option[ujson.Value] = None,
-        title: Option[String] = None
+        title: Option[String] = None,
+        minimum: Option[Double] = None,
+        exclusiveMinimum: Option[Boolean] = None,
+        maximum: Option[Double] = None,
+        exclusiveMaximum: Option[Boolean] = None,
+        multipleOf: Option[Double] = None
```
julienrf, line 96: This change breaks backward binary compatibility. We need to bump the major version of all the interpreters.

bmeesters: All? The others don't depend on the openapi right? Do we need to also make normal classes for these (maybe in a separate PR)?

julienrf: All the server interpreters depend on the `openapi` interpreter, so we’ll have to bump them as well.

Yes, turning these classes into normal classes would help, but IMHO major version bumps in the interpreters is not an issue, what’s really important is to keep backward binary compatibility for the algebra as long as possible (because the algebra is the artifact that is shared between multiple services, in a microservices architecture, and we don’t want to create incompatibilities between microservices).

bmeesters: Ah right that makes sense. I don't know where to update the version though :sweat_smile: Btw the Numeric problem is solved, though there is some other error that I occasionally see (something with a server not responding).

julienrf: I’ve bumped the server interpreters, but I’ve just realized that client interpreters also depend on the `openapi` module, for some reason, so we’ll have to bump them as well…

bmeesters: Shouldn't we check whether that is necessary? It seems to be that both server and clients are independent, though I might be missing something obvious.

julienrf: Well, the clients don’t really use the OpenAPI stuff, but the `JsonSchemas` interpreter based on `ujson` to decode and encode JSON entities. I think it’s OK to bump their version as well.

bmeesters:
Thanks @julienrf, appreciate you helping me get this over the finish line!

julienrf:
Thank you @bmeesters for the contribution :wink:

PR #685, opened 2020-11-19
Add convenience method withSecuritySchemas to OpenAPI

marconilanna:


julienrf:
Thanks @marconilanna for the PR!

While I understand the benefits of this method, I would rather not have it. In the [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#componentsObject), the property `securitySchemes` belongs to the `components` object, so I think it’s simpler to have it only in our `Components` class, to mirror that. There is a `security` property in the [OpenAPI object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#oasObject) that we might want to add eventually, it which should not be confused with the security schemes.

What do others think?

marconilanna:
@julienrf What's the recommended way then to add a `SecurityScheme` to `OpenAPI`?

As it is now, this is how we had to do it:

```scala
  val api: OpenApi = {
    val t = openApi(
      Info(...)
    )(
      // endpoints ...
    )

    t.withComponents(
      t.components
        .withSecuritySchemas(
          //...
        )
    )
  }
```

With the proposed method, it would be simplified to:

```scala
  val api: OpenApi = openApi(
    Info(...)
  )(
    // endpoints ...
  ).withSecuritySchemas(
    //...
  )
```

julienrf:
The way you wrote it is the recommended way. I know it’s more steps that with your shortcut but I don’t think this is something we need to do very often. You can use `.pipe` instead of creating a `val`, as an slight alternative:

~~~ scala
import scala.util.chaining._
val api: OpenApi =
  openApi(
    Info(...)
  )(
    // endpoints ...
  ).pipe { api =>
    api.withComponents(
      api.components.withSecuritySchemas(...)
    )
  }
~~~

I wonder if we could have lenses/optics to simplify this, though.

julienrf:
I’m closing the PR but feel free to re-open it if you want to continue discussing this point.

Issue #60, opened 2017-10-30
HList based tupler


Fristi:
I've used shapeless' hlist's to build up product types as seen here: https://github.com/Fristi/endpoints-proto/blob/master/algebra/src/main/scala/itinere/Tupler.scala.

Advantages:
- No 22 tuple limit
- We can use the `Generic` machinery of shapeless to map to case classes

Krever:
@julienrf Do you think it pays the cost of introducing first dependency to core algebra?

Idea looks interesting and for sure worth investigating but I'm not sure if its beneficial enough. Mapping to case classes may be a really nice feature though. 

I had a different idea at some point, to remove tupler at all. User then has to deal with `Unit`s but it would simplify the api a lot.

julienrf:
In the beginning I considered using shapeless’ `HList` instead of tuples but I thought it was not worth introducing a dependency on shapeless just to get an `HList` definition (we use no operations on `HList` and no feature of shapeless). But in practice I often have shapeless in my classpath anyway because I do typeclass derivation in some places. So, why not use shapeless’ `HList` instead of tuples, if that also makes life easier for case classes mapping.

Note that it should still be possible to map to case classes without introducing shapeless in the core, but at the (cheap) cost of an additional implicit parameter.

julienrf:
One advantage of using tuples is that the syntax to “call” an endpoint is very similar to calling a method with several parameters (excepted that it’s a single tuple parameter):

~~~ scala
val foo: Endpoint[(Int, String), Int] = …
foo((42, "bar"))
~~~

On the other hand, if we use `HList`s we will have to write the following:

~~~ scala
val foo: Endpoint[Int :: String :: HNil, Int] = …
foo(42 :: "bar" :: HNil)
~~~

Which is less nice imho. We could support passing a tuple as parameter as well, by having an `apply` method taking an implicit `Generic` that would convert the tuple into the expected `HList` but then error messages in case of type mismatch will be disguised into “implicit not found” errors :(

julienrf:
I think we should wait for Scala 3 to implement this feature. In Scala 3, tuples can be used like Shapeless’ HLists.

harpocrates:
[...]

PR #715, opened 2021-01-03
Fix for ambiguous tupler

### LINKED ISSUE #713: Ambiguous Tupler since 0.11.x
```scala
    implicitly[endpoints.Tupler[(String, String, String),                                         (String, String, String, String)]]
    implicitly[endpoints.Tupler[(String, String, String, String),                                 (String, String, String, String)]]
    implicitly[endpoints.Tupler[(String, String, String, String, String),                         (String, String, String, String)]]
    implicitly[endpoints.Tupler[(String, String, String, String, String, String),                 (String, String, String, String)]]
    implicitly[endpoints.Tupler[(String, String, String, String, String, String, String),         (String, String, String, String)]]
    implicitly[endpoints.Tupler[(String, String, String, String, String, String, String, String), (String, String, String, String)]]
```
results in
```
cmd2.sc:1: ambiguous implicit values:
 both method tupler1And4 in trait Tupler3 of type [A, B, C, D, E]endpoints.Tupler[A,(B, C, D, E)]{type Out = (A, B, C, D, E)}
 and method tupler3Append in trait TuplerAppend of type [T1, T2, T3, T4]endpoints.Tupler[(T1, T2, T3),T4]{type Out = (T1, T2, T3, T4)}
 match expected type endpoints.Tupler[(String, String, String),(String, String, String, String)]
val res2_0 = implicitly[endpoints.Tupler[(String, String, String),                                         (String, String, String, String)]]
                       ^
cmd2.sc:2: ambiguous implicit values:
 both method tupler1And4 in trait Tupler3 of type [A, B, C, D, E]endpoints.Tupler[A,(B, C, D, E)]{type Out = (A, B, C, D, E)}
 and method tupler4Append in trait TuplerAppend of type [T1, T2, T3, T4, T5]endpoints.Tupler[(T1, T2, T3, T4),T5]{type Out = (T1, T2, T3, T4, T5)}
 match expected type endpoints.Tupler[(String, String, String, String),(String, String, String, String)]
  val res2_1 = implicitly[endpoints.Tupler[(String, String, String, String),                                 (String, String, String, String)]]
                         ^
cmd2.sc:3: ambiguous implicit values:
 both method tupler1And4 in trait Tupler3 of type [A, B, C, D, E]endpoints.Tupler[A,(B, C, D, E)]{type Out = (A, B, C, D, E)}
 and method tupler5Append in trait TuplerAppend of type [T1, T2, T3, T4, T5, T6]endpoints.Tupler[(T1, T2, T3, T4, T5),T6]{type Out = (T1, T2, T3, T4, T5, T6)}
 match expected type endpoints.Tupler[(String, String, String, String, String),(String, String, String, String)]
  val res2_2 = implicitly[endpoints.Tupler[(String, String, String, String, String),                         (String, String, String, String)]]
                         ^
cmd2.sc:4: ambiguous implicit values:
 both method tupler1And4 in trait Tupler3 of type [A, B, C, D, E]endpoints.Tupler[A,(B, C, D, E)]{type Out = (A, B, C, D, E)}
 and method tupler6Append in trait TuplerAppend of type [T1, T2, T3, T4, T5, T6, T7]endpoints.Tupler[(T1, T2, T3, T4, T5, T6),T7]{type Out = (T1, T2, T3, T4, T5, T6, T7)}
 match expected type endpoints.Tupler[(String, String, String, String, String, String),(String, String, String, String)]
  val res2_3 = implicitly[endpoints.Tupler[(String, String, String, String, String, String),                 (String, String, String, String)]]
                         ^
cmd2.sc:5: ambiguous implicit values:
 both method tupler1And4 in trait Tupler3 of type [A, B, C, D, E]endpoints.Tupler[A,(B, C, D, E)]{type Out = (A, B, C, D, E)}
 and method tupler7Append in trait TuplerAppend of type [T1, T2, T3, T4, T5, T6, T7, T8]endpoints.Tupler[(T1, T2, T3, T4, T5, T6, T7),T8]{type Out = (T1, T2, T3, T4, T5, T6, T7, T8)}
 match expected type endpoints.Tupler[(String, String, String, String, String, String, String),(String, String, String, String)]
  val res2_4 = implicitly[endpoints.Tupler[(String, String, String, String, String, String, String),         (String, String, String, String)]]
                         ^
cmd2.sc:6: ambiguous implicit values:
 both method tupler1And4 in trait Tupler3 of type [A, B, C, D, E]endpoints.Tupler[A,(B, C, D, E)]{type Out = (A, B, C, D, E)}
 and method tupler8Append in trait TuplerAppend of type [T1, T2, T3, T4, T5, T6, T7, T8, T9]endpoints.Tupler[(T1, T2, T3, T4, T5, T6, T7, T8),T9]{type Out = (T1, T2, T3, T4, T5, T6, T7, T8, T9)}
 match expected type endpoints.Tupler[(String, String, String, String, String, String, String, String),(String, String, String, String)]
  val res2_5 = implicitly[endpoints.Tupler[(String, String, String, String, String, String, String, String), (String, String, String, String)]]
                         ^
Compilation Failed
```

blast-hardcheese:
Resolves #713

Initially I thought it was as simple as bumping `TuplerAppend` to `[4..21`, but I was apparently mistaken. I don't _quite_ understand why 3 is the magic number here, but I'm perfectly happy to revise my approach to keep `1And4` if there's a preferred way to approach this.

Currently I'm stuck, as the instances are defined on `Tupler`'s companion, so I can't even disable the offending implicit via name collision.

(PS: I had to recreate this PR entirely, as apparently force pushing is disallowed on this repo -- I've never seen that setting used before, could someone kindly explain why it is desirable?)


julienrf:
Thank you @blast-hardcheese, I will have a look this week.

blast-hardcheese:
For future contributors, the one-liner for generating the unit tests here is:

```scala
@ val terms = List("a", "b", "c", "d", "e", "f", "g", "h", "i").map(Term.Name(_))
terms: List[Term.Name] = List(Term.Name("a"), Term.Name("b"), Term.Name("c"), Term.Name("d"), Term.Name("e"), Term.Name("f"), Term.Name("g"), Term.Name("h"), Term.Name("i"))

@
println(q"""..${terms.reverse.tails.map(_.reverse).flatMap(xs => (1 until xs.length).map(xs.splitAt(_)).map({ case (fst, snd) => q"""tupling(${if (fst.length > 1) { q"(..${fst})" } else { fst.head } }, ${if (snd.length > 1) { q"(..${snd})" } else { snd.head } }): (..${xs.map { case Term.Name(s) => Type.Name(s.toUpperCase) } })""" })).toList}""".syntax)
{
  tupling(a, (b, c, d, e, f, g, h, i)): (A, B, C, D, E, F, G, H, I)
  tupling((a, b), (c, d, e, f, g, h, i)): (A, B, C, D, E, F, G, H, I)
  tupling((a, b, c), (d, e, f, g, h, i)): (A, B, C, D, E, F, G, H, I)
  ...
```

then resolving the compiler errors (as the return type ascriptions are all the same, most of them will be wrong, but the compiler will tell you what the correct ones are). This was just to cover enough area to try to reduce the likelihood of this in the future.

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
```diff
@@ -5,18 +5,45 @@ class TuplerTests {
...
+    tupling(a, (b, c, d))  : (A, (B, C, D))
+    tupling((a, b, c, d), e)  : (A, B, C, D, E)
+    tupling((a, b, c), (d, e)): (A, B, C, (D, E))
     tupling((a, b), (c, d, e)): (A, B, C, D, E)
-    tupling((a, b, c), d): (A, B, C, D)
-    tupling((a, b, c, d), e): (A, B, C, D, E)
-    tupling((a, b, c, d, e), f): (A, B, C, D, E, F)
-    tupling(a, ()): A
-    tupling((), a): A
+    tupling(a, (b, c, d, e))  : (A, (B, C, D, E))
```
julienrf, line 20: We should get a flat tuple of type `(A, B, C, D, E)` here. The line you deleted from `Tupler.scala` should produce such a result.

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
```diff
@@ -5,18 +5,45 @@ class TuplerTests {
...
+    tupling((a, b, c, d, e), f)  : (A, B, C, D, E, F)
+    tupling((a, b, c, d), (e, f)): (A, B, C, D, (E, F))
```
julienrf, line 22: Here as well, I would like to return a flat tuple `(A, B, C, D, E, F)`

julienrf (commented):
Thanks @blast-hardcheese for investigating this issue.

I don’t think it is safe or correct to remove `tupler1and4`. Unfortunately, the current approach based on tuples and lots of implicit definitions does not scale.

I guess the “best” way to solve your problem would be to add more specific implicit definitions for `Tupler[(A, B, C), (D, E, F, G)]` and so on for your use cases… You can do that in your project, by defining these implicit definitions in a trait and mixing this trait in your algebra.

Ultimately, I believe the right way to address the issue would be to switch to an inductive representation of tuples---an `HList`, so that we could just concatenate them with a few rules. This idea is tracked by #60. I’m not sure whether we should wait one more year before we can leverage Scala 3 tuples, or if we should fix the problem now by using shapeless.

blast-hardcheese:
Understood, regarding 1and4.

To justify my attempt to remove `1and4`, it seemed like that was the most straightforward way to at least make all representations compile out of the box, as there are legitimate programs that are no longer representable past 0.9.

I'll see if I can resolve this by addressing your feedback here, as well as fitting new tuplers in the hierarchy in order to resolve the ambiguity.

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
blast-hardcheese: Yep, confirmed. I'll not break this with further work.

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
blast-hardcheese: This would be a change to tupler. I've updated the PR with a reduced diff in favor of easy reviewing, I'll re-sort towards the end of this process.

If at all possible, it would be ideal to permit rebasing and force-push for branches, but if you'll squash at the end I'll just not be concerned about producing unnecessary intermediate commits here.

blast-hardcheese: Also, sorry, to what end are we doing this?

You asked about
```scala
    tupling((a, b, c, d), (e, f)): (A, B, C, D, (E, F))
```
but
```scala
    tupling((a, b, c), (d, e, f)): (A, B, C, (D, E, F))
    tupling((a, b), (c, d, e, f)): ((A, B), C, D, E, F)
    tupling(a, (b, c, d, e, f)): (A, (B, C, D, E, F))
```
would also not be covered by this. Isn't this just that we only have so many tupler extensions so far? [I've worked around this when attempting to generate tuple structures compatible with the current Tuple semantics](https://github.com/twilio/guardrail/blob/v0.62.0/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/EndpointsClientGenerator.scala#L290-L294), so I'd err on the side of stability until we have the truly generic solution, if I'm being selfish.

The change in order to make this test pass is
```diff
diff --git a/json-schema/json-schema/src/main/scala/endpoints4s/Tupler.scala b/json-schema/json-schema/src/main/scala/endpoints4s/Tupler.scala
index 929d19e8..88bfcfb2 100644
--- a/json-schema/json-schema/src/main/scala/endpoints4s/Tupler.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/Tupler.scala
@@ -114,7 +114,23 @@ trait Tupler3 extends Tupler2 {
 
 }
 
-trait TuplerLeftUnit extends TuplerAppend {
+trait Tupler4 extends TuplerAppend {
+
+  implicit def tupler4And2[A, B, C, D, E, F]
+      : Tupler[(A, B, C, D), (E, F)] { type Out = (A, B, C, D, E, F) } =
+    new Tupler[(A, B, C, D), (E, F)] {
+      type Out = (A, B, C, D, E, F)
+      def apply(abcd: (A, B, C, D), ef: (E, F)): (A, B, C, D, E, F) =
+        (abcd._1, abcd._2, abcd._3, abcd._4, ef._1, ef._2)
+      def unapply(out: (A, B, C, D, E, F)): ((A, B, C, D), (E, F)) = {
+        val (a, b, c, d, e, f) = out
+        ((a, b, c, d), (e, f))
+      }
+    }
+
+}
+
+trait TuplerLeftUnit extends Tupler4 {
 
   implicit def leftUnit[A]: Aux[Unit, A, A] =
     new Tupler[Unit, A] {
diff --git a/json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala b/json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
index 67c00c4d..4f630166 100644
--- a/json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
+++ b/json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
@@ -20,7 +20,7 @@ class TuplerTests {
 
     tupling(a, (b, c, d)): (A, (B, C, D))
     tupling((a, b, c), (d, e)): (A, B, C, (D, E))
-    tupling((a, b, c, d), (e, f)): (A, B, C, D, (E, F))
+    tupling((a, b, c, d), (e, f)): (A, B, C, D, E, F)
     tupling((a, b, c), (d, e, f)): (A, B, C, (D, E, F))
     tupling((a, b), (c, d, e, f)): ((A, B), C, D, E, F)
     tupling(a, (b, c, d, e, f)): (A, (B, C, D, E, F))
```
but adding the various other combinations, `1And5`, `3And3`, `4And2`, made a bunch of other tests fail so I did not include it here.

blast-hardcheese:
@julienrf Anyway, ready for re-review. If you'd like more changes, I've got a better understanding of why the previous mechanics were not behaving as expected. I can also reorder the tests to be coherent when you feel comfortable with where this PR stands.

Thanks

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
julienrf: > Also, sorry, to what end are we doing this?
> 
> You asked about
> 
> ```scala
>     tupling((a, b, c, d), (e, f)): (A, B, C, D, (E, F))
> ```
> 
> but
> 
> ```scala
>     tupling((a, b, c), (d, e, f)): (A, B, C, (D, E, F))
>     tupling((a, b), (c, d, e, f)): ((A, B), C, D, E, F)
>     tupling(a, (b, c, d, e, f)): (A, (B, C, D, E, F))
> ```
> 
> would also not be covered by this. Isn't this just that we only have so many tupler extensions so far?

The goal of `Tupler` is to concatenate tuples, so that we end-up with a flat tuple instead of nested tuples (and occurrences of `Unit` are removed). The cases you mention should be fixed as well… The problem is that the `Tupler` instances currently cover only the most common cases, and being exhaustive would require a lot of implicit definitions (and headaches).

I suggest that we remove these lines from the test, as this is not something we want to support (I believe).

blast-hardcheese: Maybe split them out into a "stability" section? The reason for this PR in the first place is because an otherwise helpful addition ended up making some valid usecases completely unrepresentable.

At some point this breaks down, for instance a `Tuple22` combined with anything. I've needed to work around this for particularly complex OpenAPI specifications, which is what led me to `endpoints` in the first place.

Either way, your points about defining disambiguating `Tupler` instances in my own project are completely valid, just looking to wrap up this issue to help out others that may run into this same problem before the HList matter is settled.

I'll move the tests to a separate section that can act as a canary for when things change, but if you feel they should be removed I'm perfectly happy to do that as well.

Thanks again for your time here.

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
```diff
@@ -5,19 +5,56 @@ class TuplerTests {
...
     tupling((a, b, c, d), e): (A, B, C, D, E)
     tupling((a, b, c, d, e), f): (A, B, C, D, E, F)
-    tupling(a, ()): A
-    tupling((), a): A
-    tupling((), ()): Unit
+    tupling((a, b, c, d, e, f), g): (A, B, C, D, E, F, G)
+    tupling((a, b, c, d, e, f, g), h): (A, B, C, D, E, F, G, H)
+    tupling((a, b, c, d, e, f, g, h), i): (A, B, C, D, E, F, G, H, I)
+
+    // Stability tests
```
julienrf, line 30: Not sure what you mean by “stability” test, do you mind elaborating?

blast-hardcheese: "Stability of implementation", whether a proposed change alters the way things stand today. Any alterations to these "out of spec" tests would be reflected in a change to these lines, signaling to both maintainers and users of this library when the semantics change.

julienrf (approved):

julienrf:
Thanks for your help and patience @blast-hardcheese!

PR #716, opened 2021-01-04
Build sttp module for JS aswell

Fristi:


Fristi:
Not needed, as we have more optimized implementations

PR #720, opened 2021-01-08
Provide `JsonSchema[Set[?]]` instance out of the box

### LINKED ISSUE #682: genericJsonSchema doesn't support Set
The following code compiles:

```scala
final case class A (
  a: String,
  b: Seq[Int]
)

final implicit lazy val aSchema: JsonSchema[A] = genericJsonSchema
```

The following doesn't:

```scala
final case class B (
  a: String,
  b: Set[Int]
)

final implicit lazy val bSchema: JsonSchema[B] = genericJsonSchema
```

```
Unable to derive an instance of JsonSchema[B].
The type B must be a sealed trait or a case class.
f it is a sealed trait, make sure it is only extended by case classes and that a JsonSchema can be derived for each of these case classes (see hereafter).
If it is a case class, make sure each field of the case class has an implicit JsonSchema.
You can check that a JsonSchema is available for a type T by compiling the expression `implicitly[JsonSchema[T]]`.
```

julienrf:
We achieve this by relaxing the constraint on the type signature of the existing method `arrayJsonSchema`. It now applies to any `Iterable` collection instead of the more specific type `Seq`.

Fixes #682

### DIFF
--- a/json-schema/json-schema-circe/src/main/scala/endpoints4s/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints4s/circe/JsonSchemas.scala
@@ -348,7 +348,7 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
   implicit def byteJsonSchema: JsonSchema[Byte] =
     JsonSchema(implicitly, implicitly)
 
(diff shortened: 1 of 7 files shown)


julienrf:
cc @marconilanna

PR #724, opened 2021-01-18
Add JsonEntitiesFromSchemas to Play client

### LINKED ISSUE #723: Generate Play client from Schema
The Akka HTTP client module supports both `JsonEntitiesFromCodecs` and `JsonEntitiesFromSchemas`. However, the Play client module only supports `JsonEntitiesFromCodecs`. A schema is required to generate OpenAPI documentation. So it is not possible to, at the same time, generate docs and a Play client?

marconilanna:
Fixes #723


julienrf (commented):
Nice, thanks @marconilanna! I wonder if there are some tests that we can apply to this interpreter?

julienrf (approved):

julienrf:
Thanks a lot @marconilanna! I’ve pushed a commit to fix the compilation error in the tests (sorry, the architecture of the tests is a mess…, we should fix that, eventually)

PR #728, opened 2021-01-22
Add a method that returns an `Uri` to the class `Endpoint` in the akka-http-server interpreter

julienrf:
It is useful to produce links, for instance to reference endpoints from other endpoints (e.g., in a HATEOAS API). This was supported from the beginning by the play-server interpreter. We now also support it in the akka-http-server interpreter.


julienrf:
@LeifW, maybe you would be interested in reviewing this?

PR #742, opened 2021-02-12
Add support for trailing slash in urls

agboom:
- Add path test with trailing slash
- Implement trailing slash support in Akka HTTP, Play and Http4s interpreters

### DIFF
--- a/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Urls.scala
+++ b/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Urls.scala
@@ -61,9 +61,16 @@ trait Urls extends algebra.Urls with StatusCodes {
     ): Option[Validated[A]]
 
     final def directive: Directive1[A] = {
-      (Directives.path(Directives.Segments) & Directives.parameterMultiMap)
-        .tflatMap { case (segments, query) =>
-          validateUrl(segments, query) match {
+      (Directives.extractUri & Directives.path(
+        Directives.Segments ~ Directives.Slash.?
+      ) & Directives.parameterMultiMap)
+        .tflatMap { case (uri, segments, query) =>
+          val segmentsLeadingTrailingSlash = "" :: segments ++ {
+            if (uri.path.endsWithSlash && uri.path != Uri.Path.SingleSlash)
+              List("")
+            else Nil
+          }
+          validateUrl(segmentsLeadingTrailingSlash, query) match {
             case None               => Directives.reject
             case Some(inv: Invalid) => handleClientErrors(inv)
             case Some(Valid(a))     => Directives.provide(a)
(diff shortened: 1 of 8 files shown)


agboom:
The CI checks fail because of a spurious error in the shutdown of the connection pool. I couldn't find a way to restart the CI build, but on my machine the tests succeed.

julienrf (commented):
At first I thought that the syntax `path / "foo" / ""` does not look super nice (I’m not a big fan of empty strings), but I must say that the diff is quite simple and I find it cool that we can support trailing slashes in URLs with so few changes!

I think it would be nice to also have client tests, to check that client interpreters do produce correct URIs. Would you be interested in adding that?

agboom:
> Thank you @agboom! At first I thought that the syntax `path / "foo" / ""` does not look super nice (I’m not a big fan of empty strings), but I must say that the diff is quite simple and I find it cool that we can support trailing slashes in URLs with so few changes!

Thanks, I had the same train of thought! I was thinking we could also add an alternative method to allow for a nicer syntax for trailing slashes, something like: `path / "foo"./` or even `trailingSlash(path / "foo")`. On the other hand, the current syntax is not that bad and at least it's consistent.

> I think it would be nice to also have client tests, to check that client interpreters do produce correct URIs. Would you be interested in adding that?
That's a good point, I'll look into it.

agboom:
@julienrf I added a client test case. It revealed a problem in the scalaj client so it's a good thing you called it :)
I had to change the way the url string is constructed in scalaj-client to make the test pass. Could you see if the change is sensible?

I also added an alternative syntax for trailing slash, which allows for a way to create a trailing slash without using the empty string:
```scala
path / "foo"./
```
or:
```scala
path / segment[String]./
```
The syntax is a bit less obvious, but I do think it looks nice and intuitive. WDYT?

I also added some more test cases for the server test suite, in segments as well as query strings. Should I do the same for the client?

agboom:
While trying some things, I discovered a discrepancy between server implementations regarding matching double slashes in urls.
The following uncommon but possible situtation:
```
      decodeUrl(path / "" / "foo")("//foo") shouldEqual Matched(())
```
succeeds for play-server, but fails for akka-http-server and http4s-server. For akka-http-server I'm almost certain because the pathmatcher rejects it. I'm yet to find the relevant documentation for this. I don't have any experience with http4s

[RFC2396](https://www.ietf.org/rfc/rfc2396.txt) allows for double slashes in urls, but it just means that the segment is empty. It could be that some servers implement this differently to be more liberal, but I could be wrong.

julienrf:
> I had to change the way the url string is constructed in scalaj-client to make the test pass. Could you see if the change is sensible?

Thank you so much for fixing it!

> I also added an alternative syntax for trailing slash

I am not sure yet… :slightly_smiling_face: I like the fact that the `path / "foo" / ""` syntax did not introduce new methods in the algebra, and I think it is self-explanatory.

About double slashes: that’s interesting, although I’m not sure what a good use case for them is?

julienrf (approved):
Thank you again @agboom for your work! I suggest not introducing the special syntax for trailing slashes now. We can do it later if we feel a real need…

I have one last request: do you mind updating the documentation about the URL algebra to include an example with a trailing slash?

agboom:
> Thank you again @agboom for your work! I suggest not introducing the special syntax for trailing slashes now. We can do it later if we feel a real need…

Sure, that sounds better, I agree the basic syntax is fine for now.

> I have one last request: do you mind updating the documentation about the URL algebra to include an example with a trailing slash?

Sure, no problem!

agboom:
@julienrf I've rebased the commits, removing the syntax alternative (but keeping some tests from that commit). I've also added a single example of a trailing slash to Endpoints#url in the docs.

agboom:
Should I squash the commits in this PR or is it fine the way it is?

julienrf:
@agboom Thank you! I prefer if you squash the commits, indeed.

agboom:
> @agboom Thank you! I prefer if you squash the commits, indeed.

Done

agboom:
Thanks for merging :smile:

PR #830, opened 2021-05-28
Implement endpoints4s.generic.JsonSchemas for Scala 3

julienrf:


julienrf:
I’m not super happy with this implementation (I counted 24 calls to `asInstanceOf` :disappointed: ) but at least all the existing tests pass!

PR #833, opened 2021-06-05
Add JsonSchema.lazySchema

mleclercq:
Allow more general recursive schemas than `lazyRecord` or `lazyTagged`. For instance schema for an expression language like:

```scala
sealed trait Expression
object Expression {
  case class Literal(value: Int) extends Expression
  case class Add(x: Expression, y: Expression) extends Expression
}
val expressionSchema: JsonSchema[Expression] = {
  implicit val self: JsonSchema[Expression] = lazySchema(expressionSchema, "Expression")
  intJsonSchema
    .orFallbackTo(field[Expression]("x") zip field[Expression]("y"))
    .xmap[Expression] {
      case Left(value)   => Expression.Literal(value)
      case Right((x, y)) => Expression.Add(x, y)
    } {
      case Expression.Literal(value) => Left(value)
      case Expression.Add(x, y)      => Right((x, y))
    }
}
```

Which supports JSON like:

```json
{"x": 1, "y": {"x": 2, "y": 3}}
```

Maybe `JsonSchema.lazyRecord` and `JsonSchema.lazyRecord` could be deprecated since they are just redefinitions of `JsonSchema.lazySchema` ?

### DIFF
--- a/json-schema/json-schema-circe/src/main/scala/endpoints4s/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints4s/circe/JsonSchemas.scala
@@ -166,8 +166,9 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
     )
   }
 
-  private def lazySchema[A](
-      schema: => JsonSchema[A]
+  def lazySchema[A](
+      schema: => JsonSchema[A],
+      name: String
   ): JsonSchema[A] = {
     // The schema won’t be evaluated until its `encoder` or `decoder` is effectively used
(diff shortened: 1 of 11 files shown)


mleclercq:
I realized that a different approach to tackle recursive schemas based on some sort of a Y-combinator is also possible (see prototype at https://github.com/mleclercq/endpoints/tree/Y-jsonschema).

That approach has the added benefit of avoiding the double definitions of the types in the open-api schema (like in https://github.com/endpoints4s/endpoints4s/blob/master/openapi/openapi/src/test/scala/endpoints4s/openapi/EndpointsTest.scala#L110-L144).

I'm closing that PR for now and will open a ticket.

Issue #834, opened 2021-06-06
Support any kind of recursive JSON schema


mleclercq:
Existing support for recursive JSON schema only supports object schemas (with or without a discriminator field). It seems that it is not possible to write a JSON schema for an ADT like the following where `Expression.Literal` is represented as a JSON number and `Expression.Add` is represented as a JSON object:

```scala
sealed trait Expression
object Expression {
  case class Literal(value: Int) extends Expression
  case class Add(x: Expression, y: Expression) extends Expression
}
```

An example of a value would be:

```json
{"x": 1, "y": {"x": 2, "y": 3}}
```

That may sounds like a contrived example, but I do have an existing REST API that I would like to convert to endpoints4s that have constructs similar to this...

I prototyped two different approaches to tackle this:

1. generalize `lazyRecord` and `lazyTagged` with a new `lazySchema` method: 
   ```scala
   def lazySchema[A](schema: => JsonSchema[A], name: String): JsonSchema[A]
   ```
   It just so happen that both `lazyRecord` and `lazyTagged` can be implemented in term of `lazySchema` at very low cost (see https://github.com/mleclercq/endpoints/commit/c76f52a06726355cc3f89648d9e509002f964582)
2. add a new `recursiveSchema` method that works like a Y-combinator:
   ```scala
   def recursiveSchema[A](name: String)(f: JsonSchema[A] => JsonSchema[A]): JsonSchema[A]
   ```
   This approach requires a little bit more work especially in the openApi interpreter (see https://github.com/mleclercq/endpoints/commit/cc06cea55eac2db092e376fa6bd123b0607e8299)

With the first approach, the JSON schema for the `Expression` type above can be written as:

```scala
implicit val expressionSchema: JsonSchema[Expression] = {
  val self: JsonSchema[Expression] = lazySchema(expressionSchema, "Expression")
  intJsonSchema
    .orFallbackTo(field("x")(self) zip field("y")(self))
    .xmap[Expression] {
      case Left(value)   => Expression.Literal(value)
      case Right((x, y)) => Expression.Add(x, y)
    } {
      case Expression.Literal(value) => Left(value)
      case Expression.Add(x, y)      => Right((x, y))
    }
}
```

With the second approach, the JSON schema can be written as:

```scala
implicit val expressionSchema =
  recursiveSchema[Expression]("Expression") { implicit self =>
    intJsonSchema
      .orFallbackTo(field[Expression]("x") zip field[Expression]("y"))
[...]

PR #839, opened 2021-06-13
Add JsonSchema.lazySchema.

mleclercq:
Allow more general recursive schemas than `lazyRecord` or `lazyTagged`.
For instance schema for an expression language like:

```json
  {"op": "add", "x": 1, "y": {"op": "add", "x": 2, "y": 3}}
```

See discussion at https://github.com/endpoints4s/endpoints4s/issues/834

### DIFF
--- a/json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
+++ b/json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
@@ -107,6 +107,31 @@ class JsonSchemasTest extends AnyFreeSpec {
     )
     assert(JsonSchemasCodec.recursiveSchema.encoder(rec) == json)
   }
+  "recursive expression type" in {
+    val json = Json.obj(
+      "x" -> Json.obj("x" -> Json.fromInt(1), "y" -> Json.fromInt(2)),
+      "y" -> Json.fromInt(3)
+    )
+    val expr = JsonSchemasCodec.Expression.Add(
+      JsonSchemasCodec.Expression
+        .Add(JsonSchemasCodec.Expression.Literal(1), JsonSchemasCodec.Expression.Literal(2)),
+      JsonSchemasCodec.Expression.Literal(3)
+    )
+    assert(
+      JsonSchemasCodec.expressionSchema.decoder.decodeJson(json).exists(_ == expr)
+    )
+    assert(JsonSchemasCodec.expressionSchema.encoder(expr) == json)
+  }
+  "mutually recursive types" in {
+    val json = Json.obj("b" -> Json.obj("a" -> Json.obj()))
+    val rec = JsonSchemasCodec.MutualRecursiveA(
+      Some(JsonSchemasCodec.MutualRecursiveB(Some(JsonSchemasCodec.MutualRecursiveA(None))))
+    )
+    assert(
+      JsonSchemasCodec.mutualRecursiveA.decoder.decodeJson(json).exists(_ == rec)
+    )
+    assert(JsonSchemasCodec.mutualRecursiveA.encoder(rec) == json)
+  }
 
   "tuple" in {
     val json = Json.arr(Json.True, Json.fromInt(42), Json.fromString("foo"))
--- a/json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
+++ b/json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
@@ -141,6 +141,29 @@ class JsonSchemasTest extends AnyFreeSpec {
     )
     assert(JsonSchemasCodec.mutualRecursiveB.encoder(recB) == jsonB)
   }
+  "recursive tagged" in {
+    val json = Json.obj(
+      "kind" -> Json.fromString("A"),
+      "a" -> Json.fromString("foo"),
+      "next" -> Json.obj(
+        "kind" -> Json.fromString("B"),
+        "b" -> Json.fromInt(42)
+      )
+    )
+    val rec = JsonSchemasCodec.TaggedRecursiveA(
+      a = "foo",
+      next = Some(
+        JsonSchemasCodec.TaggedRecursiveB(
+          b = 42,
+          next = None
+        )
+      )
+    )
+    assert(
+      JsonSchemasCodec.taggedRecursiveSchema.decoder.decodeJson(json).exists(_ == rec)
+    )
+    assert(JsonSchemasCodec.taggedRecursiveSchema.encoder(rec) == json)
+  }
 
   "tuple" in {
     val json = Json.arr(Json.True, Json.fromInt(42), Json.fromString("foo"))
--- a/json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
+++ b/json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
@@ -123,14 +123,23 @@ class JsonSchemasTest extends AnyFreeSpec {
     assert(JsonSchemasCodec.expressionSchema.encoder(expr) == json)
   }
   "mutually recursive types" in {
-    val json = Json.obj("b" -> Json.obj("a" -> Json.obj()))
-    val rec = JsonSchemasCodec.MutualRecursiveA(
+    val jsonA = Json.obj("b" -> Json.obj("a" -> Json.obj()))
+    val recA = JsonSchemasCodec.MutualRecursiveA(
       Some(JsonSchemasCodec.MutualRecursiveB(Some(JsonSchemasCodec.MutualRecursiveA(None))))
     )
     assert(
-      JsonSchemasCodec.mutualRecursiveA.decoder.decodeJson(json).exists(_ == rec)
+      JsonSchemasCodec.mutualRecursiveA.decoder.decodeJson(jsonA).exists(_ == recA)
     )
-    assert(JsonSchemasCodec.mutualRecursiveA.encoder(rec) == json)
+    assert(JsonSchemasCodec.mutualRecursiveA.encoder(recA) == jsonA)
+
+    val jsonB = Json.obj("a" -> Json.obj("b" -> Json.obj()))
+    val recB = JsonSchemasCodec.MutualRecursiveB(
+      Some(JsonSchemasCodec.MutualRecursiveA(Some(JsonSchemasCodec.MutualRecursiveB(None))))
+    )
+    assert(
+      JsonSchemasCodec.mutualRecursiveB.decoder.decodeJson(jsonB).exists(_ == recB)
+    )
+    assert(JsonSchemasCodec.mutualRecursiveB.encoder(recB) == jsonB)
   }
 
   "tuple" in {
--- a/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
@@ -225,6 +225,7 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     * @param name A unique name identifying the schema
     * @group operations
     */
+  @deprecated("Use `lazySchema` instead", "1.4.0")
   def lazyRecord[A](schema: => Record[A], name: String): JsonSchema[A]
 
   /** Captures a lazy reference to a JSON schema currently being defined.
@@ -236,6 +237,7 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     * @param name A unique name identifying the schema
     * @group operations
     */
+  @deprecated("Use `lazySchema` instead", "1.4.0")
   def lazyTagged[A](schema: => Tagged[A], name: String): JsonSchema[A]
 
   /** A lazy JSON schema that can references schemas currently being defined:
@@ -254,7 +256,8 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     * @param schema The record JSON schema whose evaluation should be delayed
     * @group operations
     */
-  def lazySchema[A](name: String)(schema: => JsonSchema[A]): JsonSchema[A]
+  def lazySchema[A](name: String)(schema: => JsonSchema[A]): JsonSchema[A] =
+    sys.error(s"Unsupported algebra version: 1.4.0. Please update your interpreter.")
 
   /** The JSON schema of a record with no fields
     *
--- a/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
@@ -238,6 +238,24 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     */
   def lazyTagged[A](schema: => Tagged[A], name: String): JsonSchema[A]
 
+  /** A lazy JSON schema that can references schemas currently being defined:
+    *
+    * {{{
+    *   case class Recursive(next: Option[Recursive])
+    *   val recursiveSchema: JsonSchema[Recursive] = lazySchema("Rec")(
+    *     optField("next")(recursiveSchema)
+    *   ).xmap(Recursive)(_.next)
+    * }}}
+    *
+    * Interpreters should return a JsonSchema value that does not evaluate
+    * the given `schema` unless it is effectively used.
+    *
+    * @param name A unique name identifying the schema
+    * @param schema The record JSON schema whose evaluation should be delayed
+    * @group operations
+    */
+  def lazySchema[A](name: String)(schema: => JsonSchema[A]): JsonSchema[A]
+
   /** The JSON schema of a record with no fields
     *
     *   - Encoder interpreters produce an empty JSON object,
--- a/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
@@ -225,7 +225,7 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     * @param name A unique name identifying the schema
     * @group operations
     */
-  @deprecated("Use `lazySchema` instead", "1.4.0")
+  @deprecated("Use `lazyRecord(name)(...)` instead", "1.4.0")
   def lazyRecord[A](schema: => Record[A], name: String): JsonSchema[A]
 
   /** Captures a lazy reference to a JSON schema currently being defined.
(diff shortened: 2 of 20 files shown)


json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
```diff
@@ -238,6 +238,24 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
...
+    * }}}
+    *
+    * Interpreters should return a JsonSchema value that does not evaluate
+    * the given `schema` unless it is effectively used.
+    *
+    * @param name A unique name identifying the schema
+    * @param schema The record JSON schema whose evaluation should be delayed
+    * @group operations
+    */
+  def lazySchema[A](name: String)(schema: => JsonSchema[A]): JsonSchema[A]
```
julienrf, line 257: In order to stay backward binary compatible, we need to provide a default implementation (which throws an error):

```suggestion
  def lazySchema[A](name: String)(schema: => JsonSchema[A]): JsonSchema[A] =
    sys.error(s"Unsupported algebra version: 1.4.0. Please update your interpreter.")
```

json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
```diff
@@ -107,6 +107,31 @@ class JsonSchemasTest extends AnyFreeSpec {
...
+  }
+  "mutually recursive types" in {
+    val json = Json.obj("b" -> Json.obj("a" -> Json.obj()))
+    val rec = JsonSchemasCodec.MutualRecursiveA(
+      Some(JsonSchemasCodec.MutualRecursiveB(Some(JsonSchemasCodec.MutualRecursiveA(None))))
+    )
+    assert(
+      JsonSchemasCodec.mutualRecursiveA.decoder.decodeJson(json).exists(_ == rec)
+    )
+    assert(JsonSchemasCodec.mutualRecursiveA.encoder(rec) == json)
```
julienrf, line 133: Could we also test `mutualRecursiveB`? (just to double check that there is no initialization issue.

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -130,6 +130,23 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
         }
     }
 
+    sealed abstract class RecursiveSchema extends DocumentedJsonSchema {
```
julienrf, line 133: Would it be possible to reuse `LazySchema` instead? (by adding a `def name: String` to it)

openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
```diff
@@ -130,6 +130,11 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
       val decoder = json => schema.decoder.decode(json)
       val encoder = value => schema.encoder.encode(value)
     }
+  def lazySchema[A](name: String)(schema: => JsonSchema[A]): JsonSchema[A] =
+    new JsonSchema[A] {
+      val decoder = json => schema.decoder.decode(json)
+      val encoder = value => schema.encoder.encode(value)
+    }
```
julienrf, line 137: I guess `lazyRecord` and `lazyTagged` could be implemented by calling `lazySchema`?

json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
julienrf: Also, are there any benefits in keeping `lazyRecord` and `lazyTagged`? Should we just deprecate them in favor of `lazySchema`? If we keep them, would it make sense to add two overloads with signatures looking similar to `lazySchema`?

openapi/openapi/src/test/scala/endpoints4s/openapi/EndpointsTest.scala
```diff
@@ -143,6 +143,83 @@ class EndpointsTest extends AnyWordSpec with Matchers with OptionValues {
...
+            None,
+            None
+          )
+        ),
+        None
+      )
+    Fixtures.toSchema(Fixtures.expressionSchema.docs) shouldBe expectedSchema
+  }
+  "Mutually recursive types" in {
+    Fixtures.toSchema(Fixtures.mutualRecursiveA.docs) shouldBe Schema.Reference(
```
julienrf, line 182: What does `mutualRecursiveB.docs` look like?

julienrf (commented):
Thank you @mleclercq! I have a few comments.

openapi/openapi/src/test/scala/endpoints4s/openapi/EndpointsTest.scala
mleclercq: Here are the schema we got for `Expression` and the `MutualRecursiveA/B`:

```json
{
  "schemas": {
    "Expression": {
      "oneOf": [
        {
          "type": "integer",
          "format": "int32"
        },
        {
          "type": "object",
          "properties": {
            "x": {
              "$ref": "#/components/schemas/Expression"
            },
            "y": {
              "$ref": "#/components/schemas/Expression"
            }
          },
          "required": [
            "x",
            "y"
          ]
        }
      ]
    },
    "A": {
      "type": "object",
      "properties": {
        "b": {
          "$ref": "#/components/schemas/B"
        }
      }
    },
    "B": {
      "type": "object",
      "properties": {
        "a": {
          "$ref": "#/components/schemas/A"
        }
      }
    }
  }
}
```

openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
mleclercq: Done

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
mleclercq: Done

json-schema/json-schema-circe/src/test/scala/endpoints4s/circe/JsonSchemasTest.scala
mleclercq: Done

json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
mleclercq: I marked `lazyRecord` and `lazyTagged` as deprecated. The migration is indeed pretty straight forward.

json-schema/json-schema/src/test/scala/endpoints4s/algebra/JsonSchemasDocs.scala
```diff
@@ -66,7 +66,7 @@ trait JsonSchemasDocs extends JsonSchemas {
   case class Recursive(next: Option[Recursive])
 
   val recursiveSchema: Record[Recursive] = (
-    optField("next")(lazyRecord(recursiveSchema, "Rec"))
+    optField("next")(lazySchema("Rec")(recursiveSchema))
   ).xmap(Recursive(_))(_.next)
```
julienrf, line 70: It would be nice to be able to write this:

~~~ scala
val recursiveSchema: Record[Recursive] =
  lazyRecord("Rec") {
    optField("next")(recursiveSchema)
  }.xmap(Recursive(_))(_.next)
~~~

But that does not work currently because `lazyRecord` returns a `JsonSchema` instead of a `Record`. Would it be possible to make it return a `Record` instead?

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -116,13 +116,19 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
 
     // A documented JSON schema that is unevaluated unless its `value` is accessed
     sealed abstract class LazySchema extends DocumentedJsonSchema {
+      def name: String =
+        sys.error(s"Unsupported algebra version: 2.1.0. Please update your interpreter.")
```
julienrf, line 120: In my opinion, it’s fine to leave it abstract and to bump the major version number.

It’s important to keep binary compatibility as long as possible in the algebra modules, because in a microservice architecture you may deal with APIs of multiple (independent) services, each using a different version of the algebra of endpoints4s, so we want them to be all binary compatible. However, dependencies on interpreters (such as openapi) are typically direct dependencies (as opposed to transitive dependencies), so it is less important to keep them binary compatible.

openapi/openapi/src/test/scala/endpoints4s/openapi/EndpointsTest.scala
```diff
@@ -309,6 +386,9 @@ class EndpointsTest extends AnyWordSpec with Matchers with OptionValues {
     }
   }
 
+  "Foo" in {
+    println(OpenApi.stringEncoder.encode(MyFixture.api))
+  }
```
julienrf, line 391: This looks like a leftover :blush:

openapi/openapi/src/test/scala/endpoints4s/openapi/EndpointsTest.scala
```diff
@@ -422,3 +502,51 @@ object Fixtures
     )
 
 }
+
+trait MyFixture extends algebra.Endpoints with algebra.JsonEntitiesFromSchemas {
```
julienrf, line 506: Is this used?

mleclercq: Oupsy, That was just a quick test I did that wasn't intended to be checked-in. I'll clear that out...

json-schema/json-schema/src/test/scala/endpoints4s/algebra/JsonSchemasDocs.scala
mleclercq: I'm not sure I follow you. `lazySchema` (not `lazyRecord`) cannot return a `Record` because it doesn't always encode to a Json object. For instance with the `Expression` example, a simple literal value is encoded as a Json number.

As for `lazyRecord` and `lazyTagged`, they are deprecated. Are you suggesting that we add new versions of these of the following form?

```scala
def lazyRecord(name: String)(schema: => Record): Record
def lazyTagged(name: String)(schema: => Tagged): Tagged
```

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
mleclercq: Sounds, good. To be honest, I'm not very used to these binary compatibility considerations, I just try to get the github action happy 😊

json-schema/json-schema/src/test/scala/endpoints4s/algebra/JsonSchemasDocs.scala
julienrf: Yes, I realize now that we want a way to eventually return a `Record` or `Tagged`, like we could do in the `recursiveSchema` example. What do you think?

mleclercq:
@julienrf I'm a bit confused by the versioning checks.

Here is what I did in the latest commit: I bumped the minor version in the top-level `build.sbt`:

https://github.com/endpoints4s/endpoints4s/blob/8aef72fe226fbcbd9d0003a8faf24334277a8fc2/build.sbt#L42

And bumped the major version in `openapi/build.sbt`:

https://github.com/endpoints4s/endpoints4s/blob/8aef72fe226fbcbd9d0003a8faf24334277a8fc2/openapi/build.sbt#L10

But now there are new errors in the versioning checks of all the interpreters that depends on openapi. Should I bump their major versions too ?

json-schema/json-schema/src/test/scala/endpoints4s/algebra/JsonSchemasDocs.scala
mleclercq: Ok I've added these two methods. It was pretty easy in the `JsonSchemas` interpreters except for openapi. 

The problem there is that a `Record` is a class that expects a `DocumentedRecord` value in its constructor so the following doesn't work:

```scala
  override def lazyRecord[A](name: String)(schema: => Record[A]): Record[A] =
    new Record[A](
      ujsonSchemas.lazyRecord(name)(schema.ujsonSchema),
      LazySchema(name, schema.docs) // Wrong need a DocumentedRecord here
    )
```

So I made `DocumentedRecord` a `sealed abstract class` with two sub-classes. One is a simple `case class`, the other is a `Lazy` version. We also have to be careful about the other operations in the algebra that manipulate a `Record`.

For instance:

```scala
val recursiveSchema: Record[Recursive] =
  lazyRecord("Rec") {
    optField("next")(recursiveSchema)
  }
    .xmap(Recursive(_))(_.next)
    // We have to be careful that this doesn't result in a NPE
    .withTitle("Recursive is hard")
```

julienrf:
@mleclercq sorry this is a bit complicated because this project does not follow the usual convention (where all the modules of the repository usually have the same version). I’ve added a commit stating that we intend to break binary compatibility in the openapi module.

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -58,14 +58,74 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
+        def fields: List[Field] = evaluatedDocs.fields
+        def additionalProperties: Option[DocumentedJsonSchema] =
+          evaluatedDocs.additionalProperties
+        def name: Option[String] = Some(n)
+        def description: Option[String] = evaluatedDocs.description
+        def example: Option[ujson.Value] = evaluatedDocs.example
+        def title: Option[String] = evaluatedDocs.title
+
+        def withName(n: String): DocumentedRecord =
+          new Lazy(n, this)
```
julienrf, line 102: Should it be the following, just to avoid one more level of indirection?

```suggestion
          new Lazy(n, docs)
```

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -74,14 +134,78 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
+        def withTitle(t: String): DocumentedCoProd =
+          new Lazy(n, this) {
+            override def title: Option[String] = Some(t)
+          }
+        def withDescription(d: String): DocumentedCoProd =
+          new Lazy(n, this) {
+            override def description: Option[String] = Some(d)
+          }
+        def withDiscriminatorName(d: String): DocumentedCoProd =
+          new Lazy(n, null) {
```
julienrf, line 194: Why `null` here?

openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
```diff
@@ -29,7 +29,8 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
 
   }
 
-  abstract class Tagged[A](val discriminator: String) extends Record[A] {
+  trait Tagged[A] extends Record[A] {
+    def discriminator: String
```
julienrf, line 33: Was there a reason for changing the parameter into an abstract method? (I’m fine with this change, which I believe is equivalent to the previous state, I’m just curious if there is something that _required_ the change or not)

json-schema/json-schema/src/test/scala/endpoints4s/algebra/JsonSchemasDocs.scala
julienrf: What do you mean by “we have to be careful that this doesn’t result in a NPE”? Do users of endpoints4s have to be careful?

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -325,15 +457,15 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
       djs match {
-        case s: DocumentedRecord => s.copy(example = Some(exampleJson))
-        case s: DocumentedCoProd => s.copy(example = Some(exampleJson))
+        case s: DocumentedRecord => s.withExample(exampleJson)
+        case s: DocumentedCoProd => s.withExample(exampleJson)
         case s: Primitive        => s.copy(example = Some(exampleJson))
         case s: Array            => s.copy(example = Some(exampleJson))
         case s: DocumentedEnum   => s.copy(example = Some(exampleJson))
-        case s: LazySchema       => LazySchema(updatedDocs(s.value))
+        case s: LazySchema       => LazySchema(s.name, updatedDocs(s.value))
```
harpocrates, line 468: Why should `LazySchema` be handled as a case in this match, but `DocumentedRecord.Lazy` and `DocumentedCoProd.Lazy` are not? What is the fundamental conceptual difference? It would be really nice if there were some unified handling of lazy schemas.

openapi/openapi/src/test/scala/endpoints4s/openapi/EndpointsTest.scala
harpocrates: This is very nice!

openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
mleclercq: This is because in `lazyTagged` we return a `Tagged` instance as follow:

```scala
override def lazyTagged[A](name: String)(schema: => Tagged[A]): Tagged[A] = {
  lazy val evaluatedSchema = schema
  new Tagged[A] {
    // Here we need to keep the discriminator lazy
    lazy val discriminator = evaluatedSchema.discriminator
    def findDecoder(tag: String): Option[Decoder[ujson.Value, A]] =
      evaluatedSchema.findDecoder(tag)
    def tagAndObj(a: A): (String, ujson.Obj) =
      evaluatedSchema.tagAndObj(a)
  }
}
```

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
mleclercq: Oups, don't know how that one got there 😊

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
mleclercq: Indeed. For `withName` we can avoid that indirection.

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
mleclercq: `DocumentedRecord.Lazy` and `DocumentedCoProd.Lazy` are handled by the case on `DocumentedRecord` and `DocumentedCoProd`. They cannot by handled the same way as `LazySchema` because `updatedDocs(s.value)` returns a `DocumentedJsonSchema` while `DocumentedRecord.Lazy` requires a `DocumentedRecord` so a case like the following doesn't work:

```
  case s: DocumentedRecord.Lazy => DocumentedRecord.Lazy(s.name, updatedDocs(s.value))
```

This is why I added the `withName`, `withTitle` etc. on `DocumentedRecord` and `DocumentedTagged`. I guess I could also add these methods on `LazySchema`, or even on `DocumentedJsonSchema` so we wouldn't even need these pattern matchings.

WDYT?

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -74,14 +134,78 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
+        def name: Option[String] = Some(n)
+        def discriminatorName: String = evaluatedDocs.discriminatorName
+        def description: Option[String] = evaluatedDocs.description
+        def example: Option[ujson.Value] = evaluatedDocs.example
+        def title: Option[String] = evaluatedDocs.title
+
+        def withName(n: String): DocumentedCoProd =
+          new Lazy(n, docs)
+        def withExample(e: => ujson.Value): DocumentedCoProd =
+          new Lazy(n, this) {
```
julienrf, line 182: I think we should do the same everywhere, right?
```suggestion
          new Lazy(n, docs) {
```

mleclercq: I don't think so because if we chain multiple `.withExample(...).withTitle(...)` the instance returned by `withTitle` would not return the example set by `withExample`.

json-schema/json-schema/src/test/scala/endpoints4s/algebra/JsonSchemasDocs.scala
julienrf: Could you please clarify the above question?

mleclercq: Ho sorry, I meant that we have to be careful in the implementation of the interpreter that it doesn't raise an NPE. This is not a concern for the end user.

In the tests, I added a couple of the `withXYZ` on the recursive schemas to make sure they work properly when they are used "insie" `lazyRecord` or on the result of it:

```scala
  val taggedRecursiveSchema: Tagged[TaggedRecursive] =
    lazyTagged("TaggedRec") {
      val aSchema = ...
      val bSchema = ...

      (aSchema.tagged("A") orElseMerge bSchema.tagged("B"))
        // Put some doc inside lazyTagged, and some more outside to make sure it works both way
        .withDiscriminator("kind")
        .withDescription("TaggedRec description")
    }
      .withTitle("TaggedRec title")
      .withExample(TaggedRecursiveA("foo", None))
```

julienrf: Awesome, thank you for the explanation!

julienrf (approved):

julienrf:
Thank you @mleclercq !

PR #842, opened 2021-06-18
Add sttp3 interpreter

Fristi:


julienrf:
Thank you! Just a quick question: is there a strong need for keeping both sttp2 and sttp3 interpreters? Could we just update the sttp2 interpreter to use sttp3?

Fristi:
We'll switch, but not sure if other users are still stuck with sttp 2.x. Your call :)

julienrf:
I can see in the stats that some people use the sttp interpreter (like 20% of the endpoints4s users). This is significant, but on the other hand, it’s really hard for me to maintain all these interpreters in the main repository. Ideally, I would rather have some of them taken over in separate repositories.

So, I propose that we upgrade the sttp2 interpreter to sttp3, and if needed we can resurrect the old one outside of the main endpoints4s repository. What do you think?

Fristi:
Updated PR, now only minimal changes required. If all is good, would you be able to include in a release later this week? :)

julienrf:
Yes, let’s do a release when this PR and #839 are merged.

sttp/client/src/main/scala/endpoints4s/sttp/package.scala
```diff
@@ -1,4 +0,0 @@
-package endpoints4s
-
-/** Client interpreter using Sttp */
-package object sttp
```
julienrf, line 4: Would you mind adding back this file, I think it’s nice to have this documentation when we browse the packages: https://endpoints4s.github.io/api/endpoints4s/sttp/index.html

sttp/build.sbt
```diff
@@ -10,11 +10,11 @@ val `sttp-client` =
       publishSettings,
       `scala 2.12 to 2.13`,
       name := "sttp-client",
-      version := "3.0.0+n",
+      version := "3.0.0",
       versionPolicyIntention := Compatibility.BinaryAndSourceCompatible,
```
julienrf, line 14: We need to bump the version to 4.0.0, and to declare that we will break the binary compatibility with 3.0.0

```suggestion
      version := "3.0.0+n",
      versionPolicyIntention := Compatibility.None,
```

Fristi:
I've processed your comments! 😃 

> So, I propose that we upgrade the sttp2 interpreter to sttp3, and if needed we can resurrect the old one outside of the main endpoints4s repository. What do you think?

Sorry for the delayed reply, I think that's a good idea moving forward

julienrf (approved):

PR #871, opened 2021-07-31
enable format for url parameters in openapi

### LINKED ISSUE #854: Type format hint is not supported for OpenApi parameters
`format` hint is currently supported for request and response body content, but not for parameters.  

For example, this definition
```scala
  val foo: Endpoint[(Long, Long, Long), Long] =
    endpoint(
      request = put(
        path / "foo" / segment[Long]("id") /? qs[Long]("long"),
        jsonRequest[Long]
      ),
      response = ok(jsonResponse[Long])
    )

```
generates parameters without `format` hint
```json
  "parameters": [
    {
      "name": "id",
      "in": "path",
      "schema": {
        "type": "integer"
      },
      "required": true
    },
    {
      "name": "long",
      "in": "query",
      "schema": {
        "type": "integer"
      },
      "required": true
    }
  ],
```
but request and response body content with `format` hint
```json
  "content": {
    "application/json": {
      "schema": {
        "type": "integer",
        "format": "int64"
      }
    }
  }
```

minedeljkovic:
Fixes #854

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -1524,10 +1524,10 @@ object Schema {
 
   val simpleUUID = Primitive("string", format = Some("uuid"), None, None, None)
   val simpleString = Primitive("string", None, None, None, None)
-  val simpleInt = Primitive("integer", format = Some("int32"), None, None, None)
+  val simpleInteger = Primitive("integer", format = Some("int32"), None, None, None)
   val simpleLong = Primitive("integer", format = Some("int64"), None, None, None)
   val simpleBoolean = Primitive("boolean", None, None, None, None)
-  val simpleDouble = Primitive("number", format = Some("double"), None, None, None)
+  val simpleNumber = Primitive("number", format = Some("double"), None, None, None)
 
 }
 
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -1524,9 +1524,10 @@ object Schema {
 
   val simpleUUID = Primitive("string", format = Some("uuid"), None, None, None)
   val simpleString = Primitive("string", None, None, None, None)
-  val simpleInteger = Primitive("integer", None, None, None, None)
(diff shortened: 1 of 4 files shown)


openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -1524,9 +1524,10 @@ object Schema {
 
   val simpleUUID = Primitive("string", format = Some("uuid"), None, None, None)
   val simpleString = Primitive("string", None, None, None, None)
-  val simpleInteger = Primitive("integer", None, None, None, None)
+  val simpleInt = Primitive("integer", format = Some("int32"), None, None, None)
```
julienrf, line 1527: This breaks the binary compatibility, do you mind keeping the previous name, `simpleInteger`? (same for `simpleNumber`)

julienrf (approved):
Thank you @minedeljkovic! I think this is only a partial fix for #854 because it is still not possible for user-defined query parameter types to include a format.

I think this is still an improvement over the status quo, though, thank you!

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
minedeljkovic: Oh, right.
Do you prefer keeping the old format-less values just for binary compatibility?
Or actually use `simpleInteger` with `int32` format and `simpleNumber` with `double` format?

julienrf: I would just keep the previous names:

```suggestion
  val simpleInteger = Primitive("integer", format = Some("int32"), None, None, None)
```

julienrf (approved):

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
minedeljkovic: Binary compatibility should be fixed now. @julienrf  Sorry for the glitch!

julienrf: _I_ am sorry. Managing the compatibility is complex. In this case, I thought we could relax the compatibility guarantees to `BinaryCompatible` only (instead of `BinaryAndSourceCompatibility`) instead of making `simpleLong` private. Let me try that.

PR #874, opened 2021-08-05
Http4s chunked encoding

GrafBlutwurst:
This PR is based on https://github.com/endpoints4s/endpoints4s/pull/510 thanks for all the people doing the hard parts there already.

This PR has two major problems: 

- The dependency that was used to convert between Akka types and FS2 was hosted on BinTray. To get this PR ready so it can be looked at I put the code in locally. I think we need a better solution for this before this can be merged.

- It turned out to be necessary to force the stream in the `ChunkedEntities.jsonChunksResponse` for http4s-client. This is because in the endpoint itself we end up with a Resource.use that yields an `F[Stream[F, Resp]]` when this stream is finally evaluated it will try to access the Stream of the original http4s `Response` and because the resource is already closed, come up empty. At least that's what I think is happening. 

- I did not encounter any reframing issues when using `bodyText` which I would have expected leading to parsing failures. If someone knows more about this, this might also still need addressing.


http4s/client/src/main/scala/endpoints4s/http4s/client/ChunkedEntities.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import fs2.Stream
+import cats.syntax.all._
+import endpoints4s.Valid
+import endpoints4s.Invalid
+
+trait ChunkedEntities extends endpoints4s.algebra.ChunkedEntities with EndpointsWithCustomErrors {
+
+  //#stream-type
+  type Chunks[A] = fs2.Stream[Effect, A]
```
julienrf, line 14: Just one thought: it means that the type of an `Endpoint[A, Chunks[B]]` expands to `A => Effect[Stream[Effect, B]]`, right? Is that OK?

http4s/client/src/main/scala/endpoints4s/http4s/client/ChunkedEntities.scala
```diff
@@ -0,0 +1,73 @@
...
+    req.withEntity(stream.map(encoder.encode))
+  }
+
+  def jsonChunksResponse[A](implicit
+      codec: JsonCodec[A]
+  ): ResponseEntity[Chunks[A]] = { response =>
+    val decoder = stringCodec[A](codec)
+
+    //I seem to have to force the entire stream here or I'll just get empty results, I can't quite put my finger where it happens. but i suspect if we propagate the stream wrong here the underlying http4s thing is already gone
+    // I suspect it's because of the use of `client.run(req).use(...) in def endpoint in the Endpoints.scala. Nothing forces compilation of the stream until that use is over, thus it gets closed and the underlying stream ends up empty.
```
julienrf, line 53: Why do we have to do this change only here, and not in `textChunksResponse`?

http4s/client/src/main/scala/endpoints4s/http4s/client/ChunkedEntities.scala
julienrf: Should we change the meaning of the type `Endpoint[A, B]` to the following?

~~~ scala
type Endpoint[A, B] = A => Effect[Resource[Effect, B]]
~~~

So that downstream users of the client could control the usage of the resource.

http4s/client/src/main/scala/endpoints4s/http4s/client/ChunkedEntities.scala
GrafBlutwurst: If I read the tests correctly, which I'm still bad at, I think it's just not tested. I'll get on that.

http4s/client/src/main/scala/endpoints4s/http4s/client/ChunkedEntities.scala
GrafBlutwurst: Addressed by changing Endpoints to a trait that contains a convenience function that consumes the resource directly and one that exposes it

http4s/client/src/main/scala/endpoints4s/http4s/client/ChunkedEntities.scala
GrafBlutwurst: added a couple of tests, but with the new design that exposes the resource we no longer need to force the streams in the first place.

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
```diff
@@ -239,22 +239,29 @@ trait EndpointsWithCustomErrors
...
-  //#endpoint-type
+  trait Endpoint[A, B] {
+    def send(request: A): Resource[Effect, B]
+
+    /** Note that this method CAN be unsafe if a lazy resource is returned as the Response due to closing the underlying Resource
+      *
+      * @param request
+      * @return
+      */
+    def sendAndConsume(request: A): Effect[B] = send(request).use(effect.pure)
```
julienrf, line 250: Maybe consider adding a deprecated `apply` method that would suggest using `sendAndConsume` instead, to ease migration?

http4s/server/src/test/scala/endpoints4s/http4s/server/ChunkedEntitiesServerInterpreterTest.scala
```diff
@@ -0,0 +1,125 @@
...
+      BlazeServerBuilder[IO](implicitly[ExecutionContext])
+        .bindHttp(port, "localhost")
+        .withHttpApp(httpApp)
+    server.resource.use(_ => IO(runTests(port))).unsafeRunSync()
+    ()
+  }
+
+  override def serveIdentityEndpoint[Resp](endpoint: serverApi.Endpoint[Resp, Resp])(
+      runTests: Int => Unit
+  ): Unit = ???
```
julienrf, line 112: Should we implement this?

http4s/server/src/test/scala/endpoints4s/http4s/server/ServerInterpreterTest.scala
```diff
@@ -10,7 +10,8 @@ import endpoints4s.algebra.server.{
   EndpointsTestSuite,
   JsonEntitiesFromSchemasTestSuite,
   SumTypedEntitiesTestSuite,
-  TextEntitiesTestSuite
+  TextEntitiesTestSuite,
+  ChunkedJsonEntitiesTestSuite
```
julienrf, line 14: Is this import used?

json-schema/json-schema/src/test/scala/endpoints4s/TuplerTests.scala
```diff
@@ -5,7 +5,8 @@ class TuplerTests {
   def tupling[A, B](a: A, b: B)(implicit tupler: Tupler[A, B]): tupler.Out =
     tupler(a, b)
 
-  def forall[A, B, C, D, E, F, G, H, I](a: A, b: B, c: C, d: D, e: E, f: F, g: G, h: H, i: I): Unit = {
+  def forall[A, B, C, D, E, F, G, H, I](a: A, b: B, c: C, d: D, e: E, f: F, g: G, h: H, i: I)
+      : Unit = {
```
julienrf, line 9: Would it be possible to revert the formatting changes in this file?

julienrf (changes requested):

julienrf:
Thank you @GrafBlutwurst, the changes look good to me but it would be great to include some documentation about the support of chunked entities.

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
```diff
@@ -243,23 +243,40 @@ trait EndpointsWithCustomErrors
       responseA(sc, hs)
         .map(_.andThen(_.map(_.asLeft[B])))
         .orElse(responseB(sc, hs).map(_.andThen(_.map(_.asRight[A]))))
-
-  //#endpoint-type
-  type Endpoint[A, B] = Kleisli[Effect, A, B]
+   //#endpoint-type
+  trait Endpoint[A, B] {
+    /** This method returns a Resource[Effect,B] unlike sendAndConsume this is always safe to use in regard to lazyness
```
julienrf, line 248:
```suggestion
    /** This method returns a Resource[Effect, B] unlike `sendAndConsume` this is always safe to use in regard to laziness
```

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
```diff
@@ -243,23 +243,40 @@ trait EndpointsWithCustomErrors
...
+      *
+      * @param request
+      * @return
```
julienrf, line 251:
```suggestion
```

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
julienrf: Could you maybe add a comment indicating _when_ the request is sent to the server? (Maybe this is obvious to the users of `Resource`, though?)

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
```diff
@@ -243,23 +243,40 @@ trait EndpointsWithCustomErrors
...
+      */
+
+    def send(request: A): Resource[Effect, B]
+
+    /** Note that this method CAN be unsafe if a lazy resource is returned as the Response due to closing the underlying Resource
+      *
+      * @param request
+      * @return
```
julienrf, line 259:
```suggestion
```

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
```diff
@@ -243,23 +243,40 @@ trait EndpointsWithCustomErrors
...
+      */
+    def sendAndConsume(request: A): Effect[B] = send(request).use(effect.pure)
+
+    @Deprecated(since = "5.0.0: This calls sendAndConsume which is potentially unsafe. Use sendAndConsume if you don't suffer any issues in regard to laziness, else use send")
```
julienrf, line 263:
```suggestion
    @deprecated(since = "6.0.0", message = "This calls sendAndConsume which is potentially unsafe. Use sendAndConsume if you don't suffer any issues in regard to laziness, else use send")
```

julienrf (approved):
Thank you @GrafBlutwurst! I found some typos or minor issues.

Would it be possible to squash the commits before merging?

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
GrafBlutwurst: It is possible that this depends on the actual chosen http4s client. so it's hard to say in general

PR #876, opened 2021-08-06
Update http4s to 0.23.0

LukaJCB:
Replaces #869. Server currently passes all tests, but there's some errors in the client code, that I assume is due to Uri path issues.

### DIFF
--- a/http4s/build.sbt
+++ b/http4s/build.sbt
@@ -13,7 +13,7 @@ val `http4s-server` =
       `scala 2.12 to dotty`,
       name := "http4s-server",
       versionPolicyIntention := Compatibility.BinaryCompatible,
-      version := "6.0.0+n",
+      version := "7.0.0+n",
       libraryDependencies ++= Seq(
         ("org.http4s" %% "http4s-core" % http4sVersion).cross(CrossVersion.for3Use2_13),
         ("org.http4s" %% "http4s-dsl" % http4sVersion).cross(CrossVersion.for3Use2_13),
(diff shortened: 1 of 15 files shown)


http4s/build.sbt
```diff
@@ -13,7 +13,7 @@ val `http4s-server` =
       `scala 2.12 to dotty`,
       name := "http4s-server",
       versionPolicyIntention := Compatibility.BinaryCompatible,
-      version := "6.0.0+n",
+      version := "7.0.0+n",
```
LukaJCB, line 16: well this didn't work, any help?

julienrf: Sorry, I know this is a bit complex... In this situation we want relax the intention to Compatibility.None (in the line above, instead of changing the version).

LukaJCB: Awesome, looks like it's all green now 🙌🏼

LukaJCB:
Can I get a review and merge @julienrf? :)

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
```diff
@@ -137,7 +136,13 @@ trait EndpointsWithCustomErrors
     effect.map(effect.fromEither(url.encodeUrl(urlP)))(uri =>
       entity(
         bodyP,
-        headers(headersP, Http4sRequest(method, host.withPath(host.path + uri)))
+        headers(
+          headersP,
+          Http4sRequest(
+            method,
+            host.withPath(Uri.Path.unsafeFromString(host.path.renderString + uri.renderString))
```
julienrf, line 143: Would it make sense to change our internal model so that we don’t need to go back and forth to `String` representation? (This could be addressed as a separate PR)

julienrf (approved):
Thank you @LukaJCB!

julienrf:
May I let you squash the commits before we merge the PR?

http4s/client/src/main/scala/endpoints4s/http4s/client/Endpoints.scala
LukaJCB: Yeah I think it would! There was some discrepancies in there somewhere that resulted in test failures when I tried this earlier, but I can create a ticket after we merge!

LukaJCB:
Awesome, thank you @julienrf! Can we just use GHs squash and merge feature?

julienrf:
Sure, I can squash myself, although I usually like to let the contributors do it themselves, since they fully author the work, and they can control what to put in the commit message.

LukaJCB:
Gotcha, I don't mind either way! :) Can we get this released sometime soon?

PR #877, opened 2021-08-06
Change OpenAPI encoding of heterogeneous arrays

### LINKED ISSUE #856: Encoding of tuples in OpenAPI does not conform to spec
Example; this definition
```scala
  val foo: Endpoint[Unit, (String, Long)] =
    endpoint(
      request = get(path / "foo"),
      response = ok(jsonResponse[(String, Long)])
    )
```
produces this OpenAPI definition for response content (`items` is an array of SchemaObject)
```json
"content": {
  "application/json": {
    "schema": {
      "type": "array",
      "items": [
        {
          "type": "string"
        },
        {
          "type": "integer",
          "format": "int64"
        }
      ]
    }
  }
}
```
but, OpenAPI spec is explicitly against that. In [this part](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schema-object), it's stated:
> items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. items MUST be present if the type is array.

OpenAPI 3.0 [doesn't seem to have a proper encoding for tuples](https://stackoverflow.com/a/57465199), unfortunately, but it might be better to produce (not perfectly precise) approximation that conforms spec.

minedeljkovic:
Fixes #856

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -89,14 +89,27 @@ object OpenApi {
         }
         obj.additionalProperties.foreach(p => fields += "additionalProperties" -> schemaJson(p))
       case array: Schema.Array =>
-        val itemsSchema = array.elementType match {
-          case Left(value)  => schemaJson(value)
-          case Right(value) => ujson.Arr(value.map(schemaJson): _*)
+        fields += "type" -> "array"
+        array.elementType match {
+          case Left(value) =>
+            fields += "items" -> schemaJson(value)
+          case Right(value) =>
+            /** Best effort to represent the heterogeneous array in OpenAPI 3.0
+              * This should be changed with OpenAPI 3.1 and more idiomatic representation using `prefixItems`
+              */
+            fields ++= List(
+              "items" -> schemaJson(
+                Schema.OneOf(
+                  alternatives = Schema.EnumeratedAlternatives(value),
+                  description = None,
+                  example = None,
+                  title = None
+                )
+              ),
+              "minItems" -> ujson.Num(value.length.toDouble),
+              "maxItems" -> ujson.Num(value.length.toDouble)
+            )
         }
-        fields ++= List(
-          "type" -> "array",
-          "items" -> itemsSchema
-        )
       case enm: Schema.Enum =>
         fields ++= schemaJson(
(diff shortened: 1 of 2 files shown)


openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -89,14 +89,27 @@ object OpenApi {
...
+            fields += "items" -> schemaJson(value)
+          case Right(value) =>
+            /** Best effort to represent the heterogeneous array in OpenAPI 3.0
+              * This should be changed with OpenAPI 3.1 and more idiomatic representation using `prefixItems`
+              */
+            fields ++= List(
+              "items" -> schemaJson(
+                Schema.OneOf(
+                  alternatives = Schema.EnumeratedAlternatives(value),
+                  description = None,
```
minedeljkovic, line 104: I was thinking of maybe using the description here to stress additional constraint (fixed positions of elements in a tuple), that cannot be enforced by OpenAPI 3.0 features (along the lines of [this example](https://stackoverflow.com/a/57465199/15821235)), but I'm not sure what would be an appropriate format/language.

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -89,14 +89,26 @@ object OpenApi {
...
+            fields += "items" -> schemaJson(value)
+          case Right(value) =>
+            // Best effort (not 100% accurate) to represent the heterogeneous array in OpenAPI 3.0
+            // This should be changed with OpenAPI 3.1 and more idiomatic representation using `prefixItems`
+            fields ++= List(
+              "items" -> schemaJson(
+                Schema.OneOf(
+                  alternatives = Schema.EnumeratedAlternatives(value),
+                  description = None,
+                  example = None,
```
julienrf, line 104: We could take the first defined example of the tuple elements, if any.

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
julienrf: I think it’s fine to leave it empty. What do others think?

julienrf (approved):
Thank you @minedeljkovic!

(I’ve just edited my review and deleted a wrong comment.)

PR #879, opened 2021-08-07
Query string param default value

### LINKED ISSUE #660: Optional query string parameter with a default value
One thing that has come up at work several times: we'd like to be able to declare an optional query string parameter with a default value. OpenAPI supports describing default values (see "Default Parameter Values" on [this page](https://swagger.io/docs/specification/describing-parameters/)).

I've not given much thought to syntax for this, but ideally you could do something like this:

```scala
path / "api" / "v1" / "pets" /? qs[Int]("limit", default = 5): Url[Int]
```

Then, the endpoint would have a default value

  * server interpreters would look for an optional param and use the default value as a fallback
  * documentation interpreters would generate docs mentioning the default value (Swagger UI knows how to generate forms that fill this default in nicely, I think)
  * client interpreters would be generated always specifying the value (so never using the default)

---

As a workaround, we've been doing the following, but that forces us to put information about the default value in documentation, and the generated OpenAPI isn't as useful. 

```scala
path / "api" / "v1" / "pets" /? qs[Option[Int]]("limit").xmap(_.getOrElse(5))(Some(_)): Url[Int]
```

minedeljkovic:
Should fix #660

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
+++ b/algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
@@ -121,8 +121,9 @@ trait Urls extends PartialInvariantFunctorSyntax {
     * @group operations
     */
   def optQsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
-      value: QueryStringParam[Option[A]]
-  ): QueryString[A] = qs(name, docs)(value).xmap(_.getOrElse(default))(Some(_))
+      value: QueryStringParam[A]
+  ): QueryString[A] =
+    qs(name, docs)(optionalQueryStringParam(value)).xmap(_.getOrElse(default))(Some(_))
 
   /** Make a query string parameter optional:
     *
--- a/algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
+++ b/algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
@@ -108,6 +108,22 @@ trait Urls extends PartialInvariantFunctorSyntax {
       value: QueryStringParam[A]
   ): QueryString[A]
 
+  /** Builds a `QueryString` with one optional parameter, which has a default value.
+    *
+    * Examples:
+    *
+    * {{{
+    *   optQsWithDefault[Int]("page", 1) // optional `page` parameter, with default value 1
+    * }}}
+    *
+    * @param name Parameter’s name
+    * @tparam A Type of the value carried by the parameter
+    * @group operations
+    */
+  def optQsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
+      value: QueryStringParam[Option[A]]
+  ): QueryString[A] = qs(name, docs)(value).xmap(_.getOrElse(default))(Some(_))
+
   /** Make a query string parameter optional:
     *
     * {{{
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
@@ -1,7 +1,8 @@
-package endpoints4s
-package openapi
+package endpoints4s.openapi
 
 import java.util.UUID
+
+import endpoints4s.{ujson => _, _}
 
 import endpoints4s.algebra.Documentation
 import endpoints4s.openapi.model.Schema
@@ -48,13 +49,40 @@ trait Urls extends algebra.Urls {
       )
     )
 
-  case class DocumentedQueryStringParam(schema: Schema, isRequired: Boolean)
-  type QueryStringParam[A] = DocumentedQueryStringParam
+  override def optQsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
+      value: QueryStringParam[Option[A]]
+  ): QueryString[A] =
+    DocumentedQueryString(
+      List(
+        DocumentedParameter(
+          name,
+          required = false,
+          docs,
+          schema = {
+            val defaultJson = value.encoder.encode(Some(default))
+            value.schema.withDefinedDefault(Some(defaultJson))
+          }
+        )
+      )
+    )
+
+  case class DocumentedQueryStringParam[A](
+      schema: Schema,
+      isRequired: Boolean,
+      encoder: Encoder[A, ujson.Value]
+  )
+  type QueryStringParam[A] = DocumentedQueryStringParam[A]
 
   implicit def optionalQueryStringParam[A](implicit
       param: QueryStringParam[A]
   ): QueryStringParam[Option[A]] =
-    param.copy(isRequired = false)
+    param.copy(
+      isRequired = false,
+      encoder = {
+        case None    => ujson.Null
+        case Some(a) => param.encoder.encode(a)
+      }
+    )
 
   implicit def repeatedQueryStringParam[A, CC[X] <: Iterable[X]](implicit
       param: QueryStringParam[A],
@@ -67,7 +95,8 @@ trait Urls extends algebra.Urls {
         example = None,
         title = None
       ),
-      isRequired = false
+      isRequired = false,
+      encoder = cc => ujson.Arr(cc.map(param.encoder.encode))
     )
 
   implicit lazy val queryStringPartialInvariantFunctor: PartialInvariantFunctor[QueryString] =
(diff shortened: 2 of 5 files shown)


minedeljkovic:
@julienrf  When you have time, please check if I'm on the right track here.

The first commit is a shameless cherry-pick of [this commit](https://github.com/endpoints4s/endpoints4s/commit/20e77bf3f3f8663edb3c230a514eba4856174ad5).

algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
```diff
@@ -108,6 +108,22 @@ trait Urls extends PartialInvariantFunctorSyntax {
...
+    *
+    * {{{
+    *   qsWithDefault[Int]("page", 1) // optional `page` parameter, with default value 1
+    * }}}
+    *
+    * @param name Parameter’s name
+    * @tparam A Type of the value carried by the parameter
+    * @group operations
+    */
+  def qsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
```
minedeljkovic, line 123: I tried to add this method as the `qs` overload, but compiler complained about multiple overloads with default values. So, I named this one differently

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
```diff
@@ -48,13 +49,49 @@ trait Urls extends algebra.Urls {
...
+          docs,
+          schema = {
+            def withDefault(defaultValue: Option[ujson.Value]) = value.schema match {
+              case s: Schema.Object    => s.withDefault(defaultValue)
+              case s: Schema.Array     => s.withDefault(defaultValue)
+              case s: Schema.Enum      => s.withDefault(defaultValue)
+              case s: Schema.Primitive => s.withDefault(defaultValue)
+              case s: Schema.OneOf     => s.withDefault(defaultValue)
+              case s: Schema.AllOf     => s.withDefault(defaultValue)
+              case s: Schema.Reference => s
```
minedeljkovic, line 69: Default value is not supported for `Reference` schema.
Should we better fail here, instead of doing nothing?
Anyway, `Reference` schema in the position of any kind of query string parameter (required, optional or with default value) is probably unexpected.

minedeljkovic:
Hmm, why [this compiler error](https://github.com/endpoints4s/endpoints4s/pull/879/checks?check_run_id=3270765891#step:8:104) appeared? It's not obvious to me how is it related to these changes.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
julienrf: Looking at #881, do you think we could follow a similar approach here and attach the default value to the `DocumentedQueryStringParam` instead of the `Schema`, to avoid this issue?

julienrf (commented):
Thank you @minedeljkovic! It seems that @harpocrates found in #881 a better way of modeling default values in documentation than what I originally proposed, I think it is worth investigating that approach, what do you think?

algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
minedeljkovic: Thanks to [this great comment](https://github.com/endpoints4s/endpoints4s/pull/879#discussion_r687485809), I added a default implementation to algebra, which should solve compatibility issues and make this change much simpler in general.
I'm not sure why the CI fails for `versionPolicyCheck`, though, I don't have any issues running it locally. Would a squash help maybe?

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Urls.scala
```diff
@@ -205,6 +205,18 @@ trait Urls extends algebra.Urls with StatusCodes {
...
+  def qsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
+      value: QueryStringParam[Option[A]]
+  ): QueryString[A] = {
+    val optQs = qs(name, docs)(value)
+    new QueryString[A] {
+      def validate(params: Map[String, List[String]]): Validated[A] =
+        optQs.validate(params).map(_.getOrElse(default))
+      def encode(a: A): Uri.Query =
+        optQs.encode(Some(a))
+    }
```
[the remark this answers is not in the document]
minedeljkovic: Thanks! Added this as default implementation in algebra.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
minedeljkovic: I guess there is already agreement that #885 workaround is the right solution to this in OpenAPI 3.0.
So, I think it's best to rebase this on top of those changes, once they are stable. Do you agree @julienrf ?

julienrf: Yes, I agree

harpocrates: I think using `DocumentedQueryStringParam` is still a good idea though, right? #885 just makes it possible to reliably fix a default value onto a `Schema`.

harpocrates: Also, I just realized you may also want to use [`Schema#withDefinedDefault(default: Option[String]): Schema`](https://github.com/endpoints4s/endpoints4s/pull/881/files#diff-1f590f2eefffc13ac5521a40b4b8f3a3d5e539bdb5c8551508d2868c0c6c5508R975) which I just added in #881. Maybe that method should also have been added as part of #885 - sorry 😬

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
```diff
@@ -48,13 +49,49 @@ trait Urls extends algebra.Urls {
...
+            withDefault(Some(value.encoder.encode(Some(default))))
+          }
+        )
+      )
+    )
+
+  case class DocumentedQueryStringParam[A](
+      schema: Schema,
+      isRequired: Boolean,
+      encoder: Encoder[A, ujson.Value]
```
harpocrates, line 81: I don't think this field should be an `encoder` - it should be fine to just have `default: ujson.Value`. Provided you have a `tpe: JsonSchema[A]` in scope when you create your `DocumentedQueryStringParam`, you can use your `defaultValue: A` and do

```scala
   val default: ujson.Value = Some(tpe.ujsonSchema.encoder.encode(defaultValue))
```

algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
minedeljkovic: CI error disappeared after rebase.
Note I added `versionPolicyIntention := Compatibility.None` to `openapi` module build definition, as I understand that's the thing to do before the next release when there is a incompatible change in interpreter.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
minedeljkovic: There is no `JsonSchema[A]` in scope when `DocumentedQueryStringParam` is created, and that is exactly why I added `encoder` here, to be able to encode `default` value when setting it in schema. I see `encoder` as just what is minimally needed here (`JsonSchema[A]` also provides other features, not needed here at the moment). I'm open to include `JsonSchema[A]` if there is a simpler way than this, but I'm not sure I see it at the moment.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
minedeljkovic: Rebased to the latest master that contains #885 

>Looking at #881, do you think we could follow a similar approach here and attach the default value to the DocumentedQueryStringParam instead of the Schema, to avoid this issue?

>I think using a default value fields on DocumentedQueryStringParam is still a good idea though, right? #885 just makes it possible to reliably fix a default value onto a Schema.

As far as I understand, default value should be attached to the `Schema`, to be included in the generated OpenAPI in the end. In that sense, having it in the `DocumentedQueryStringParam` too seems redundant, since it already contains `Schema`.

> Also, I just realized you may also want to use Schema#withDefinedDefault(default: Option[String]): Schema which I just added in #881. Maybe that method should also have been added as part of #885 - sorry 😬

Yeah, I tried to add that before here, but without #885 it wasn't clean. I copied that method here, if you don't mind, as it should be merged away anyway. Now that duplicated logic is removed from `Url` trait.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
harpocrates: Oh right, I see what you mean.

algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
```diff
@@ -108,6 +108,22 @@ trait Urls extends PartialInvariantFunctorSyntax {
...
+    * {{{
+    *   optQsWithDefault[Int]("page", 1) // optional `page` parameter, with default value 1
+    * }}}
+    *
+    * @param name Parameter’s name
+    * @tparam A Type of the value carried by the parameter
+    * @group operations
+    */
+  def optQsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
+      value: QueryStringParam[Option[A]]
```
harpocrates, line 124: Shouldn't this just be `value: QueryStringParam[A]`? In the body we can use

```scala
qs(name, docs)(optionalQueryStringParam(value)).xmap(_.getOrElse(default))(Option(_))
```

pretty much exactly as is done for optional fields.

minedeljkovic: Great suggestion!
Fixed.

build.sbt
```diff
@@ -36,7 +36,7 @@ ThisBuild / sonatypeProjectHosting := Some(
 // We want to keep binary compatibility as long as we can for the algebra,
 // but it is OK to publish breaking releases of interpreters. So,
 // interpreter modules may override this setting.
-ThisBuild / versionPolicyIntention := Compatibility.BinaryAndSourceCompatible
+ThisBuild / versionPolicyIntention := Compatibility.BinaryCompatible
```
julienrf, line 39: I don’t think we even need this now?

julienrf: I was wrong, I’m deleting this discussion.

algebras/algebra/src/main/scala/endpoints4s/algebra/Urls.scala
```diff
@@ -108,6 +108,23 @@ trait Urls extends PartialInvariantFunctorSyntax {
...
+    *   optQsWithDefault[Int]("page", 1) // optional `page` parameter, with default value 1
+    * }}}
+    *
+    * @param name Parameter’s name
+    * @tparam A Type of the value carried by the parameter
+    * @group operations
+    */
+  def optQsWithDefault[A](name: String, default: A, docs: Documentation = None)(implicit
+      value: QueryStringParam[A]
+  ): QueryString[A] =
```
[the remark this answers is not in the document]
julienrf: So, the problem with `QueryString[A]` is that clients will have to provide a value, even though it’s optional, and the problem with `QueryString[Option[A]]` is that servers will have to deal with `None` although in practice there will always be a value.

Note that the situation is the same with `optFieldWithDefault`, which returns a `JsonSchema[A]`. This means that clients have to provide a value. I guess this is not a problem because `optFieldWithDefault` is mostly used via the [generic derivation mechanism](https://github.com/endpoints4s/endpoints4s/blob/19f893087c0f8f9060f14e79e3925b4030dba738/json-schema/json-schema-generic/src/main/scala-2/endpoints4s/generic/JsonSchemas.scala#L134), so eventually, the compiler does provide the default value (when you create a case class with a field that has a default value, the compiler supplies that value for you).

I wonder what we should do here. Maybe introduce a type `WithDefault[A]` which would be `Option[A]` on the client interpreters and `A` on the server interpreters?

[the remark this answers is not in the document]
minedeljkovic: @julienrf I just opened #941 with, what I understand, is your idea to solve this

julienrf:
I am closing this one in favor of #941

PR #881, opened 2021-08-09
OpenAPI: document default field values

### LINKED ISSUE #880: OpenAPI: default values of optional fields are not documented
Taken from <https://swagger.io/specification/#schema-object>:

> The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification.
>
> [snipped]
> * default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if `type` is `string`, then default can be `"foo"` but cannot be `1`.

For instance, consider the following snippet of code that generates the OpenAPI doc with a schema for `Foo`:

```scala
// import $ivy.`org.endpoints4s::algebra:1.4.0`
// import $ivy.`org.endpoints4s::json-schema-generic:1.4.0`
// import $ivy.`org.endpoints4s::openapi:3.0.0`

import endpoints4s.openapi.model.{OpenApi, Info}
import endpoints4s.Encoder
import java.nio.file._

final case class Foo(x: Int, y: Int = 1)

object Example
  extends endpoints4s.algebra.Endpoints
    with endpoints4s.algebra.JsonEntitiesFromSchemas
    with endpoints4s.generic.JsonSchemas
    with endpoints4s.openapi.Endpoints
    with endpoints4s.openapi.JsonEntitiesFromSchemas {

  implicit lazy val fooSchema: JsonSchema[Foo] = genericJsonSchema[Foo]

  lazy val foo: Endpoint[Unit, Foo] =
    endpoint(
      request = get(path / "foo"),
      response = ok(jsonResponse[Foo])
    )

  lazy val schema = implicitly[Encoder[OpenApi, String]]
  lazy val api = openApi(Info(title = "thatDot Connect API", version = "0.0.1"))(foo)
}

Files.write(Paths.get("doc.json"), Example.schema.encode(Example.api).getBytes("UTF-8"))
```

The schema for `Foo` in `doc.json` contains:

```json
      ".line3..read..iw..iw.Foo": {
        "type": "object",
        "properties": {
          "x": {
            "type": "integer",
            "format": "int32"
          },
          "y": {
            "type": "integer",
            "format": "int32"
          }
        },
        "required": [
          "x"
        ]
      }
```

I would like to instead have

```json
      ".line3..read..iw..iw.Foo": {
        "type": "object",
        "properties": {
          "x": {
            "type": "integer",
            "format": "int32"
          },
          "y": {
            "type": "integer",
            "format": "int32",
            "default": 1
          }
        },
        "required": [
          "x"
        ]
      }
```

Note that tools like Swagger UI _do_ use this `default` field:

Screen Shot 2021-08-09 at 6 51 06 AM

harpocrates:
`optFieldWithDefault` is no longer final and the OpenAPI interpreter
overrides it so as to track the default value in the documentation.

Fixes #880

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
@@ -1,6 +1,13 @@
 package endpoints4s.openapi
 
-import endpoints4s.{NumericConstraints, PartialInvariantFunctor, Tupler, Validated, algebra}
+import endpoints4s.{
+  Hashing,
+  NumericConstraints,
+  PartialInvariantFunctor,
+  Tupler,
+  Validated,
+  algebra
+}
 import endpoints4s.algebra.Documentation
 import endpoints4s.openapi.model.Schema
 import endpoints4s.openapi.model.Schema.{DiscriminatedAlternatives, EnumeratedAlternatives}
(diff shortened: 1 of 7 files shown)


harpocrates:
What is the procedure around versioning? I guess a change like this warrants an update in `openapi` version number. Is that supposed to happen in this PR or later?

julienrf (changes requested):
Thank you @harpocrates for the PR! I think it would be good to include a test showing the JSON representation of the OpenAPI document (the test you added only checks that the OpenAPI model is correct).

Also, unfortunately adding a field to the case class `Field` breaks binary compatibility. This could be fixed by turning `Field` into a regular class having all the methods it currently has (including the overrides of `toString`, `hashCode`, and `equals`), plus a new constructor that would take the `defaultValue` parameter.

harpocrates:
> Also, unfortunately adding a field to the case class Field breaks binary compatibility. This could be fixed by turning Field into a regular class having all the methods it currently has (including the overrides of toString, hashCode, and equals), plus a new constructor that would take the defaultValue parameter.

Is that enough? Doesn't the non-final nature of `Field` mean that we'll break binary compatibility by adding _any_ member? (And making it `final` would also itself break binary compatibility.)

julienrf:
If we manage to provide the same ABI as in the previous version, yes that will be enough (I might be missing something, but it seems to me that we can provide the same ABI).

harpocrates:
Indeed! I think I've fixed the Mima problem, although the diff is quite a bit more ugly. Is this the sort of thing we can clean up when we next make a major version bump?

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -127,12 +134,83 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
+      }
+
+      def this(
+          name: String,
+          tpe: DocumentedJsonSchema,
+          isOptional: Boolean,
+          documentation: Option[String]
+      ) = this(name, tpe, isOptional, None, documentation)
+
+      def copy(
```
julienrf, line 176: This method should be deprecated, instead we should add methods like `withName`, `withType`, etc.

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -127,12 +134,83 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
+
+      def apply(
+          name: String,
+          tpe: DocumentedJsonSchema,
+          isOptional: Boolean,
+          defaultValue: Option[ujson.Value],
+          documentation: Option[String]
+      ): Field = new Field(name, tpe, isOptional, defaultValue, documentation)
+
+      def unapply(field: Field): Option[(String, DocumentedJsonSchema, Boolean, Option[String])] =
```
julienrf, line 204: This extractor should be deprecated.

openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala
```diff
@@ -127,12 +134,103 @@ trait JsonSchemas extends algebra.JsonSchemas with TuplesSchemas {
...
+        case 4 => documentation
+        case _ => throw new IndexOutOfBoundsException(idx.toString)
+      }
+      override def toString: String = s"Field($name,$tpe,$isOptional,$defaultValue,$documentation)"
+      override def hashCode: Int = Hashing.hash(name, tpe, isOptional, defaultValue, documentation)
+      override def equals(other: Any): Boolean = other match {
+        case field: Field =>
+          name == field.name && tpe == field.tpe && isOptional == field.isOptional &&
+            defaultValue == field.defaultValue && documentation == field.documentation
+        case _ => false
```
harpocrates, line 166: @julienrf do you know the "right" way to handle this Scala issue (IMHO, really a wart)? 

```
[warn] openapi/openapi/src/main/scala/endpoints4s/openapi/JsonSchemas.scala:163:14: The outer reference in this type test cannot be checked at run time.
[warn]         case field: Field =>
[warn]              ^
[warn] one warning found
```

Should we switch `Field` back to not be `final`?

julienrf: Good point, maybe we can use `deprecatedInheritance` instead?

harpocrates:
I think this is ready for review again!

julienrf (approved):

PR #883, opened 2021-08-12
Ensure insertion endpoints is preserved in OpenAPI

bmeesters:


bmeesters:
Hello @julienrf, I ultimately had to change the `openApi` method internally, but also the `OpenAPI` class, since it expects an immutable Map. And I don't think I can go from a mutable to an immutable map without losing the order again. Though, I think the current approach is backwards compatible, and fairly small, so I think this the way to go?

openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
```diff
@@ -24,22 +26,21 @@ trait EndpointsWithCustomErrors
...
-        .groupBy(_.path)
-        .map { case (k, es) =>
-          (
-            k,
-            es.tail.foldLeft(PathItem(es.head.item.operations)) { (item, e2) =>
-              PathItem(item.operations ++ e2.item.operations)
-            }
-          )
-        }
+    val map = mutable.LinkedHashMap.empty[String, PathItem]
```
julienrf, line 29: I think `items` was a more descriptive name, or maybe `pathItems`?

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -1,21 +1,22 @@
...
 
-import java.io.Serializable
-
-import endpoints4s.{Encoder, Hashing}
 import endpoints4s.algebra.{ExternalDocumentationObject, Tag}
+import endpoints4s.{Encoder, Hashing}
 
-import scala.collection.mutable
+import java.io.Serializable
+import scala.collection.{Map, mutable, immutable}
```
julienrf, line 7: This import changed all the places where we were previously using `Map` (which was `scala.collection.immutable.Map`). I suggest removing the import and fully qualifying the name `scala.collection.Map` instead, where we use it.

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -1,21 +1,22 @@
...
 
 /** @see [[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md]]
   * @note Throws an exception on creation if several tags have the same name but not the same other attributes.
   */
 final class OpenApi private (
     val info: Info,
-    val paths: Map[String, PathItem],
+    private val innerPaths: Map[String, PathItem],
```
julienrf, line 14: I think it should be possible to use `scala.collection.SeqMap` instead, to make sure we don’t transform it into a `Map` type that is not ordered, by mistake.

After a second read, I realize that this may be difficult because the method `OpenApi.apply` would have an incompatible signature (maybe this would be fixed by adding an overload, though).

julienrf (approved):
Thank you @bmeesters, this is a nice improvement! You just need to fix the way you import `Map`, see below.

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
bmeesters: Will check, thanks for the quick feedback!

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -1,21 +1,22 @@
 package endpoints4s.openapi.model
 
-import java.io.Serializable
-
-import endpoints4s.{Encoder, Hashing}
 import endpoints4s.algebra.{ExternalDocumentationObject, Tag}
+import endpoints4s.{Encoder, Hashing}
 
+import java.io.Serializable
```
julienrf, line 6: I don’t think we need this import

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -1,21 +1,22 @@
...
 
 /** @see [[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md]]
   * @note Throws an exception on creation if several tags have the same name but not the same other attributes.
   */
 final class OpenApi private (
     val info: Info,
-    val paths: Map[String, PathItem],
+    private val innerPaths: collection.Map[String, PathItem],
     val components: Components
 ) extends Serializable {
```
julienrf, line 16: We have to add the following filter in `openapi/build.sbt`, in the settings of the `openapi` project:

~~~ scala
mimaBinaryIssueFilters ++= Seq(
  ProblemFilters.exclude[IncompatibleMethTypeProblem]("endpoints4s.openapi.model.OpenApi.this")
)
~~~

See https://stackoverflow.com/questions/31010755/scala-is-not-creating-private-constructor-for-companion-class

julienrf (approved):

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
bmeesters: Not sure how that got inserted :sweat_smile:

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
bmeesters: Does it matter where in the file this is added? (for example within the openApi val, or is at the end of the file fine?)

bmeesters:
Thanks for the quick feedback loop @julienrf, nice to get this in so fast :smile: Also nice to learn some new stuff about MiMa

PR #885, opened 2021-08-12
OpenAPI: use allOf with $ref whenever appropriate

harpocrates:
`$ref` JSON schemas are a bit special in that they aren't supposed to
have any sibling properties at all. Properties that would normally be
fields in the JSON schema object are supposed to be unified with the
`$ref` schema object with an explicit `allOf`.

Setting examples, descriptions, etc. on a reference schema now behaves
in the intuitive way (at least intuitive for me): it will override the
example, description, etc. on the _local instance_ of the schema, but
it won't otherwise modify the undderlying schema.

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -163,6 +163,9 @@ object OpenApi {
     for (title <- schema.title) {
       fields += "title" -> title
     }
+    for (default <- schema.default) {
+      fields += "default" -> default
+    }
     new ujson.Obj(fields)
   }
 
@@ -926,6 +929,7 @@ object MediaType {
 sealed trait Schema {
   def description: Option[String]
   def example: Option[ujson.Value]
+  def default: Option[ujson.Value]
   def title: Option[String]
 
   /** @return The same schema with its description overridden by the given `description`,
@@ -957,19 +961,20 @@ object Schema {
       val additionalProperties: Option[Schema],
       val description: Option[String],
       val example: Option[ujson.Value],
-      val title: Option[String]
+      val title: Option[String],
+      val default: Option[ujson.Value]
   ) extends Schema
       with Serializable {
 
     override def toString: String =
-      s"Object($properties, $additionalProperties, $description, $example, $title)"
+      s"Object($properties, $additionalProperties, $description, $example, $title, $default)"
 
     override def equals(other: Any): Boolean =
       other match {
         case that: Object =>
           properties == that.properties && additionalProperties == that.additionalProperties &&
             description == that.description && example == that.example &&
-            title == that.title
+            title == that.title && default == that.default
         case _ => false
       }
 
@@ -979,7 +984,8 @@ object Schema {
         additionalProperties,
         description,
         example,
-        title
+        title,
+        default
       )
 
     private[this] def copy(
@@ -987,9 +993,10 @@ object Schema {
         additionalProperties: Option[Schema] = additionalProperties,
         description: Option[String] = description,
         example: Option[ujson.Value] = example,
-        title: Option[String] = title
+        title: Option[String] = title,
+        default: Option[ujson.Value] = default
     ): Object =
-      new Object(properties, additionalProperties, description, example, title)
+      new Object(properties, additionalProperties, description, example, title, default)
 
     def withProperty(properties: List[Property]): Object =
       copy(properties = properties)
@@ -1005,6 +1012,9 @@ object Schema {
 
     def withTitle(title: Option[String]): Object =
       copy(title = title)
+
+    def withDefault(default: Option[ujson.Value]): Object =
+      copy(default = default)
   }
 
   object Object {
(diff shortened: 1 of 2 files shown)


harpocrates:
@julienrf I'm very interested in hearing what you have to say about this. I think this fixes elegantly multiple problems I've run into over the last several months:

  * `default` (I cherry-picked your commit again)
       * this is going to make #881 more robust (eg. if you try to set a default value for a field that is actually a reference to a schema)
       * #879 should also work smoothly (specifically, [this comment's concern]( https://github.com/endpoints4s/endpoints4s/pull/879#discussion_r684675142) would be handled)
  * `example`
       * one of the really annoying issues I've had at work in our use of _endpoints4s_ is that we can't override examples on a per-endpoint basis when the schema is actually a reference. With this change, that becomes possible. If there is interest, I could also upstream our `def jsonRequestWithExample[A](example: A)(implicit codec: JsonRequest[A]): RequestEntity[A]` combinators (ditto for responses and text entities).
  * `title` (although overriding the title doesn't seem useful to me, Swagger appears to support it)
  * `description`
       * you currently can try to set a description on a schema reference, but it will generate an invalid OpenAPI document

julienrf:
Interesting, thank you for the investigation @harpocrates! Do you know if this trick is well established in OpenAPI? Or is it just working with Swagger UI?

AFAICT, it should be possible to change a few things here and there to make things binary compatible, would you be fine with doing this work?

harpocrates:
It looks like siblings to `$ref` will be allowed in OpenAPI 3.1 (here's an [issue on the OpenAPI spec repo](https://github.com/OAI/OpenAPI-Specification/issues/1514)). In OpenAPI 3.0 and OpenAPI 2.0, the `allOf` tricks is widely accepted. Here's a [SO post on exactly this topic](https://stackoverflow.com/a/41752575/3072788). On the Swagger UI bug tracker, the consensus seems to be that the `allOf` approach is standards compliant and not just a Swagger UI thing.

I'm not sure how they worked around it, but the underlying issue was apparently due to wanting to follow the JSON Schema specification, which (as per the [section on `$ref` from the "understanding JSON Schema" website](https://json-schema.org/understanding-json-schema/structuring.html#ref) is not supposed to have siblings for `$ref`.

> In Draft 4-7, `$ref` behaves a little differently. When an object contains a `$ref` property, the object is considered a reference, not a schema. Therefore, any other properties you put in that object will not be treated as JSON Schema keywords and will be ignored by the validator. `$ref` can only be used where a schema is expected. 

Given that _endpoints4s_ has `OpenApi.openApiVersion == "3.0.0"`, I think we should use the `allOf` trick. However, since this appears to also be changing soon<sup>TM</sup>, maybe `def openApi` could also take an option that would explicitly opt-in to the new 3.1.0 behavior? Does that sound OK? If so, I'll go fix the Mima issues and add the option.

julienrf:
Awesome, thank you for the detailed explanation!

I would say that there is no need to rush for supporting OpenAPI 3.1, but if you want to add support for it that’s great!

> Does that sound OK?

OK!

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -1467,38 +1576,49 @@ object Schema {
...
-      val description: Option[String]
+      val description: Option[String],
+      val example: Option[ujson.Value],
+      val title: Option[String], // you probably want the title on the original schema!
+      override val default: Option[ujson.Value]
   ) extends Schema
       with Serializable {
 
-    override val example: None.type = None // Reference objects can’t have examples
-    override val title: None.type = None // Reference objects can’t have a title
```
harpocrates, line 1482: These will cause binary incompatibility due to the type going from `None.type` to `Option[String]`. Any ideas for resolving that?

julienrf: I don’t think `None` has methods that don’t exist on `Option[String]`, so I think it’s OK to add a MiMa filter here.

harpocrates: Is it similarly OK to add exclusions for the `private` constructors then? For instance, the one on the lines that follow, eg:

```scala
     @deprecated("use `Reference.apply` instead", "3.1.0")
     private[openapi] def this(
         name: String,
         original: Option[Schema],
         description: Option[String]
     ) = this(name, original, description, None, None, None)
```

These should not be reachable outside of the `openapi` project since they are marked as `private`. My guess is that Mima flags them because they aren't actually `private` in the JVM class file sense, since they have to be accessible in the companion object for the `apply` method.

julienrf: Yes

harpocrates:
I believe I've finished applying feedback. 

Since this is logically two different changes (the cherry-picked addition of `default` and then the schema `$ref` fix), I've kept the two commits separate and independently passing Mima and tests. This way, you can pick pretty much whatever merge strategy you want.

julienrf (approved):

julienrf:
Thank you @harpocrates!

Issue #887, opened 2021-08-16
Option to skip default values in `JsonSchema` encoders 


harpocrates:
The current handling of default values in `JsonSchema` encoders means that all the fields that have default values will have those values populated in the encoder output. It would be nice to have some options such that the generated encoder would instead skip encoding a field if its value is the default.

This has two potential benefits:

  1. OpenAPI examples can omit optional fields
  2. Generated clients can send smaller payloads
  
The downside is that you end up needing to do an equality check on the value of the field (and if that field is a big nested class, that could be expensive). My main motivation is for 1 though, in which case the extra equality checks occur only once when docs are generated.

PR #889, opened 2021-08-17
Ujson, OpenAPI: option to skip default field values

harpocrates:
Adds a new `defaultFieldValuesEncoding` option to the `ujson`
interpreter and then enable that by default in the OpenAPI use of
`ujson`. This option makes the encoder check field values against the
default field value before serializing it. If they match, it skips the
field.

Implements #887


harpocrates:
Is this a reasonable feature/approach? If so, I'll add a couple tests. The option seemed useful to me mostly for making shorter and simpler serialized example values, but maybe it is unnecessary complexity.

openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
```diff
@@ -164,6 +164,69 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
...
+  }
+
+  /** Override this method to customize the behaviour of encoders produced by
+    * [[optFieldWithDefault]] when encoding a field value that corresponds to
+    * the specified default value.
+    * By default, it uses [[FieldDefaultValueEncoding.Always]].
+    *
+    * @see [[FieldDefaultValueEncoding]]
+    */
+  def defaultFieldValuesEncoding: FieldDefaultValueEncoding = FieldDefaultValueEncoding.Always
```
julienrf, line 228: I am not sure that it is easy for end-users to override this method. Could you add a test with that?

json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
```diff
@@ -336,7 +336,8 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     *   - Decoders fallback to the `defaultValue` if the field is absent from
     *     the decoded JSON object. They fail if the field is present but has an
     *     invalid value,
-    *   - Encoders always produce the field (even when it has the `defaultValue`),
+    *   - Encoders must emit the field if it is not `defaultValue`, but can
+    *     choose to omit it when it has the `defaultValue`
```
julienrf, line 340: We should probably mention the fact that the openapi interpreter is configurable with that regard.

julienrf (commented):
Overall, the changes look good, but I have a couple of remarks, see below.

Also, we should add some documentation about it in the page `json-schemas.md` (it seems we had nothing about optional fields btw…).

openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
harpocrates: Yep, will do. I think it should be easy to override. It would be tough to override in the OpenAPI interpreter (possibly even impossible right now, unless we thread through a duplicate option to that interpreter).

json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
harpocrates: The OpenAPI interpreter is not (right now) configurable at all. Should we make it? Only the ujson interpreter is configurable.

julienrf: > The OpenAPI interpreter is not (right now) configurable at all. Should we make it?

No, but we should point users to whatever is configurable.

harpocrates:
I think this is ready for review.

I'm really stumped about what to document and where to document it though. This is a _ujson_-specific option and although I suppose it could be generalized to other interpreters, I'm not sure there is value there but there is a cost if there are other JSON schema implementations in the wild. I don't see any section in the docs for _ujson_ either.

Along the way, I found and sort of addressed an issue that looks like bad interplay of _paradox-material-theme_ with a recent change made to _paradox_ (https://github.com/lightbend/paradox/pull/474).

  * Before: Screen Shot 2021-08-22 at 6 51 31 AM
  * After: Screen Shot 2021-08-22 at 7 15 46 AM

harpocrates:
Any idea when a release including this might occur? There isn't any urgent rush - I'm just asking so I can decide whether to rebase now or later some branch at work.

julienrf:
Hello! Yes, I want to cut a release very soon (hopefully this week). We are currently testing the new features that have been recently merged, just to double-check everything is ready for release.

Issue #890, opened 2021-08-18
Handling of NaN, infinities in double schema


harpocrates:
The algebra for `doubleJsonSchema` should probably decide how edge cases should be handled by encoders/decoders and the documentation interpreters should document that. It would be nice if:

  * encoders/decoders would roundtrip (or else throw exceptions)
  * encoder output matches the documentation

## Roundtrip client/server

As far as roundtrip goes, the `ujson` interpreter at least does not roundtrip.

```scala
object Example extends endpoints4s.ujson.JsonSchemas

def roundtrip(d: Double): endpoints4s.Validated[Double] = {
  val encoded: String = Example.doubleJsonSchema.stringCodec.encode(d)
  Example.doubleJsonSchema.stringCodec.decode(encoded)
}

roundtrip(Double.NaN) // Invalid(List(Invalid number value: "NaN"))
roundtrip(Double.PositiveInfinity) // Invalid(List(Invalid number value: "Infinity"))
roundtrip(Double.NegativeInfinity) // Invalid(List(Invalid number value: "-Infinity"))
```

I think we should fix this with special cases for decoding the strings `"NaN"`, `"Infinity"` and `"-Infinity"` back into doubles.

## OpenAPI

Here, I'm less certain about what to do. Technically, we should call out that the double is either a number or a string. OTOH, that will complicate schemas significantly. Perhaps this could be configurable?

PR #893, opened 2021-08-19
Ujson: decoding floating infinite and NaN values

harpocrates:
Ujson already supports encoding `Double.PositiveInfinity`,
`Double.NegativeInfinity`, and `Double.NaN` (as the strings
"Infinity", "-Infinity", and "NaN" respectively). This just adds support
to also decode those values back into doubles.

See #890

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/ujson/JsonSchemas.scala
@@ -356,15 +356,25 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
       val encoder = x => ujson.Num(x.doubleValue)
     }
 
+  private[this] object JsonDouble {
+     def unapply(json: ujson.Value): Option[Double] = json match {
+       case ujson.Num(x) => Some(x)
+       case ujson.Str("NaN") => Some(Double.NaN)
+       case ujson.Str("Infinity") => Some(Double.PositiveInfinity)
+       case ujson.Str("-Infinity") => Some(Double.NegativeInfinity)
+       case _ => None
+     }
+   }
+
   override def floatWithConstraintsJsonSchema(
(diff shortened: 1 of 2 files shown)


harpocrates:
Although I don't particularly like this solution, I like even less that we currently can't count on the client always deserializing server-serialized payloads.

I ran into this in a situation where we were packing JVM metrics using `case class Metric(name: String, value: Double)`. We had a rather large payload which the the client interpreter would sometimes fail to decode due to a a `NaN` somewhere in one of the metrics (which we ironically weren't using).

harpocrates:
I wonder why the 2.13 runs are timing out after 6 hours!

julienrf (approved):
This looks reasonable to me, thanks for the fix! I guess this should be fixed in ujson itself, though, what do you think? We are currently using an old version of ujson, maybe this has been fixed upstream?

harpocrates:
> guess this should be fixed in ujson itself, though, what do you think?

This was my first instinct, but I'm not sure how it could be. What exactly would we want? We could request

  1. that the string `"Infinity"` be parsed directly into `ujson.Num` (but that's a terrible idea since it might collide with legitimate string values).
  2. that the `ujson.Num` extractor match `ujson.Str("Infinity")` implicitly, but then the _order_ of the `ujson.Str` and `ujson.Num` cases in match on `ujson` could lead to subtly different behaviour (and warnings are bad on non-default extractors).

I think this code has to live somewhere where we have schema information about the fact we are expecting a `Double`/`Float`. Incidentally, _upickle_ lives in that space and [it handles the infinity + nan in its equivalent to `JsonSchema` too][0].

-----

The Mima issue OK to add an exception for, right (looks just like an artifact of how Scala compiles concrete members in traits)?

[0]: https://github.com/com-lihaoyi/upickle/blob/e7e0a2a71f8ef4431794fe4c04d14502bb0b72f7/implicits/src/upickle/implicits/Readers.scala#L52

Issue #794, opened 2021-04-19
Be able to compose endpoints at the algebra level


agboom:
Previously discussed in https://gitter.im/endpoints4s/endpoints4s?at=60548b4aa7dfe1372ee3921a

## Problem description

With the current algebras, it is not possible to compose endpoints. For example, the following can be declared in an abstract algebra, but its implementation can only exist in a concrete interpreter (Akka, http4s etc.):

```scala
def authenticatedEndpoint[A, B, AA, BB](endpoint: Endpoint[A, B]): Endpoint[AA, BB] // tuplers left out for brevity
```

In a sense, this method adds functionality to an existing endpoint and returns a new endpoint. This can then be passed to another similar function, hence composability. The current way to add functionality to an existing endpoint is to define a new method that accepts all the constituents (method, url, response etc.) and proxy these to the existing `endpoint` method. This is done for example in the [BasicAuthentication][basic-auth] algebra. For a single layer this is fine, but it poses a problem if you want to add multiple types of functionalities to an endpoint. For example, consider an endpoint that you want to have some custom errors, an authentication layer and some versioning in the url. At least in our use cases we have encountered this a few times.

It would be great to be able to compose it like `authenticatedEndpoint(endpointWithError(versionedEndpoint(endpoint(request, response))))`. However, this is not possible because the algebra types are abstract and/or do not have the necessary operations to achieve this.

[basic-auth]: https://github.com/endpoints4s/endpoints4s/blob/70e0b9a2530f81d60938572ecfdcb06342d7c851/algebras/algebra/src/main/scala/endpoints4s/algebra/BasicAuthentication.scala#L57

## Possible design directions

As discussed in Gitter, there are several possible design directions to consider. We have come up with two general directions, both with plusses and minuses. They are general directions in the sense that they're not fully working yet. We felt that sharing these early on would help to get a discussion going and have the highest chance to be accepted.  Both designs are pretty disruptive for the fundamental algebras, so that's one of the things we could discuss and see how to improve on that.

### Concrete datatype in algebra

One way to deal with this is to define a separate concrete datatype for `Request[A]`, `Response[A]` and/or `Endpoint[A, B]`, such as `RequestPayload[UrlP, BodyP, HeadersP]`. This allows you to access its constituents at the algebra level and thus implement methods such as the one above. We've experimented with this [in a separate branch][datatype]. 

[...]

PR #898, opened 2021-08-26
Add mapRequest and mapResponse to algebra and interpreters

bmeesters:
Based on https://github.com/endpoints4s/endpoints4s/pull/847 but only introduces `mapRequest` and `mapResponse`, still needs some unit tests.

There are few open questions from me:
* ~~Does this PR also need to include the `mapX` functions for the request and response properties (headers, url, ..)?~~ Done
* ~~How to deal with versioning? This changes some interpreters in some incompatible way, so they need a version bump. However we also have some other changes on the horizon. Do we want new releases for each of them?~~
* ~~How to deal with `Kleisli` in the http4s? I currently added an explicit conversion from the new `Endpoint` case class to the Kleisli, but I am not sure whether that is satisfactory.~~ Done
* ~~I also throw errors in the algebra to ensure it is backwards compatible. I target the 1.6.0 version, since I assume that will be the version where this will land.~~ Done

Partially solves https://github.com/endpoints4s/endpoints4s/issues/794

---

- [x] openapi interpreter
- [x] play-server interpreter
- [x] make client tests work
- [x] make documentation compile
- [x] play-client interpreter
- [x] akka-http-server interpreter (assigned to @bmeesters)
- [x] http4s-server interpreter
- [x] http4s-client interpreter
- [x] sttp-client interpreter
- [x] scalaj-client interpreter


openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
```diff
@@ -25,155 +23,194 @@ trait EndpointsWithCustomErrors
...
+      schemas = captureSchemas(documentedEndpoints),
+      securitySchemes = captureSecuritySchemes(documentedEndpoints)
     )
-    OpenApi(info, pathItems, components)
+    OpenApi(info, items, components)
   }
 
-  type Endpoint[A, B] = DocumentedEndpoint
+  case class Endpoint[A, B](
+      request: Request[A],
```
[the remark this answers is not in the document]
julienrf: I have just pushed a commit that addresses this point.

algebras/algebra/src/main/scala/endpoints4s/algebra/Endpoints.scala
```diff
@@ -54,6 +54,56 @@ trait EndpointsWithCustomErrors extends Requests with Responses with Errors {
...
+    * @group operations
+    * @return The endpoint with the mapped documentation.
+    */
+  def mapEndpointDocs[A, B](
+      endpoint: Endpoint[A, B],
+      func: EndpointDocs => EndpointDocs
+  ): Endpoint[A, B] = invalidInterpreterVersion
+
+  private def invalidInterpreterVersion: Nothing =
+    sys.error(s"Unsupported algebra version: 1.6.0. Please update your interpreter.")
```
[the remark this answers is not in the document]
julienrf: I have pushed a commit that addresses this point.

julienrf:
I have rebased the history because there were conflicts.

bmeesters:
Hello @julienrf thanks for getting this further. Is there anything left open that I should take on? I see you did some work though it is not entirely clear to me if there is any left-over work.

> It says that the changes are not backward source compatible. So, we have to relax the compatibility guarantees to Compatibility.BinaryCompatible, in the root build.sbt.

Can you explain to me why it is not source compatible? I assumed that adding new methods with implementations should not give that problem.

On a entirely unrelated side: love the work you are doing with the Scala Center that is posted on https://contributors.scala-lang.org/ :smile: Keep up the good work!

julienrf:
Hey @bmeesters, thank you for your kind words! I hope you enjoyed your vacations :)

> Is there anything left open that I should take on?

I am currently writing some documentation, and after that I want to add a few more tests, but if you want to help with the tests that would be great (it’s OK if you don’t :smiley:)

julienrf:
I just pushed the documentation, it looks like this:

image


I think this page https://endpoints4s.github.io/guides/custom-authentication.html could be rewritten to take advantage of the middleware, but I think it’s OK to leave it like that for now, and to rewrite it only when we will have the full power of the middleware.

I think the PR is ready.

bmeesters:
>  I hope you enjoyed your vacations :)

 Thanks I did :smile: 

> but if you want to help with the tests that would be great

Yes definitely! I suppose you already have an idea of what you want to test and how? I can help there if you want.

julienrf:
I started writing some more tests, but I see a lot of failures, unfortunately. I can’t continue right now, if someone wants to take over, you are welcome!

bmeesters:
The latest WIP commit on this branch refactors `Request` in akka-http-server so it also contains the other data. However, unless I missed something, the problem is that the new code does not get called. It seems the `directive` in `Request` is the only thing that matters for the actual implementation. So the problem might be in how directives are combine, and whether it is even possible in Akka to combine directives with different query parameters.

julienrf:
I guess in the akka-http-server interpreter, the `directive` of the type `Request[A]` should extract the “request context” (using `extractRequestContext`, and then write its own logic by calling `url`, `headers`, and `entity`.

bmeesters:
> That being said, could you maybe elaborate on the scenario you tested?

I used the unit test you added (and fails rightfully) `handle mapped endpoints` in `EndpointsTestSuite`, and the debugger to step through the break points to see what is happening. From what I can tell, the previously existing `uri` and the new `url` method do not get called in that scenario, so their implementation has no influence whatsoever on how the server works.

> In addRequestHeaders the added headers will be extracted from the incoming request only if the previous headers were present (that’s the way flatMap works).

Though I suppose this approach fails if no headers are present, the actual test already had an endpoint with headers before mapping it. So that should not be the case of the failure.

> Ideally, I think it would be great to try to extract all the query parameters and all the request headers from the incoming requests and report all the errors in the response, if any, or decode the request entity otherwise.

Yes, I agree. 

> I guess in the akka-http-server interpreter, the directive in type Request[A] should extract the “request context” (using extractRequestContext, and then write its own logic by calling url, headers, and entity.

I'll see if I can try that, not sure if it will be today though.

As always, I appreciate the pointers and the patience you show! I am aware this is also a time sink for you. For us this has some priority since we are actively working with 100+ micro-services where we want to leverage endpoints4s and this would be a good step in making a bit more ergonomic. But I fully understand if does not have the same priority for you since I know you are also working on other stuff. Feel free to take more time in responding. I see you are doing some late hours and definitely do not want you to feel pressured to work on this.

julienrf:
Hey!

A heads-up that it turned out to require more work than initially estimated. This is mainly due to the willing of providing better error messages. E.g., if you add a header to a request definition with `mapRequest(_.addHeader(...))`, I want bad requests to get a response listing _all_ the errors (not just the first error). This was the case so far as long as it was not possible to transform requests afterwards. But introducing high-level transformation operations required to restructure the code to still be able to appropriately report errors.

The good news is that I found a design that I am happy with for the play-server interpreter. It can be used as inspiration for the other interpreters. However, I have limited time to allocate to this issue, so any help would be appreciated. I have updated the PR description with a list of what remains to be done. Please leave a comment here if you start working on an item.

bmeesters:
Hello @julienrf,

Great that you have been able to push this forward. Sorry that I did not contribute as much lately as I would have wanted. Some things came in between and since I was the one originally requesting this I put my priorities for the time elsewhere. I am still planning to finish what I started. It will likely be two more weeks until I have a good chunk of time.

I will look at your changes and take them as guidance for the akka-http-server and see whether I can also take on any of the other interpreters.

algebras/algebra/src/main/scala/endpoints4s/algebra/Responses.scala
```diff
@@ -144,6 +144,14 @@ trait Responses extends StatusCodes with InvariantFunctorSyntax {
       headers: ResponseHeaders[B] = emptyResponseHeaders
   )(implicit tupler: Tupler.Aux[A, B, R]): Response[R]
 
+  /** Add the provided `headers` to the `response`.
+    */
+  def addResponseHeaders[A, H](
+      response: Response[A],
+      headers: ResponseHeaders[H]
+  )(implicit tupler: Tupler[A, H]): Response[tupler.Out] =
```
julienrf, line 152: I just noticed that there is a small issue with this operation. It adds a header to a `Response`. However a response can be a list of possible responses (when you use `orElse` between responses). In such a case, `addResponseHeaders` adds the header to all the possible responses. Not sure whether this is a problem or not.

bmeesters: I am not sure either, I am also not entirely sure when you would use this function. It is also relatively easy to workaround by using the `addResponseHeaders` before the `orElse` operation. Might be more of a caveat that needs to be documented (inline) than something that needs to be fixed. WDYT?

julienrf: True, this is not really a problem. I just found it surprising when I realized it. I am fine with keeping the method as it is.

julienrf (approved):

PR #900, opened 2021-08-30
fix http4s output encoding to UTF-8

hangilc:
Currently, in http4s server JsonEntities outputs String with default charset. This is a problem in my case, which is Windows OS in Japan with default charset of windows-31j (Shift-JIS). If I set environ variable JAVA_TOOL_OPTIONS to be '-Dfile.encoding=UTF8', output is encoded with UTF-8 as expected. However, if output is encoded properly without setting the environment variable, it would be better. The cause of this behavior seems to be encodeJsonResponse method of JsonEntities uses getBytes(), which encodes with default charset. By modifying it to getBytes("UTF-8"), it becomes to encode output with UTF-8 irrespective of default charset. I think the modified behavior is favorable because almost anyone expected Json strings are encoded with UTF-8. So, would you please consider incorporating thie modification to the library code?

### DIFF
--- a/http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
+++ b/http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
@@ -93,6 +93,6 @@ private object JsonEntities {
       endpoints: EndpointsWithCustomErrors
   )(encoder: Encoder[A, String]): endpoints.ResponseEntity[A] =
     EntityEncoder[endpoints.Effect, Chunk[Byte]]
-      .contramap[A](value => Chunk.array(encoder.encode(value).getBytes("UTF-8")))
+      .contramap[A](value => Chunk.array(encoder.encode(value).getBytes(http4s.Charset.`UTF-8`.nioCharset)))
       .withContentType(`Content-Type`(MediaType.application.json))
 }
--- a/http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
+++ b/http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
@@ -93,6 +93,6 @@ private object JsonEntities {
       endpoints: EndpointsWithCustomErrors
   )(encoder: Encoder[A, String]): endpoints.ResponseEntity[A] =
     EntityEncoder[endpoints.Effect, Chunk[Byte]]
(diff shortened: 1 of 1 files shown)


http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
```diff
@@ -93,6 +93,6 @@ private object JsonEntities {
       endpoints: EndpointsWithCustomErrors
   )(encoder: Encoder[A, String]): endpoints.ResponseEntity[A] =
     EntityEncoder[endpoints.Effect, Chunk[Byte]]
-      .contramap[A](value => Chunk.array(encoder.encode(value).getBytes()))
+      .contramap[A](value => Chunk.array(encoder.encode(value).getBytes("UTF-8")))
```
julienrf, line 96: Would it be OK to use a predefined value instead of a string literal?

```suggestion
      .contramap[A](value => Chunk.array(encoder.encode(value).getBytes(http4s.Charset.`UTF-8`.nioCharset)))
```

julienrf (approved):
Thank you for fixing this issue! I have a minor suggestion, let me know what you think about it.

http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
hangilc: Yes, I agree that would be better.

julienrf: @hangilc, I think you didn’t push your changes

hangilc: Sorry. I pushed the changes.

julienrf:
You need to format the code by running `scalafmt` in sbt.

http4s/server/src/main/scala/endpoints4s/http4s/server/JsonEntities.scala
harpocrates: In the name of consistency with the rest of the project, shouldn't this be `StandardCharsets.UTF_8`? Functionally speaking, they're all equivalent.

julienrf: I’m happy with either :)

PR #916, opened 2021-09-28
Bump circe to 0.14.1

MasseGuillaume:


MasseGuillaume:
No rush I’m just keeping my libraries up to date. It’s the version caliban (a graphql client) is using.

PR #925, opened 2021-10-12
Add ability to create int based enum json schema

ytalashko:
It may be useful to allow creating value based enums, not only based by strings (at least the ones backed by integers).
There is no open issue for this, but I have faced such a need during my recent work experience.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ ThisBuild / sonatypeProjectHosting := Some(
 // We want to keep binary compatibility as long as we can for the algebra,
 // but it is OK to publish breaking releases of interpreters. So,
 // interpreter modules may override this setting.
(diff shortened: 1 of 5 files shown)


julienrf:
Thank you, I think it’s good for now to not introduce many more methods unless we clearly identified a need for them. I am happy with `intEnumeration` :)

julienrf:
Could you please relax the compatibility intention?

https://github.com/endpoints4s/endpoints4s/blob/8209faec7aa7539c31b3a9232bb85abd645c18c1/build.sbt#L39

Set it to `Compatibility.BinaryCompatible`.

Then, could you please squash your commits?

ytalashko:
> Then, could you please squash your commits?

Just to clarify, do you mean squash merge (when merging PR)? Or to squash and force push (override commits history) in the PR branch?
Thanks!

Issue #287, opened 2019-05-19
Add a Scala.js client interpreter based on Fetch


julienrf:
The [Fetch](https://fetch.spec.whatwg.org/) API has the advantage of supporting backpressure for streamed responses, through the [ReadableStream](https://streams.spec.whatwg.org/#example-rs-push-backpressure) API.

A drawback is that `Fetch` might be less supported by old web browsers than `XMLHttpRequest`.

PR #934, opened 2021-10-26
Scala.js client interpreter based on Fetch

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/287


ptrdom:
Looked through the spec again, I now think that the use of `WritableStream` can be avoided, but either https://github.com/scala-js/scala-js-dom/issues/622 needs to be implemented, or endpoints4s needs its own Fetch wrapper with a facade that exposes constructor of the `ReadableStream`.

armanbilge:
Also FYI scala-js-dom recently bumped to v2 but you are currently on v1 here:
- https://github.com/endpoints4s/endpoints4s/pull/931

ptrdom:
As for point raised by @armanbilge, I guess we will have to see how the PRs/releases align and decide on the endpoint4s possible temporary need of a custom `Fetch` facade.

armanbilge:
Also not sure if of its interest, but I see you already have an http4s interpreter here. So you could just use that with the http4s `FetchClient` in [http4s-dom](https://github.com/http4s/http4s-dom).

ptrdom:
@armanbilge Current http4s client interpreter is JVM one, so `http4s-dom` could be used to implement a new one for JS. As for this PR, I think the goal is to bring in as little dependencies as possible, ideally only depending on `scala-js-dom`, same as with XHR interpreter.

armanbilge:
> Current http4s client interpreter is JVM one

You don't need another project, you just need to enable a cross-build for it. Http4s has an identical client API for JVM/JS.

> As for this PR, I think the goal is to bring in as little dependencies as possible

👍 that's fair. But then you run into problems with `WritableStream`, piping, etc.

armanbilge:
Oh, also because you are interacting with `Future` here, you should be aware of this project:
https://github.com/scala-js/scala-js-macrotask-executor

The next version of Scala.js will be issuing a warning about this:
- https://github.com/scala-js/scala-js/pull/4556

ptrdom:
> 👍 that's fair. But then you run into problems with `WritableStream`, piping, etc.

That will not be an issue with a facade that exposes `ReadableStream` constructor, it should allow an implementation based on [this](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#fetch_stream) example.

Did not know about the `Future` executor changes, thanks for pointing them out! Will need to also update the XHR interpreter.

armanbilge:
> That will not be an issue with a facade that exposes ReadableStream constructor, it should allow an implementation based on this example.

I'm still confused what you hope to accomplish with that, since browsers don't supported `ReadableStream` bodies, but whatever works :)

ptrdom:
Oh, it is just that `type Chunks[A]` could be a `ReadableStream[A]`, so chunked response handling could be implemented more nicely then.

armanbilge:
I see, and then you would consume the `ReadableStream` before calling `Fetch`?

fetch/client/src/main/scala/endpoints4s/fetch/ChunkedEntities.scala
```diff
@@ -0,0 +1,91 @@
...
+import org.scalajs.dom.experimental.ReadableStream
+
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.scalajs.js.Thenable.Implicits._
+import scala.scalajs.js.typedarray.Uint8Array
+
+trait ChunkedEntities extends algebra.ChunkedEntities with EndpointsWithCustomErrors {
+
+  type Chunks[A] = ChunksSource[A]
```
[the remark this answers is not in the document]
ptrdom: `type Chunks[A]` can be `ReadableStream[A]`.

fetch/client/src/main/scala/endpoints4s/fetch/ChunkedEntities.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  def textChunksResponse: ResponseEntity[Chunks[String]] =
+    chunkedResponseEntity(byteString => Right(byteString))
+
+  def bytesChunksResponse: ResponseEntity[Chunks[Array[Byte]]] =
+    chunkedResponseEntity(byteString => Right(byteString.getBytes))
+
+  //TODO complicated, needs some thought on how to write chunks to a body with native, well supported APIs
+  def textChunksRequest: RequestEntity[Chunks[String]] = ???
+  def bytesChunksRequest: RequestEntity[Chunks[Array[Byte]]] = ???
```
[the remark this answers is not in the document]
ptrdom: Browsers do not implement chunked requests yes, so implementation could be done, but it will not work until some browser eventually implements it.

fetch/client/src/test/scala/endpoints4s/fetch/future/FetchClientEndpointsTest.scala
```diff
@@ -0,0 +1,75 @@
...
+    with algebra.SumTypedEntitiesTestApi
+    with algebra.circe.JsonFromCirceCodecTestApi
+    with JsonEntitiesFromCodecs
+    with algebra.circe.JsonEntitiesFromCodecs
+    with ChunkedJsonEntities
+    with ChunkedJsonEntitiesTestApi
+    with algebra.circe.ChunkedJsonEntitiesTestApi
+
+//TODO needs Scala.js test setup
+class FetchClientEndpointsTest
```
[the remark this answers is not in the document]
ptrdom: Only issue with tests seemed to be chunked responses, but actually both WireMock (http://wiremock.org/docs/simulating-faults/#chunked-dribble-delay) and MockServer (`httpResponse.connectionOptions.chunkSize`) can produce multiple chunks when they are of same size

ptrdom:
Discovered one serious issue with chunked entity implementation - chunks sent by server side are not directly translated to chunks in `ReadableStream`, meaning that one chunk in `ReadableStream` might actually be `1 + n` server side chunks. Also, `ReadableStream` does not expose any chunk extensions. I then checked what is the XHR behavior, and it basically appends chunks to `responseText` in same manner. With current server interpreter implementation, where each chunk translates to a single stream element, I do not think chunked entities can be implemented for browser clients. I think there are two ways to proceed from here:

1.  Work on custom protocol, either something similar to gRPC streaming implementation or https://github.com/eBay/jsonpipe, so that endpoints4s client and server interpreters can communicate.
2. Put this development on hold until (if ever) something new comes up in the spec.

Thoughts?

julienrf:
Hello! I saw a lot of activity in this PR, this is really exciting! I will have a look before the end of the week.

ptrdom:
> I think there are two ways to proceed from here:

Third way - WebSockets 😎

ptrdom:
Some additional thoughts on the `WebSocket` route. Having all the findings in mind, I think this could be the way to proceed.

This PR gets finished without chunked entities, but with tests, then later on the test implementation could be used for `XHR` client too. Then `WebSocket` implementation could be done in its own PR in one of two ways - new algebra or new interpreters. Choice here depends on what we want to achieve - new `WebSocket` interpreters for client and server would not require changes to core, but library users would then have to use two endpoint implementations to make use of `WebSockets`, for example Scala.js client would be composed of `XHR` or `Fetch` client interpreter and a `WebSocket` interpreter implementations, so that is not as neat to use. New algebra, with something like `websocketEndpoint`, would be neater for library users, but that requires changes to the core. In both cases, chunked entity implementation would not be altered, but having done the investigation for this PR, I feel that it should have been done with `WebSockets` rather than chunked transfer encoding. This would have been similar to the route that the `Lagom` framework `ServiceClient` took - if either request or response is a stream, the underlying implementation uses a `WebSocket`, rather than a `HTTP` request. Having said that, rewriting chunked entity implementation would be a major breaking change, and likely not worth it.

julienrf:
From what I understand in this discussion, it turns out that it is _not possible_ to implement a `ChunkedEntities` client interpreter for browsers.

I saw two main issues: the fact that browsers don’t support `ReadableStream` in request bodies (but this is something that could be worked around, right?), and the fact that the current specification of `ChunkedEntities` in endpoints4s requires that no reframing happens so that one frame sent by a server is always seen as exactly one frame received by the client. If I understood correctly, browsers do reframe what they receive.

It seems to me that the way `ChunkedEntities` currently works is brittle and it should be improved. It looks like jsonpipe might be a better approach.

Regarding the usage of WebSockets: if that allows us to implement a client interpreter for the algebra `ChunkedEntities`, that’s great! Introducing a new algebra could also be useful on its own, but that would not solve the current issue that `ChunkedEntities` can’t be consumed by the Scala.js client interpreters.

ptrdom:
Yes, the main issue with chunking implementation is that chunks defined by the protocol are reframed by the browsers. I have a feeling that even if `Fetch` request in future accepts `ReadableStream`, it will reframe and break the endpoints4s server interpreter.

My proposal to not touch current `ChunkedEntities` implementation was based on perceived importance to stay backwards compatible. There might very well be users of this library that use this algebra with its current implementation for non-browser communication and are happy with how it works. If it was okay to break it, I would propose reworking it through WebSockets for the long-term health of this project, and this would result in changes both to core and to interpreters. I could definitely help out on core changes and Akka HTTP, Play, browser interpreters.

New algebra would allow backwards compatibility and even eventual deprecation of `ChunkedEntities`, maybe it could be called `StreamedEntities` to emphasize what it is really about. Ideally there would not be a split between `endpoint` and `websocketEndpoint`/`streamedEndpoint`, but interpreter implementation could pick the request implementation based on what is being transported - this is the approach https://github.com/lagom/lagom took, which I described before. This feels like the best in-between solution.

Full algebra could be implemented through WebSockets, not only `ChunkedEntities`, and that would be the best way to go if avoiding core changes is important, but library users will have to use a mix of interpreters if they only want to use WebSockets for `ChunkedEntities`. It does feel as a bit confusing and more short-sighted solution.

Overall this is not an easiest choice to make, there are tradeoffs in each approach, @julienrf, as I understand you are the one responsible for the direction of this project, so ultimately the decision falls on you.

julienrf:
Thanks a lot for writing up this detailed message!

I agree that we can’t change the behavior of `ChunkedEntities`. What I meant by following a different approach was to introduce a new algebra (`StreamedEntities` sounds like a good name), but not change the current behavior (at least, not before endpoints4s 2.x, but this is not really at the horizon yet).

I realize that I wrote something wrong when I said “this is great if we can implement the `ChunkedEntities` algebra with WebSocket”. This is obviously wrong since the algebra `ChunkedEntities` is about how the entities of regular endpoints (not WebSocket) are transferred.

> Ideally there would not be a split between `endpoint` and `websocketEndpoint`/`streamedEndpoint`, but interpreter implementation could pick the request implementation based on what is being transported - this is the approach https://github.com/lagom/lagom took, which I described before. This feels like the best in-between solution.

I don’t really understand this part, do you mind elaborating? I wish we wouldn’t have to introduce a `streamedEndpoint` constructor, and we could instead model streaming at the level of entities (like the current `ChunkedEntities` does).

ptrdom:
> > Ideally there would not be a split between `endpoint` and `websocketEndpoint`/`streamedEndpoint`, but interpreter implementation could pick the request implementation based on what is being transported - this is the approach https://github.com/lagom/lagom took, which I described before. This feels like the best in-between solution.
> 
> I don’t really understand this part, do you mind elaborating? I wish we wouldn’t have to introduce a `streamedEndpoint` constructor, and we could instead model streaming at the level of entities (like the current `ChunkedEntities` does).

Of course, so overall there are 4 endpoint types, and based on request or response being a stream, different protocol could be chosen:

- `def unary[A, B](request: Request[A]): Response[B]` - HTTP request
- `def clientStreaming[A, B](request: Request[Stream[A]]): Response[B]` - WebSocket
- `def serverStreaming[A, B](request: Request[A]): Response[Stream[B]]` - WebSocket
- `def bidirectionalStreaming[A, B](request: Request[Stream[A]]): Response[Stream[B]]` - WebSocket

Now that I have looked through core implementation again, maybe this could be achieved without changes to existing algebras - using flexible definition of `Request` and `RequestEntity` in interpreters, that would allow `RequestEntity` to change the protocol, would be enough. What do you think about this?

julienrf:
> Now that I have looked through core implementation again, maybe this could be achieved without changes to existing algebras - using flexible definition of `Request` and `RequestEntity` in interpreters, that would allow `RequestEntity` to change the protocol, would be enough. What do you think about this?

IMHO, if we are doing streaming on top of regular HTTP endpoints, there should be no need to use a different constructor than `endpoint`. As you noticed, the fact that the request or response entities are streamed should be modeled by `RequestEntity` or `ResponseEntity`.

ptrdom:
> As you noticed, the fact that the request or response entities are streamed should be modeled by `RequestEntity` or `ResponseEntity`.

Thank you for confirming this, I think it is clear how to proceed then.

ptrdom:
@julienrf Would it be okay if I made changes to client test setup in this PR? To support JS environment, I would need to do two things:

1.  Remove any blocking, which means basing tests on  `AsyncWordSpec` and modifying test implementations to fit its requirements.
2. Move WireMock to Docker, configure all stubs ahead on the time in `.json` configuration file and modify test implementations to not do any stubbing.

Not difficult, but not trivial either, so I would understand if it should be done in its own PR. But then this PR either needs to not come with tests, or needs custom test implementation, which is basically a JS compatible copy of what is already there and it will soon be thrown out.

julienrf:
That would be a lot of changes in a single PR. I have a preference for splitting in two PRs, but to be honest I don’t want to make things to complicated for you because I am super happy that you are willing to help with the tests ;)

Regarding Wiremock, I was considering removing it (and replacing it with an Akka-HTTP server or http4s server, whatever works best) because it has some limitations that made it impossible to implement some test cases (IIRC). But feel free to suggest an approach that works better for you!

ptrdom:
No worries, wanting to split it up is completely reasonable :) I will then implement JS environment tests with existing client test copypasta and WireMock running on Docker, this approach should require the least effort. Only thing I wanted to split in existing implementation with this PR was URL encoding tests - they are located in a trait that requires WireMock although test cases only depend on `encodeUrl`.

Regarding replacement of WireMock, I think the issue was with `ChunkedEntities` tests, http://wiremock.org/docs/simulating-faults/#chunked-dribble-delay should allow allow producing chunks of even size, but looking forward to `StreamedEntities`, WireMock definitely does not have WebSockets. Both Akka-HTTP and http4s are good choices, my personal preference is the former, I was just thinking how would it be ran for tests? `sbt run` a separate app with stubs before starting tests or is there some better way?

algebras/algebra/src/test/scala/endpoints4s/algebra/client/EncodeTestSuite.scala
```diff
@@ -0,0 +1,172 @@
...
+      //FIXME
+      // This assertion looks suspicious - space should be encoded as + only if content type is application/x-www-form-urlencoded.
+      // URLEncoder is known to not encode according to the standard - https://github.com/sksamuel/elastic4s/issues/2352.
+      // JS assertion should be the only one that is correct.
+      VM.current match {
+        case VM.JS =>
+          assert(encoded == "?q=foo%20bar%2Fbaz")
+        case VM.JVM =>
+          assert(encoded == "?q=foo+bar%2Fbaz" || encoded == "?q=foo%20bar/baz")
+      }
```
ptrdom, line 48: I think this should be fixed in a separate issue

fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
```diff
@@ -0,0 +1,3 @@
+package endpoints4s.fetch
+
+case class EndpointsSettings(host: Option[String])
```
julienrf, line 3: I suggest we design it to be “evolvable” from the ground start. This means, it has to be a plain old class like this one for example:

https://github.com/endpoints4s/endpoints4s/blob/0ecaa1d861f631465fc797d195b91372f009cf54/algebras/algebra/src/main/scala/endpoints4s/algebra/Endpoints.scala#L64-L134

fetch/client/src/main/scala/endpoints4s/fetch/FetchUrl.scala
```diff
@@ -0,0 +1,3 @@
+package endpoints4s.fetch
+
+case class FetchUrl(val underlying: String) extends AnyVal
```
julienrf, line 3: What is the purpose of this class?

fetch/client/src/main/scala/endpoints4s/fetch/TextDecoder.scala
```diff
@@ -0,0 +1,11 @@
+package endpoints4s.fetch
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+import scala.scalajs.js.typedarray.ArrayBufferView
+
+@js.native
+@JSGlobal
+class TextDecoder(utfLabel: js.UndefOr[String] = js.undefined) extends js.Object {
```
julienrf, line 9: Should we make this private? Or `private[fetch]`?

fetch/client/src/main/scala/endpoints4s/fetch/TextEncoder.scala
```diff
@@ -0,0 +1,12 @@
+package endpoints4s.fetch
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+import scala.scalajs.js.typedarray.Uint8Array
+
+//TODO could be used for chunked requests
+@js.native
+@JSGlobal
+class TextEncoder(utfLabel: js.UndefOr[String] = js.undefined) extends js.Object {
```
julienrf, line 10: Same here regarding privateness.

fetch/client/src/main/scala/endpoints4s/fetch/future/Endpoints.scala
```diff
@@ -0,0 +1,31 @@
...
+import endpoints4s.fetch
+
+import scala.concurrent.Future
+import scala.concurrent.Promise
+
+trait Endpoints extends fetch.Endpoints with EndpointsWithCustomErrors
+
+trait EndpointsWithCustomErrors extends fetch.EndpointsWithCustomErrors {
+
+  type Result[A] = Future[A]
```
julienrf, line 12: To be honest, this `type Result[A]` is not mandatory to have in Scala.js interpreters. I introduced it in the `xhr-client` interpreter to give more control to the end-users (I think `scala.concurrent.Future` introduces significant overhead in Scala.js, so I wanted to provide an escape hatch), but it’s fine to not introduce the abstract type `Result[A]` in the first place in the fetch interpreter, if you think Fetch users will most likely all choose the same `Result` type, ultimately.

fetch/client/src/test/scala/endpoints4s/fetch/future/BasicAuthenticationTestApi.scala
```diff
@@ -0,0 +1,19 @@
+package endpoints4s.fetch.future
+
+import endpoints4s.algebra
+import endpoints4s.algebra.EndpointsTestApi
+
+trait BasicAuthenticationTestApi extends EndpointsTestApi with algebra.BasicAuthentication {
```
julienrf, line 6: Don’t we already have a `TestApi` class that we should reuse here?

algebras/algebra/src/test/scala/endpoints4s/algebra/client/EncodeTestSuite.scala
```diff
@@ -0,0 +1,172 @@
+package endpoints4s.algebra.client
+
+import java.util.UUID
+
+import endpoints4s.algebra
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+import endpoints4s.{Invalid, Valid}
+
+trait EncodeTestSuite[T <: algebra.Endpoints] extends AnyWordSpecLike with Matchers {
```
julienrf, line 10: This is not super important, but I think a name like `UrlEncodingTestSuite` or `UrlsTestSuite` would be clearer.

fetch/client/src/test/resources/wiremock/mappings/mappings.json
```diff
@@ -0,0 +1,296 @@
+{
```
julienrf, line 1: I wonder if this file won’t be too difficult to maintain?

julienrf (commented):
Thank you @ptrdom this is a great contribution!

In addition to my comments below, I think we need to update the documentation:

https://github.com/endpoints4s/endpoints4s/blob/master/documentation/manual/src/main/paradox/algebras-and-interpreters.md

You may want to add a page like this one too:

https://github.com/endpoints4s/endpoints4s/blob/master/documentation/manual/src/main/paradox/interpreters/scalajs-web.md

We should also include the fetch interpreter API in the API documentation:

https://github.com/endpoints4s/endpoints4s/blob/0ecaa1d861f631465fc797d195b91372f009cf54/documentation/build.sbt#L53-L73

Last, it would be good to check that it compiles with Scala 3 by including it here:

https://github.com/endpoints4s/endpoints4s/blob/0ecaa1d861f631465fc797d195b91372f009cf54/.github/workflows/ci.yml#L48

fetch/client/src/main/scala/endpoints4s/fetch/Endpoints.scala
```diff
@@ -0,0 +1,304 @@
...
+
+  type Response[A] = js.Function1[FetchResponse, Option[ResponseEntity[A]]]
+
+  implicit lazy val responseInvariantFunctor: InvariantFunctor[Response] =
+    new InvariantFunctor[Response] {
+      def xmap[A, B](fa: Response[A], f: A => B, g: B => A): Response[B] =
+        xhr => fa(xhr).map(mapResponseEntity(_)(f))
+    }
+
+  type ResponseEntity[A] = FetchResponse => Future[Either[Throwable, A]]
```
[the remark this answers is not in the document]
ptrdom: Makes sense, will make it native.

fetch/client/src/main/scala/endpoints4s/fetch/FetchUrl.scala
ptrdom: Since `type Request[A] = A => (FetchUrl, FetchRequestInit)`, wanted to make it something stronger than a `String` or a type alias. What approach would you prefer here?

fetch/client/src/main/scala/endpoints4s/fetch/TextDecoder.scala
ptrdom: Will remove, no longer needed because `ChunkedEntities` will not be implemented.

fetch/client/src/main/scala/endpoints4s/fetch/TextEncoder.scala
ptrdom: Will remove, no longer needed because `ChunkedEntities` will not be implemented.

fetch/client/src/main/scala/endpoints4s/fetch/future/Endpoints.scala
ptrdom: I followed this structure to allow both `Future` and `js.Thenable` endpoint results, just how `XHR` did it. I guess there is no harm in leaving it like that, unless I misunderstood your point?

fetch/client/src/test/scala/endpoints4s/fetch/future/BasicAuthenticationTestApi.scala
ptrdom: Issue with previous client test implementation is that same endpoint is re-mocked with `WireMock` to produce a different result. Because I had to set up the mocks before launching tests, I needed new endpoints for each unique result.

fetch/client/src/test/resources/wiremock/mappings/mappings.json
ptrdom: I could think of ways to implement this to be more maintainable, for example by interacting with `WireMock` through HTTP API to submit JSON stubs from tests, but I am hesitant to commit more effort to this because it will be all made redundant rather quickly with the implementation of https://github.com/endpoints4s/endpoints4s/issues/945.

fetch/client/src/main/scala/endpoints4s/fetch/FetchUrl.scala
julienrf: OK I see. No problem but I would rather make this class `private[fetch]` then, if that works for you.

julienrf: Hmm, I don’t think it’s possible to make it private since it will part of the public signature of request constructors… It’s OK to leave it as it is. Just remove the `val` before the `underlying` field.

fetch/client/src/main/scala/endpoints4s/fetch/future/Endpoints.scala
julienrf: True, there is no harm. I just saw an opportunity to simplify things.

fetch/client/src/test/resources/wiremock/mappings/mappings.json
julienrf: OK, let’s keep it for now and see how it goes after #945 is solved.

julienrf:
I have a doubt now: the initial goal was to support `ChunkedEntities`, but we concluded that this is not possible with `Fetch` because of reframing. Then, what will be the value of using `fetch-client` over `xhr-client`?

ptrdom:
Fetch gives benefits if we decide to fix up `ChunkedEntities`:

1. XHR `responseText` grows endlessly, so browser could technically OOM with a long running `ChunkedEntities` response.
2. XHR appends chunks to `responseText`, while Fetch gives `ReadableStream`, which allows more efficient and idiomatic consumer implementation.

Other than those points - this PR is mostly complete and we might as well give users the choice.

@julienrf Do you still have doubts over proceeding with this PR?

julienrf:
Great, so IIUC the real value of `fetch-client` over `xhr-client` will become more apparent once we fix the issue with `ChunkedJsonEntities`?

ptrdom:
Yes, just a small correction - `ChunkedEntities` in general, because with the correct approach taken to fix them it should not matter what they transport.

julienrf:
Do you mean that even when streaming binary content we should split data in explicit “messages” prefixed by their length?

In any case, I’m in favor of merging the PR, just in case this wasn’t clear :)

ptrdom:
> Do you mean that even when streaming binary content we should split data in explicit “messages” prefixed by their length?

In a way - what I had in mind was that it does not matter what the chunk content is if we prefix the messages with their length, it can be `Chunk[Byte]`, `Chunk[Array[Byte]]`, `Chunk[Int]`, `Chunk[String]`, `Chunk[JsObject]` or `Chunk[JsArray]` - it will work.

> In any case, I’m in favor of merging the PR, just in case this wasn’t clear :)

Thank you for the confirmation, I was wondering if I should continue with it 😅

fetch/build.sbt
```diff
@@ -0,0 +1,75 @@
...
+  project
+    .in(file("client"))
+    .enablePlugins(ScalaJSPlugin)
+    .configure(_.disablePlugins(ScoverageSbtPlugin))
+    .settings(
+      publishSettings,
+      `scala 2.12 to 2.13`,
+      name := "fetch-client",
+      version := "3.1.0+n",
+      versionPolicyIntention := Compatibility.BinaryCompatible,
```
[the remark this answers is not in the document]
ptrdom: Do I correctly understand that either this or the following fix to this file a required, not both?

ptrdom:
@julienrf Scala 3 compilation fails with the same error for both XHR and Fetch clients - `[error] error while loading package, Missing dependency 'Add -Ytasty-reader to scalac options to parse the TASTy in \endpoints4s\algebras\algebra\.js\target\scala-3.0.2\classes\endpoints4s\package.class', required by \endpoints4s\algebras\algebra\.js\target\scala-3.0.2\classes\endpoints4s\package.class`. Maybe it would be a better idea to figure this out for both Scala.js interpreters in a separate issue?

fetch/build.sbt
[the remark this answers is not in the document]
julienrf: Good point! Actually, just setting `versionPolicyIntention := Compatibility.None` will work (without setting `versionPolicyCheck / skip`).

fetch/build.sbt
```diff
@@ -0,0 +1,73 @@
...
+    .settings(
+      publishSettings,
+      `scala 2.12 to 2.13`,
+      name := "fetch-client",
+      version := "3.1.0+n",
+      versionPolicyIntention := Compatibility.None,
+      //disable coverage for scala.js: https://github.com/scoverage/scalac-scoverage-plugin/issues/196
+      coverageEnabled := false,
+      libraryDependencies ++= Seq(
+        "org.scala-js" %%% "scalajs-dom" % "1.2.0",
```
julienrf, line 17: Should we use version 2.0.0?

julienrf:
> Scala 3 compilation fails

Did you set the setting `scala 2.12 to dotty` to the project?

https://github.com/endpoints4s/endpoints4s/blob/d8ed723fd115b46a0c0917fca97e6d1ad4b24833/project/EndpointsSettings.scala#L67-L70

fetch/build.sbt
ptrdom: We can, I do not think that using the latest version available could cause issues.

ptrdom:
> Did you set the setting `scala 2.12 to dotty` to the project?

That was it, thanks!

fetch/build.sbt
ptrdom: This also must be done for Scala 3 support.

ptrdom:
@julienrf Finally got the CI all green, could you check if this PR is good to go?

ptrdom:
> @julienrf Finally got the CI all green, could you check if this PR is good to go?

Particularly the documentation, I was wondering how to best split the naming for Scala.js between XHR and Fetch.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       with:
         java-version: adopt@1.11
     - run: npm install jsdom@12.2.0
+    - run: docker-compose -f ./fetch/client/src/test/resources/wiremock/docker-compose.yml up -d
```
julienrf, line 22: I am not familiar with this kind of setup. When is the server going to be shut down?

fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
```diff
@@ -0,0 +1,37 @@
+package endpoints4s.fetch
+
+import endpoints4s.Hashing
+
+import scala.annotation.nowarn
+
+final class EndpointsSettings(val host: Option[String]) extends Serializable {
```
julienrf, line 7:
```suggestion
final class EndpointsSettings private (val host: Option[String]) extends Serializable {
```

The expected way to build an `EndpointsSettings` with a host would be:

~~~ scala
EndpointsSettings().withHost(Some("..."))
~~~

json-schema/json-schema-generic/src/main/scala-3/endpoints4s/generic/JsonSchemas.scala
```diff
@@ -119,7 +119,7 @@ trait JsonSchemas extends algebra.JsonSchemas:
     * @param  defaultValues  List of the possible default values of the record fields (e.g.,
     *                        `(Some(42), None)`)
     */
-  private[generic] inline def summonRecord[
+  inline def summonRecord[
```
julienrf, line 122: Why were these changes necessary? I wish these methods could be kept `private`.

julienrf (commented):
Thank you @ptrdom for all this work, especially the cleaning up you did on the tests!

I have left a few more questions before we can merge. Also, I would like to first merge #898 which was started earlier and was a huge effort. You will have to rebase and add a few things after we merge it (you can copy what has been done in the xhr interpreter).

fetch/client/src/main/scala/endpoints4s/fetch/TextEncoder.scala
julienrf: Do we keep it, ultimately?

json-schema/json-schema-generic/src/main/scala-3/endpoints4s/generic/JsonSchemas.scala
ptrdom: Scala 3 compilation was complaining in Fetch JsonSchema tests about failing to summon these methods. I think other interpreters that have these tests are not compiled for Scala 3 yet.

.github/workflows/ci.yml
ptrdom: I do not think that has to be managed - once CI is done, machine gets shutdown together with any containers it spawned.

json-schema/json-schema-generic/src/main/scala-3/endpoints4s/generic/JsonSchemas.scala
julienrf: Indeed, this is due to https://github.com/lampepfl/dotty/issues/11706 but that issue does not seem to have been prioritized.

julienrf: Let’s come back to this Scala 3 issue when we become ready to publish Scala 3 artifacts…

fetch/client/src/test/scala/endpoints4s/fetch/future/FetchClientEndpointsJsonSchemaTest.scala
```diff
@@ -0,0 +1,44 @@
...
+  val client: TestJsonSchemaClient = new TestJsonSchemaClient(
+    EndpointsSettings()
+      .withHost(Some("http://localhost:8080"))
+  )
+
+  def call[Req, Resp](
+      endpoint: client.Endpoint[Req, Resp],
+      args: Req
+  ): Future[Resp] = endpoint(args)
+}
```
julienrf, line 44: Are you sure some tests are executed here?

fetch/client/src/main/scala/endpoints4s/fetch/future/Endpoints.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  override def mapEndpointResponse[A, B, C](
+      currentEndpoint: Endpoint[A, B],
+      func: Response[B] => Response[C]
+  ): Endpoint[A, C] = endpoint(currentEndpoint.request, func(currentEndpoint.response))
+
+  override def mapEndpointDocs[A, B](
+      currentEndpoint: Endpoint[A, B],
+      func: EndpointDocs => EndpointDocs
+  ): Endpoint[A, B] = currentEndpoint
```
julienrf, line 47: It looks like it would be possible to implement this in the parent trait, no?

fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  @nowarn("cat=unused")
+  private[this] def copy(
+      host: Option[String] = host
+  ): EndpointsSettings =
+    new EndpointsSettings(
+      host
+    )
+
+  def withHost(host: Option[String]): EndpointsSettings = {
```
julienrf, line 29: Should we name it `baseUri`?

fetch/client/src/test/scala/endpoints4s/fetch/future/FetchClientEndpointsJsonSchemaTest.scala
julienrf: Sorry to be nitpicking again here, but it seems that `basicAuthSuite()` is redundant with what we have in `FetchClientEndpointsTest`, no?

ptrdom: It is alright, I am just copypasting things from XHR and Akka HTTP clients, got to remember to fix these same issues there too.

julienrf (approved):
Thank you @ptrdom!

PR #939, opened 2021-11-04
Cross-build http4s-client for Scala.js

armanbilge:
Addresses https://github.com/endpoints4s/endpoints4s/issues/287 when paired with the http4s-dom fetch client backend:
https://http4s.github.io/http4s-dom/

I didn't add any tests here: the http4s-client API is identical across JVM/JS (which you are testing on JVM) and the http4s-dom fetch backend is itself tested.

_If_ you wanted to test this yourself, would require some non-trivial build machinery: browsers, browser drivers, some servers external to the test itself (e.g., started by sbt), etc. etc. This is how we test http4s-dom.


http4s/build.sbt
```diff
@@ -39,16 +44,27 @@ val `http4s-client` =
       version := "5.0.0+n",
       versionPolicyIntention := Compatibility.None,
       libraryDependencies ++= Seq(
-        ("org.http4s" %% "http4s-client" % http4sVersion).cross(CrossVersion.for3Use2_13),
-        ("org.http4s" %% "http4s-async-http-client" % http4sVersion % Test).cross(CrossVersion.for3Use2_13)
+        ("org.http4s" %%% "http4s-client" % "0.23.6").cross(CrossVersion.for3Use2_13)
```
armanbilge, line 47: http4s cross-builds for Scala.js as of 0.23.5, so I hard-coded it in here for now. Ideally you should update the version globally but I personally can't chase down the various compiler warnings etc. :)

Also, `CrossVersion.for3Use2_13` is really terrible, I hope you are not publishing any of these modules for Scala 3 like this. It's why we had to bump scalajs-dom to v2.0.0:
- https://github.com/scala-js/scala-js-dom/issues/451

julienrf: I would prefer it if we can manage only one version of http4s. But we need someone to help and address the migration warnings to upgrade to 0.23.6 on the server-side.

> Also, `CrossVersion.for3Use2_13` is really terrible, I hope you are not publishing any of these modules for Scala 3 like this. It's why we had to bump scalajs-dom to v2.0.0:

We don’t publish Scala 3 modules. But are you saying that we could remove the `for3Use2_13` from our http4s dependencies? That would be great.

julienrf:
Thank you @armanbilge! I am all in favor of cross-building this module. Could you please have a look at the CI failures?

http4s/build.sbt
armanbilge: Yes, http4s 0.23.x has been published for Scala 3 from the beginning AFAIK. So in theory you could remove that `for3Use2_13`

FWIW I also tried removing that, but got into a mess of conflicting dependencies probably because `for3Use2_13` is used in many places over your project (I understand it's necessary for Akka). I'll try it again, but no promises :)

> But we need someone to help and address the migration warnings to upgrade to 0.23.6 on the server-side.

I'm not familiar with blaze server, but if it's okay with you I can try and move you to ember server, which in general more emphasis will be placed on going forward anyway.

armanbilge: Ok, I think I was able to bump http4s to 0.23.6 and address the warnings without switching to ember. I also tried removed `for3Use2_13` but lets see what CI thinks.

julienrf (approved):
That was fast! Thank you @armanbilge

armanbilge:
Yeah, so there's still the problem with dropping `for3use2_13`:
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/home/runner/work/endpoints4s/endpoints4s/"), "http4s-server"):
[error]    org.typelevel:simulacrum-scalafix-annotations _3, _2.13
[error]    org.typelevel:cats-kernel _3, _2.13
[error]    org.typelevel:cats-core _3, _2.13
```

Essentially, you are doing `for3use2_13` on some other project that also depends on cats, causing the problem when http4s brings in cats a second time.

Any guesses what project that might be?

julienrf:
I can’t have a look right now, but I am happy to merge it like it is now (I saw that you reverted one commit), and we can improve the build later.

armanbilge:
Yeah, I think once your dependencies are all on Scala 3 this problem will vanish pretty easily :)

julienrf:
It seems there are some issue with binary suffixes. I would say there is something wrong here because a Scala 3 artifact depends on Scala library 2.13 https://search.maven.org/artifact/io.github.cquiroz/cldr-api_sjs1_3/2.7.0/jar

julienrf:
True… 🙈 

Anyway, I fixed the conflicts.

armanbilge:
> Anyway, I fixed the conflicts.

Yes, looks great!! Very happy to see that 😁

PR #942, opened 2021-11-08
Update to sjs-dom 2.0.0, laminar 0.14.0

armanbilge:
Supersedes #931 and #940.

### DIFF
--- a/documentation/build.sbt
+++ b/documentation/build.sbt
@@ -286,7 +286,7 @@ val `example-cqrs-web-client` =
       noPublishSettings,
       `scala 2.12 to 2.13`,
       libraryDependencies ++= Seq(
-        "com.raquo" %%% "laminar" % "0.13.1",
+        "com.raquo" %%% "laminar" % "0.14.0",
         "org.julienrf" %%% "faithful-cats" % "2.0.0",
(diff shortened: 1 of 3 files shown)


armanbilge:
I think you'll have to bump your version number for the xhr module, but otherwise I think this is good to be merged 👍

julienrf:
May I ask you to relax de compatibility intention in the necessary modules?

armanbilge:
Sure, I'll give a try. Is there a reason to prefer that to bumping the version?

julienrf:
> Is there a reason to prefer that to bumping the version?

Both will go together: we need to indicate that we intend to break the binary compatibility of the impacted modules, and when we will cut the release we will set the module versions accordingly.

julienrf (approved):

armanbilge:
Ah, thanks for explaining! The tooling I'm used to working with is a bit different, semantic versioning is assumed and compatibility is dictated via the version increment :)

julienrf:
No problem! You can find some more explanations here: https://github.com/endpoints4s/endpoints4s/blob/master/CONTRIBUTING.md#breakage-policy and there: https://github.com/endpoints4s/endpoints4s/blob/master/CONTRIBUTING.md#introduction-of-source-or-binary-incompatibilities

Issue #945, opened 2021-11-09
Client interpreter test compatibility with JVM and JS environments


ptrdom:
Initially discussed in https://github.com/endpoints4s/endpoints4s/pull/934.

Current client interpreter test setup is not compatible with JS environment because of the use of dependencies which are incompatible with it. Suggested fixes:

1. Basing tests on `AsyncWordSpec` - current tests are based on synchronous `AnyWordSpec` and interact with async code through blocking - this is not supported in JS environment.
2. Producing stubs with Akka-HTTP server - current tests run WireMock instances and produce stubs from within the test implementations - this cannot be done in JS environment. WireMock could be ran through Docker before starting tests, with stubs prepared through configuration, but WireMock has limited stubbing support for `ChunkedEntities` and WebSockets - which will be needed for future https://github.com/endpoints4s/endpoints4s/issues/944. Starting an actual HTTP server in a similar manner - before starting tests and with stubs prepared through its implementation - seems like the best idea.

ptrdom:
@julienrf Does the suggested implementation seem fine to you? I would proceed with this before continuing with either https://github.com/endpoints4s/endpoints4s/issues/953 or https://github.com/endpoints4s/endpoints4s/issues/954.

julienrf:
Thank you for writing this up, yes, that seems to be the way to go! Thank you also for offering your help 😄

Issue #958, opened 2021-11-21
Http4s client chunked requests do not work


ptrdom:
Noticed while working on https://github.com/endpoints4s/endpoints4s/issues/945.

Likely remained unnoticed because the test implementation discards the failing assertion:
https://github.com/endpoints4s/endpoints4s/blob/19f893087c0f8f9060f14e79e3925b4030dba738/algebras/algebra/src/test/scala/endpoints4s/algebra/client/ChunkedJsonEntitiesTestSuite.scala#L180
With `callStreamedEndpoint(uploadEndpointTest, Source(expectedItems)).futureValue` the test starts failing.

julienrf:
Thank you for reporting! I think the problem comes from the fact that the tests assume that no reframing can happen…

ptrdom:
This one is actually an issue of JVM clients 😞 
Akka HTTP client fails with `org.scalatest.exceptions.TestFailedException: Vector(Array(1, 2, 3), Array(4, 5, 6)) did not equal List(Array(1, 2, 3), Array(4, 5, 6))`.
Http4s JVM client fails with `org.scalatest.exceptions.TestFailedException: Vector(Array()) did not equal List(Array(1, 2, 3), Array(4, 5, 6))`.
While I already fixed the assertion while working on https://github.com/endpoints4s/endpoints4s/issues/945, Http4s issue remains. I am not very familiar with the library.

julienrf:
To be honest, it might be that http4s can’t support the endpoints4s way of streaming entities. If this is the case, we should remove the `ChunkedEntities` interpreters. Maybe @GrafBlutwurst can help?

GrafBlutwurst:
@ptrdom do you have a branch i can replicate this on locally? I was looking for a case like this as it was one of the open discussion points. I was unable to replicate this myself though.

ptrdom:
@GrafBlutwurst The issue can be replicated by modifying the last test in `endpoints4s.algebra.client.ChunkedJsonEntitiesTestSuite` to the following implementation:
```scala
"Encode chunks uploaded to a server" in {

  val expectedItems = List(Array[Byte](1, 2, 3), Array[Byte](4, 5, 6))

  serving(
    (post & path("upload")) {
      extractRequestEntity { entity =>
        onSuccess(entity.dataBytes.map(_.toArray).runWith[Future[Seq[Array[Byte]]]](Sink.seq)) {
          receivedItems =>
[...]

PR #959, opened 2021-11-21
Client interpreter test compatibility with JVM and JS environments

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/945


algebras/algebra/src/test/scala/endpoints4s/algebra/client/StubServer.scala
```diff
@@ -0,0 +1,437 @@
...
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+
+// can start with sbt "algebraJVM/test:runMain endpoints4s.algebra.client.StubServer 8080"
+// Win - Start-Process -NoNewWindow sbt -ArgumentList "`"algebraJVM/test:runMain endpoints4s.algebra.client.StubServer 8080`""
+// Linux - sbt "algebraJVM/test:runMain endpoints4s.algebra.client.StubServer 8080" &
+// Use sbt-revolver in development - project algebraJVM; set test/reStartArgs := Seq("8080"); set test/reStart/mainClass := Some("endpoints4s.algebra.client.StubServer"); reStart.
+// sbt-revolver not work because of https://github.com/spray/sbt-revolver/issues/82.
+object StubServer extends App {
```
ptrdom, line 26: Could not decide if this should be moved to a separate SBT project. It does make sense to separate `StubServer` lifecycle and dependencies from the actual project, and avoiding the https://github.com/spray/sbt-revolver/issues/82 issue with `sbt-revolver` would also allow its neat use during development.

julienrf: It seems fine to make it a separate project. What would be the drawbacks of doing so?

ptrdom: Cannot think of any, only thing that comes to mind is that because repository's root is an SBT project it feels kind of funky to have a SBT project within a SBT project :) I will separate it out.

julienrf:
Sorry, I thought you meant an sbt module, not project.

algebras/algebra/src/test/scala/endpoints4s/algebra/client/StubServer.scala
ptrdom: I am replying to https://github.com/endpoints4s/endpoints4s/pull/959#issuecomment-974909411 here to continue the thread.

So do you think it should be a sub-module of endpoints4s sbt project or a separate sbt project?

julienrf: A sub-module. Sorry for not being clear!

algebras/algebra/src/test/scala/endpoints4s/algebra/client/ChunkedJsonEntitiesTestSuite.scala
```diff
@@ -1,43 +1,14 @@
...
-import akka.util.ByteString
 import endpoints4s.algebra.ChunkedJsonEntitiesTestApi
 
 import scala.concurrent.Future
-import scala.concurrent.duration.DurationInt
-
-trait ChunkedJsonEntitiesTestSuite[T <: ChunkedJsonEntitiesTestApi] extends ClientTestBase[T] {
 
+protected trait ChunkedJsonEntitiesTestSuite[T <: ChunkedJsonEntitiesTestApi]
+    extends ClientTestBase[T] {
```
ptrdom, line 8: Split this one up because of https://github.com/endpoints4s/endpoints4s/issues/958.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
       with:
         java-version: adopt@1.11
     - run: npm install jsdom@12.2.0
+    - run: sbt stub-server/docker:publishLocal
+    - run: docker run -d -p 8080:8080 stubserver:1.0.0
```
ptrdom, line 23: For some reason switching Scala versions kills background SBT processes, so this seems like the only way to get the StubServer running properly. Costs about a minute and a half in CI time though.

http4s/client/js/src/test/scala/endpoints4s/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
```diff
@@ -0,0 +1,59 @@
...
+  type EffectResource[A] = effect.Resource[IO, A]
+
+  val client = new TestJsonSchemaClient[IO](
+    Uri.Authority(
+      host = Uri.RegName("localhost"),
+      port = Some(stubServerPort)
+    ),
+    Uri.Scheme.http,
+    FetchClientBuilder[IO].create
+  )
```
armanbilge, line 45: This is fantastic! Thrilled to see this tested/passing :)

ptrdom:
By the way, I think it would be okay to merge this before https://github.com/endpoints4s/endpoints4s/pull/934 - either way there will be a merge conflict and this PR has probably more value currently.

http4s/client/js/src/test/scala/endpoints4s/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
```diff
@@ -0,0 +1,59 @@
...
+      endpoint: client.Endpoint[Req, Resp],
+      args: Req
+  ): Future[Resp] = {
+    val eventualResponse = endpoint.sendAndConsume(args)
+    eventualResponse.unsafeToFuture()
+  }
+
+  clientTestSuite()
+  basicAuthSuite()
+  jsonFromCodecTestSuite()
```
julienrf, line 57:
```suggestion
  jsonFromCodecTestSuite()
  sumTypedRequestsTestSuite()
```

http4s/client/jvm/src/test/scala/endpoints4s/http4s/client/Http4sClientUrlEncodingTest.scala
```diff
@@ -0,0 +1,40 @@
+package endpoints4s.http4s.client
+
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
+import endpoints4s.algebra
+import org.http4s.Uri
+import org.http4s.asynchttpclient.client.AsyncHttpClient
+import org.scalatest.BeforeAndAfterAll
+
+class Http4sClientUrlEncodingTest
```
julienrf, line 10: We might want to consider defining a base class for the JVM and JS test suites, to avoid code duplication between them.

play/client/src/test/scala/endpoints4s/play/client/PlayClientEndpointsUrlEncodingTest.scala
```diff
@@ -0,0 +1,17 @@
...
+import endpoints4s.algebra
+import play.api.test.WsTestClient
+
+import scala.concurrent.ExecutionContext
+
+class PlayClientEndpointsUrlEncodingTest extends algebra.client.UrlEncodingTestSuite[TestClient] {
+
+  import ExecutionContext.Implicits.global
+
+  val wsClient = new WsTestClient.InternalWSClient("http", 8080)
```
julienrf, line 12: Should we also `close()` the client afterAll the tests have run?

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
```diff
@@ -0,0 +1,37 @@
...
+  override def equals(other: Any): Boolean =
+    other match {
+      case that: EndpointsSettings =>
+        host == that.host
+      case _ => false
+    }
+
+  override def hashCode(): Int = Hashing.hash(host)
+
+  @nowarn("cat=unused")
```
julienrf, line 21: It is used to implement `withHost`. Is this a leftover?

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
```diff
@@ -0,0 +1,37 @@
+package endpoints4s.xhr
+
+import endpoints4s.Hashing
+
+import scala.annotation.nowarn
+
+final class EndpointsSettings(val host: Option[String]) extends Serializable {
```
julienrf, line 7:
```suggestion
final class EndpointsSettings private (val host: Option[String]) extends Serializable {
```

julienrf, line 7: Some scaladoc would be great 😃

julienrf: By the way, I see in the tests that this `host` field is not really just a host. It also contains the port number, and even the protocol scheme (e.g., “http://localhost:8080” as opposed to just “localhost”). I wonder if there would be a better approach here. I like what we have in the http4s interpreter, although it requires maybe too much boilerplate?

xhr/client/src/test/scala/endpoints4s/xhr/future/XhrClientEndpointsTest.scala
```diff
@@ -0,0 +1,47 @@
...
+    EndpointsSettings().withHost(Some("http://localhost:8080"))
+  )
+
+  def call[Req, Resp](
+      endpoint: client.Endpoint[Req, Resp],
+      args: Req
+  ): Future[Resp] = endpoint(args)
+
+  clientTestSuite()
+  basicAuthSuite()
```
julienrf, line 43: Those two are already called in the separate class `XhrClientEndpointsJsonSchemaTest`, we should either call them here or there.

documentation/examples/quickstart/client/src/main/scala/quickstart/CounterClientFuture.scala
```diff
@@ -9,6 +10,8 @@ object CounterClientFuture
     with xhr.future.Endpoints
     with xhr.JsonEntitiesFromSchemas {
 
+  val settings: EndpointsSettings = EndpointsSettings()
```
julienrf, line 13: I think it should be a `lazy val` to guard against possible initialization issues.

algebras/algebra/src/test/scala/endpoints4s/algebra/EndpointsTestApi.scala
```diff
@@ -218,4 +218,112 @@ trait EndpointsTestApi extends algebra.Endpoints {
...
+  )
+
+  val emptyResponseErrorEndpoint = endpoint(
+    get(
+      path / "error" / "user" / segment[String]() / "description" /? (qs[String]("name") & qs[
+        Int
+      ]("age"))
+    ),
+    ok(emptyResponse)
+  )
```
julienrf, line 291: Those two endpoint definitions are the same except for the response entity. I see that the sub server always returns an empty response (unless I did a mistake). I would expect the tests that call `errorEndpoint` to fail. Am I missing something?

julienrf (commented):
Hey, thank you @ptrdom, please read my comments below!

Overall, I think this is a great step forward 🙏

http4s/client/jvm/src/test/scala/endpoints4s/http4s/client/Http4sClientUrlEncodingTest.scala
ptrdom: I thought about leaving this test in `shared`, but for some reason SBT was ignoring it. I think I had some weird caching issues from mixed use of SBT directly and through BSP, I will try it again.

play/client/src/test/scala/endpoints4s/play/client/PlayClientEndpointsUrlEncodingTest.scala
ptrdom: There are a few errors in CI that seem to indicate something being open when it should not be, I will try this out.

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
ptrdom: Looking at Akka HTTP client as an example, this could be named  `baseUrl` and be a native class, for example https://developer.mozilla.org/en-US/docs/Web/API/URL. How does that sound?

julienrf: Yes, sounds good! 👍

algebras/algebra/src/test/scala/endpoints4s/algebra/EndpointsTestApi.scala
ptrdom: I am not sure myself what is tested here, assertions are the same for both endpoints. Same situation is with detailed error endpoints. What change would you like to have here?

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
ptrdom: Looking at the constructor of URL - https://developer.mozilla.org/en-US/docs/Web/API/URL/URL - maybe that is not such a good idea because relative URLs must have a base, where XHR does not require it. Seems that we can only build our own abstraction out of native classes for this, but should we do it?

algebras/algebra/src/test/scala/endpoints4s/algebra/EndpointsTestApi.scala
julienrf: That’s a good point. After looking at the tests I am not sure what they test either… But at least I see that both tests use the same “mock” server, so what you did is correct.

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
julienrf: I think it is fine to use `String` here. I just wanted to raise the point that “host” might not be the best name, and that we should clarify in the docs what the users can put in this field.

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
ptrdom: I copy pasted this from Fetch implementation, IIRC this was required for Scala 3 support.

http4s/client/jvm/src/test/scala/endpoints4s/http4s/client/Http4sClientUrlEncodingTest.scala
ptrdom: I wonder how to set it up in sbt - `http4s-client` `shared` has to depend on algebra, which is a cross project in itself. Any ideas?

`.dependsOn(algebra % "test->test;compile->compile")`
Just throws errors:
```
found   : sbt.ClasspathDependency
 required: sbtcrossproject.CrossClasspathDependency
    .dependsOn(algebra % "test->test;compile->compile")
```

ptrdom: Figured it out!
```
.dependsOnLocalCrossProjectsWithScope(
      "algebra" -> "test->test;compile->compile"
    )
```

julienrf: Indeed!

I plan to move the “test” part of the algebra to a separate project soon, so that we can just `dependsOn` it.

xhr/client/src/test/scala/endpoints4s/xhr/future/XhrClientEndpointsTest.scala
ptrdom: Kept the ones in `XhrClientEndpointsTest`. In general it makes sense to avoid these "uber" specs to allow for improved parallel test execution - tests of a single suite are the only ones that cannot be parallelized.

julienrf: Yes, I agree! This is another thing I would like to change, eventually. Also, it’s a pain in the ass when you want to see the output of just one test.

documentation/examples/basic/client/src/main/scala/sample/Api.scala
```diff
@@ -9,4 +9,6 @@ object Api
     with JsonEntitiesFromCodecs
     with Assets
     with thenable.Endpoints
-    with BasicAuthentication
+    with BasicAuthentication {
+  val settings: EndpointsSettings = EndpointsSettings()
```
julienrf, line 13:
```suggestion
  lazy val settings: EndpointsSettings = EndpointsSettings()
```

documentation/examples/cqrs/web-client/src/main/scala/cqrs/webclient/PublicEndpoints.scala
```diff
@@ -1,8 +1,11 @@
...
 
 import endpoints4s.xhr
+import endpoints4s.xhr.EndpointsSettings
 
 object PublicEndpoints
     extends cqrs.publicserver.PublicEndpoints
     with xhr.faithful.Endpoints
-    with xhr.JsonEntitiesFromCodecs
+    with xhr.JsonEntitiesFromCodecs {
+  val settings: EndpointsSettings = EndpointsSettings()
```
julienrf, line 10:
```suggestion
  lazy val settings: EndpointsSettings = EndpointsSettings()
```

documentation/examples/quickstart/client/src/main/scala/quickstart/CounterClient.scala
```diff
@@ -10,5 +11,7 @@ import endpoints4s.xhr
 object CounterClient
     extends CounterEndpoints
     with xhr.thenable.Endpoints
-    with xhr.JsonEntitiesFromSchemas
+    with xhr.JsonEntitiesFromSchemas {
+  val settings: EndpointsSettings = EndpointsSettings()
```
julienrf, line 15:
```suggestion
  lazy val settings: EndpointsSettings = EndpointsSettings()
```

julienrf (approved):
Excellent, thank you @ptrdom! I think we should use lazy vals everywhere to provide the settings.

PR #961, opened 2021-11-25
Fix Http4s byte ChunkedEntities

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/958


julienrf (approved):
Thank you for the fix!

julienrf:
There is now a conflict because of the other PR, could you fix it?

PR #963, opened 2021-11-26
Ship reusable testing code in testkit modules

julienrf:
The tests of the algebra module don’t test anything. Instead, they provide (abstract) test specifications to be used by the interpreters to validate that they conform to the specification, or fixtures that can be conveniently reused by other modules.

By moving these testkits into separate modules, it becomes easier for third parties to validate that their interpreters are correct.

It also simplifies the build definition because we don’t need anymore dependencies like `foo % "test->test;compile->compile"`.

Last, the test modules were also used to carry snippets of code used in the documentation. Those are not included in the testkit modules (and they have been duplicated in each interpreter). The idea is to have decoupled documentation for interpreters.


julienrf:
Hey @ptrdom, I wonder if the errors I get could be related to #959?

ptrdom:
I noticed this error when working on tests, but I could not see how my changes would impact it - `StubServer` only binds to port 8080 - so I assumed the test is just flaky.

ptrdom:
Before I only checked errors on 2.13 CI job, now I checked the ones on 2.12 and those do not look like they could be flaky. Strange.

ptrdom:
I think the issue might be that since the tests are actually running in parallel and with more suites available SBT starts to parallelize more aggressively, the `9000` port servers from `cqrs.commands.CommandsTest` and `cqrs.Test` are colliding. It would be best to change all ports in tests to dynamic ones.

julienrf:
I noticed that in `master` if I run `http4s-clientJS/test`, no tests are run.

ptrdom:
I did run into this issue, but managed to get them executing without changes to code, so I assumed this was some dirty local cache issue, apparently it is not. I can replicate this now, and I cannot get those tests running. Bug of https://github.com/portable-scala/sbt-crossproject then? Could be something in the project setup messing with it, but I currently do not have ideas.

julienrf:
I have no idea either. Could be the interaction of multiple things… Good news is that this branch seems to not run into the issue.

ptrdom:
http4s-clientJS spec is failing because of missing `implicit override def executionContext: ExecutionContextExecutor = JSExecutionContext.queue`. Then the new mapping tests will fail because of bug in `StubServer`, fixed by https://github.com/endpoints4s/endpoints4s/pull/934/commits/e3f2dc3fdc70a96f167fd064032b7606174e6b07.

julienrf:
Thank you for the help @ptrdom! Now the http4s-client tests pass (locally at least!). I am looking into the cqrs tests.

julienrf:
I am stuck again. The http4s-clientJS tests fail in the CI. Locally, it runs fine though. Do you have any idea what could be wrong @ptrdom?

julienrf:
Oh, good catch @ptrdom, it seems disabling scoverage did the trick! Thank you so much!! Is it OK for you if I merge the PR?

ptrdom (approved):
LGTM

Issue #954, opened 2021-11-18
WebSocket algebra


ptrdom:
Initially discussed in https://github.com/endpoints4s/endpoints4s/issues/944.

It was decided that a `StreamedEntities` implementation which hides the implementation details of the streaming protocol is not ideal, so instead a new algebra for WebSocket based communication should be implemented.

PR #964, opened 2021-11-28
WebSocket algebra

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/954.

### DIFF
--- a/algebras/algebra/src/main/scala/endpoints4s/algebra/WebSockets.scala
+++ b/algebras/algebra/src/main/scala/endpoints4s/algebra/WebSockets.scala
@@ -1,0 +1,64 @@
+package endpoints4s.algebra
+
+import endpoints4s.Tupler
+
+trait WebSockets extends WebsocketClientStream with WebsocketServerStream with Urls {
+
+  type WebSocket[A, B]
+
+  def webSocket[URLC, CSC, In, Out](
+      url: Url[URLC],
+      clientStream: WebsocketClientStream[CSC] = emptyWebsocketClientStream,
+      serverStream: WebsocketServerStream[Out] = emptyWebsocketServerStream
+  )(implicit
+      tupler: Tupler.Aux[URLC, CSC, In]
+  ): WebSocket[In, Out]
+}
+
+trait WebSocketsWithUpgrade
+    extends WebSockets
+    with WebSocketUpgradeRequest
+    with WebSocketUpgradeResponse {
+
+  type WebSocketWithUpgrade[A, B]
+
+  def webSocketWithUpgrade[URLC, CSC, URHC, SSS, URHS, URLCAndCSCTupled, In, Out](
+      url: Url[URLC],
+      clientStream: WebsocketClientStream[CSC] = emptyWebsocketClientStream,
+      upgradeRequestHeaders: RequestHeaders[URHC] = emptyRequestHeaders,
+      serverStream: WebsocketServerStream[SSS] = emptyWebsocketServerStream,
+      upgradeResponseHeaders: RequestHeaders[URHS] = emptyRequestHeaders
+  )(
+      tuplerURLCCSC: Tupler.Aux[URLC, CSC, URLCAndCSCTupled],
+      tuplerURLCAndCSCTupledURHC: Tupler.Aux[URLCAndCSCTupled, URHC, In],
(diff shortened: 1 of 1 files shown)


ptrdom:
Since browser clients cannot interact with upgrade response/request, I though it would be best to explicitly separate `webSocket` and `webSocketWithUpgrade`. My thinking was that if an endpoint describes an interaction with headers and client is unable to provide them, it should not be able to communicate over said endpoint because it will be incompatible with server implementation. Slight issue with that approach is that browsers can implicitly pass some headers, like cookies, so server implementation can interact with them, but I am not sure if this implicit contract should be handled by the implementation.

Only `upgradeRequestHeaders` and `upgradeResponseHeaders` are provided because the spec seems to not put any importance into request/response entities.

@julienrf What do you think?

algebras/algebra/src/main/scala/endpoints4s/algebra/WebSockets.scala
```diff
@@ -0,0 +1,64 @@
...
+      serverStream: WebsocketServerStream[SSS] = emptyWebsocketServerStream,
+      upgradeResponseHeaders: RequestHeaders[URHS] = emptyRequestHeaders
+  )(
+      tuplerURLCCSC: Tupler.Aux[URLC, CSC, URLCAndCSCTupled],
+      tuplerURLCAndCSCTupledURHC: Tupler.Aux[URLCAndCSCTupled, URHC, In],
+      tuplerSSSURHS: Tupler.Aux[SSS, URHS, Out]
+  ): WebSocketWithUpgrade[In, Out]
+}
+
+trait WebsocketClientStream {
```
julienrf, line 38: I guess you mean _request_ stream?

algebras/algebra/src/main/scala/endpoints4s/algebra/WebSockets.scala
```diff
@@ -0,0 +1,64 @@
...
+}
+
+trait WebSocketsWithUpgrade
+    extends WebSockets
+    with WebSocketUpgradeRequest
+    with WebSocketUpgradeResponse {
+
+  type WebSocketWithUpgrade[A, B]
+
+  def webSocketWithUpgrade[URLC, CSC, URHC, SSS, URHS, URLCAndCSCTupled, In, Out](
```
julienrf, line 25: I am not familiar enough with the upgrade thing, but since you are just using a standard `RequestHeaders` parameter, I wonder if we should just replace the `webSocket` constructor with this one? Why do we want to have both?

julienrf: Reading your last comment again, now I think I understand: in the case of web browsers, there would be no possibility of manipulating the request or response headers?

algebras/algebra/src/main/scala/endpoints4s/algebra/WebSockets.scala
ptrdom: Terms such as request and response are not really used for anything but the upgrade in WebSocket docs that I have read, but there is a clear definition of client and server, so I thought I would use those terms.

algebras/algebra/src/main/scala/endpoints4s/algebra/WebSockets.scala
ptrdom: Yes, the exact reasoning was provided in that comment - there is no _explicit_ way to interact with the headers, but browsers can _implicitly_ pass some headers. Based on reasons provided, does this separation make sense to you?

ptrdom: Yes.

julienrf: Then, indeed, if we want to provide `webSocketWithUpgrade` (for those who can implement it), it has to be in a separate trait. Thank you for the explanation!

ptrdom:
After doing more research and thinking, I do not think that a stream is a correct abstraction to express data flowing through a WebSocket. Issue with streams is that they have a definition of an end, and WebSockets lack that - neither client nor server can communicate they either of them have stopped sending their messages, they can only close the connection. Inability to explicitly communicate these clearly separate behaviors puts a limit to the available expression and requires developers to come up with their own protocols. This makes WebSockets not as valuable as I thought before, and they are not a replacement for `ChunkedEntities`.

PR #969, opened 2021-12-02
Mark Akka dependencies as “provided”

julienrf:


julienrf:
Does anyone want to comment on this breaking change?

documentation/manual/src/main/paradox/interpreters/akka-http.md
```diff
@@ -2,6 +2,27 @@
...
+`akka-http-client` have marked their dependency to Akka as “provided”.
+
+As a consequence, to use these interpreters you will have to
+explicitly add a dependency on `akka-stream`:
+
+~~~ scala
+"com.typesafe.akka" %% "akka-stream" % "<akka-version>"
+~~~
+
+Where `<akka-version>` is higher or equal to @var[akka-version].
```
ptrdom, line 22: Should we mention binary compatibility?

ptrdom (commented):
I think it is a good idea to be explicit about how Akka operates, most libraries do not care about this, I already got used to using `dependencyOverrides` - https://www.lagomframework.com/documentation/1.6.x/scala/Akka.html#Updating-Akka-version.

ptrdom (approved):

Issue #953, opened 2021-11-18
Protocol improvements for ChunkedEntities


ptrdom:
Initially discussed in https://github.com/endpoints4s/endpoints4s/issues/944.

Because of reframing done by browsers upon the consumption of chunked encoding chunks, a chunk cannot be used as delimiting boundary of a stream element. Most flexible fix seems to be an implementation that sends stream elements prefixed by their length.

ptrdom:
More generic solution to chunk delimiting problem would be to introduce a new delimiter algebra to clearly express the issue. For example, current implementation could be called a `chunkDelimiter`, because each stream element is put in its own chunk, and browsers would not implement this algebra. But then there could be `lengthPrefixDelimiter` - this one is described the issue description - or a `newLineDelimiter` and so on, and browsers would be able to implement these, and in turn implement the endpoints using them. 

Also, to accommodate the limitations of browsers - inability to send chunked requests - `ChunkedEntities` should have separate algebras for requests and responses.

@julienrf What do you think about these suggestions?

ptrdom:
Though _delimiter_ may not be the most ideal term, maybe something like `chunksEncoder`/`chunksDecoder` makes more sense.

julienrf:
I might be wrong but I think we could leave `byteChunks` and `textChunks` constructors as they are, and document that interpreters must not assume a 1:1 correspondence between chunks sent by a party and chunks received by another party (ie, reframing can happen).

This is why I think only `ChunkedJsonEntities` needs to be fixed. I am fine with introducing a concept of delimiter in the algebra, but we should double-check what are the current practices in the wild. I have seen that `circe-fs2` supports `'\n'` delimited JSON objects, so that’s good. What about OpenAPI? Is there a way to document how “stream items” are delimited? We could check AsyncAPI as well.

Does that make sense?

ptrdom:
> I might be wrong but I think we could leave `byteChunks` and `textChunks` constructors as they are, and document that interpreters must not assume a 1:1 correspondence between chunks sent by a party and chunks received by another party (ie, reframing can happen).

I think I understand your point now, and I do agree. It is not necessary for the library to provide these, users can do it themselves if they need it. In addition to documentation, we should also alter the tests so that browser interpreters can implement `byteChunks` and `textChunks`.

[...]

PR #972, opened 2021-12-06
Protocol improvements for ChunkedEntities

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/953.

### DIFF
--- a/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
+++ b/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
@@ -68,30 +68,26 @@ trait ChunkedJsonEntities
     with ChunkedEntities
     with JsonEntitiesFromCodecs {
 
-  type RequestFraming = Flow[ByteString, ByteString, NotUsed]
-
-  type ResponseFraming = Flow[ByteString, ByteString, NotUsed]
-
   def jsonChunksRequest[A](implicit
       codec: JsonCodec[A]
-  ): RequestEntity[Chunks[A]] = jsonChunksRequest(Flow.apply[ByteString])
+  ): RequestEntity[Chunks[A]] = jsonChunksRequest(noopFraming)
 
-  def jsonChunksRequest[A](framing: RequestFraming)(implicit
+  def jsonChunksRequest[A](framing: Framing)(implicit
       codec: JsonCodec[A]
   ): RequestEntity[Chunks[A]] = {
     val encoder = stringCodec(codec)
     chunkedRequestEntity(
       ContentTypes.`application/json`,
       a => ByteString.fromString(encoder.encode(a)),
-      framing
+      framing.request
     )
   }
 
   def jsonChunksResponse[A](implicit
       codec: JsonCodec[A]
-  ): ResponseEntity[Chunks[A]] = jsonChunksResponse(Flow.apply[ByteString])
+  ): ResponseEntity[Chunks[A]] = jsonChunksResponse(noopFraming)
 
-  def jsonChunksResponse[A](framing: ResponseFraming)(implicit
+  def jsonChunksResponse[A](framing: Framing)(implicit
       codec: JsonCodec[A]
   ): ResponseEntity[Chunks[A]] = {
     val decoder = stringCodec(codec)
@@ -104,15 +100,29 @@ trait ChunkedJsonEntities
           .left
           .map(errors => new Throwable(errors.mkString(". ")))
       },
-      framing
+      framing.response
     )
   }
 
-  def newLineDelimiterRequestFraming[A]: RequestFraming =
-    Flow[ByteString].intersperse(ByteString("\n"))
+  trait Framing {
+    def request: Flow[ByteString, ByteString, NotUsed]
+    def response: Flow[ByteString, ByteString, NotUsed]
+  }
 
-  def newLineDelimiterResponseFraming[A]: ResponseFraming =
-    Flow[ByteString].via(
-      Framing.delimiter(ByteString("\n"), maximumFrameLength = Int.MaxValue, allowTruncation = true)
+  def newLineDelimiterFraming: Framing = new Framing {
+    def request = Flow[ByteString].intersperse(ByteString("\n"))
+
+    def response = Flow[ByteString].via(
+      Framing.delimiter(
+        ByteString("\n"),
+        maximumFrameLength = Int.MaxValue,
+        allowTruncation = true
+      )
     )
+  }
+
+  private def noopFraming: Framing = new Framing {
+    def request: Flow[ByteString, ByteString, NotUsed] = Flow.apply[ByteString]
+    def response: Flow[ByteString, ByteString, NotUsed] = Flow.apply[ByteString]
+  }
 }
--- a/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
+++ b/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
@@ -102,10 +102,10 @@ trait ChunkedJsonEntities
     )
   }
 
-  def newLineRequestChunkCodec[A]: Flow[ByteString, ByteString, NotUsed] =
+  def newLineRequestChunkCodec[A]: RequestChunkCodec =
     Flow[ByteString].intersperse(ByteString("\n"))
 
-  def newLineResponseChunkCodec[A]: Flow[ByteString, ByteString, NotUsed] =
+  def newLineResponseChunkCodec[A]: ResponseChunkCodec =
     Flow[ByteString].via(
       Framing.delimiter(ByteString("\n"), maximumFrameLength = Int.MaxValue, allowTruncation = true)
     )
--- a/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
+++ b/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
@@ -103,7 +103,7 @@ trait ChunkedJsonEntities
   }
 
   def newLineRequestChunkCodec[A]: Flow[ByteString, ByteString, NotUsed] =
-    Flow[ByteString].map(chunk => chunk ++ ByteString("\n"))
+    Flow[ByteString].intersperse(ByteString("\n"))
 
   def newLineResponseChunkCodec[A]: Flow[ByteString, ByteString, NotUsed] =
     Flow[ByteString].via(
--- a/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
+++ b/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
@@ -38,20 +38,20 @@ trait ChunkedEntities extends algebra.ChunkedEntities with EndpointsWithCustomEr
   private[client] def chunkedRequestEntity[A](
       contentType: ContentType,
       toByteString: A => ByteString,
-      chunkCodec: Flow[ByteString, ByteString, NotUsed] = Flow[ByteString].map(identity)
+      framing: Flow[ByteString, ByteString, NotUsed] = Flow[ByteString].map(identity)
   ): RequestEntity[Chunks[A]] =
     (as, httpRequest) =>
       httpRequest.withEntity(
-        HttpEntity.Chunked.fromData(contentType, as.map(toByteString).via(chunkCodec))
+        HttpEntity.Chunked.fromData(contentType, as.map(toByteString).via(framing))
       )
 
   private[client] def chunkedResponseEntity[A](
       fromByteString: ByteString => Either[Throwable, A],
-      chunkCodec: Flow[ByteString, ByteString, NotUsed] = Flow[ByteString].map(identity)
+      framing: Flow[ByteString, ByteString, NotUsed] = Flow[ByteString].map(identity)
   ): ResponseEntity[Chunks[A]] =
     httpEntity =>
       Future.successful(
-        Right(httpEntity.dataBytes.via(chunkCodec).map(fromByteString).flatMapConcat {
+        Right(httpEntity.dataBytes.via(framing).map(fromByteString).flatMapConcat {
           case Left(error)  => Source.failed(error)
           case Right(value) => Source.single(value)
         })
--- a/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
+++ b/akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
@@ -1,9 +1,13 @@
 package endpoints4s.akkahttp.client
 
+import akka.NotUsed
 import akka.http.scaladsl.model.{ContentType, ContentTypes, HttpEntity}
+import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.Framing
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
 import endpoints4s.algebra
+import endpoints4s.algebra.NewLineChunkCodec
 
 import scala.concurrent.Future
 
@@ -34,19 +38,21 @@ trait ChunkedEntities extends algebra.ChunkedEntities with EndpointsWithCustomEr
 
   private[client] def chunkedRequestEntity[A](
       contentType: ContentType,
-      toByteString: A => ByteString
+      toByteString: A => ByteString,
+      chunkCodec: Flow[ByteString, ByteString, NotUsed] = Flow[ByteString].map(identity)
   ): RequestEntity[Chunks[A]] =
     (as, httpRequest) =>
       httpRequest.withEntity(
-        HttpEntity.Chunked.fromData(contentType, as.map(toByteString))
+        HttpEntity.Chunked.fromData(contentType, as.map(toByteString).via(chunkCodec))
       )
 
   private[client] def chunkedResponseEntity[A](
-      fromByteString: ByteString => Either[Throwable, A]
+      fromByteString: ByteString => Either[Throwable, A],
+      chunkCodec: Flow[ByteString, ByteString, NotUsed] = Flow[ByteString].map(identity)
   ): ResponseEntity[Chunks[A]] =
     httpEntity =>
       Future.successful(
-        Right(httpEntity.dataBytes.map(fromByteString).flatMapConcat {
+        Right(httpEntity.dataBytes.via(chunkCodec).map(fromByteString).flatMapConcat {
           case Left(error)  => Source.failed(error)
           case Right(value) => Source.single(value)
         })
@@ -61,30 +67,46 @@ trait ChunkedEntities extends algebra.ChunkedEntities with EndpointsWithCustomEr
 trait ChunkedJsonEntities
     extends algebra.ChunkedJsonEntities
     with ChunkedEntities
-    with JsonEntitiesFromCodecs {
+    with JsonEntitiesFromCodecs
+    with NewLineChunkCodec {
 
-  def jsonChunksRequest[A](implicit
+  type RequestChunkCodec = Flow[ByteString, ByteString, NotUsed]
+
+  type ResponseChunkCodec = Flow[ByteString, ByteString, NotUsed]
+
+  def jsonChunksRequest[A](chunkCodec: RequestChunkCodec)(implicit
       codec: JsonCodec[A]
   ): RequestEntity[Chunks[A]] = {
     val encoder = stringCodec(codec)
     chunkedRequestEntity(
       ContentTypes.`application/json`,
-      a => ByteString.fromString(encoder.encode(a))
+      a => ByteString.fromString(encoder.encode(a)),
+      chunkCodec
     )
   }
 
-  def jsonChunksResponse[A](implicit
+  def jsonChunksResponse[A](chunkCodec: ResponseChunkCodec)(implicit
       codec: JsonCodec[A]
   ): ResponseEntity[Chunks[A]] = {
     val decoder = stringCodec(codec)
-    chunkedResponseEntity { byteString =>
-      val string = byteString.utf8String
-      decoder
-        .decode(string)
-        .toEither
-        .left
-        .map(errors => new Throwable(errors.mkString(". ")))
-    }
+    chunkedResponseEntity(
+      { byteString =>
+        val string = byteString.utf8String
+        decoder
+          .decode(string)
+          .toEither
+          .left
+          .map(errors => new Throwable(errors.mkString(". ")))
+      },
+      chunkCodec
+    )
   }
 
+  def newLineRequestChunkCodec[A]: Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString].map(chunk => chunk ++ ByteString("\n"))
+
+  def newLineResponseChunkCodec[A]: Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString].via(
+      Framing.delimiter(ByteString("\n"), maximumFrameLength = Int.MaxValue, allowTruncation = true)
+    )
 }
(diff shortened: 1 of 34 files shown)


ptrdom:
@julienrf What do you think about these changes? Obviously documentation needs to be updated, I will work on it once we finalize the implementation.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -27,39 +27,55 @@ package endpoints4s.algebra
...
 
   /** A stream of chunks of type `A`.
     *
     * @tparam A Information carried by each chunk
     * @group types
     */
   type Chunks[A]
+}
 
+trait ChunkedRequestEntities extends Chunks {
```
julienrf, line 45: I remember we had a discussion about splitting algebra to request algebra and response algebra, but I couldn’t find it back. Could you please remember me why we need to split the algebra?

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +93,56 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
+    with JsonCodecs
+    with ResponseChunkCodec {
 
   /** A response entity carrying chunks of JSON values
     *
     * @tparam A Type of values serialized into JSON
     * @group operations
     */
-  def jsonChunksResponse[A](implicit
+  def jsonChunksResponse[A](chunkCodec: ResponseChunkCodec)(implicit
```
julienrf, line 123: Unfortunately we can’t just break the API. We should keep the previous method (maybe deprecate it) and introduce an overload, instead of changing it.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +93,56 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
+}
+
+trait NewLineChunkCodec extends NewLineRequestChunkCodec with NewLineResponseChunkCodec
+
+trait NewLineRequestChunkCodec extends RequestChunkCodec {
+
+  def newLineRequestChunkCodec[A]: RequestChunkCodec
+}
+
+trait NewLineResponseChunkCodec extends ResponseChunkCodec {
```
julienrf, line 145: There are lots of traits introduced in this file. We should be careful not to introduce too many traits because they create complexity at the use-site. So, we always need to assess whether a trait is really needed or not. In general, a good reason to introduce a separate trait is to support interpreters that can’t implement one part of the algebra. Is this the case here? Do we have examples of interpreters that can implement `ChunkedRequestEntities` but can’t implement `ChunkedResponseEntities` (or _viceversa_)? Regarding `ChunkCodec`, the fact that a whole trait defines the delimiter might be complicated to use: if someone wants to use different delimiters in the same service, they will have to create two services (more or less).

julienrf (commented):
I only had a quick look at the algebra changes.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
ptrdom: It is because of browsers - there is no support in XHR and Fetch to deal with request body streams yet.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
julienrf: So, to reply to my own question: yes, browsers can’t implement `ChunkedRequestEntities` but they can implement `ChunkedResponseEntities`.

Now, regarding the delimiters, I wonder if it would be possible to introduce a `Delimiters` algebra, extended by both `ChunkedJsonRequestEntities` and `ChunkedJsonResponseEntities`:

~~~ scala
trait Delimiters {
  type Delimiter
  def newLineDelimiter: Delimiter
}
~~~

(something along those lines)

ptrdom: Just a note in regards to naming - this abstraction is more than just about the delimiters, it can be very large arrays or length prefixes too, so that is why I chose the name _codec_. It saw it also called by _framing_. Are you sure about naming this a delimiter?

Is this abstraction not more limiting? Because you are forcing interpreters to implement `newLineDelimiter` if they want to deal with `Delimiter`. That is the reason why I split it off - to allow interpreters full flexibility.

ptrdom: This `Delimiter` (_codec_, _framing_) is basically a flow that transforms the `Chunks`. If you could take a look at my provided implementations, the type definitions sort of tie together naturally.

ptrdom: That being said, and just to clarify, I do not think that algebras should tie chunks and their transformations together.

julienrf:
I agree about naming: "framing" sounds better to me.

Regarding modules, let me look at the interpreters to get a wider view.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
julienrf: After looking at the implementation of the interpreters, it seems that we always implement the newline frame delimiter. I don’t think we should provide too many options in the algebra, unless there is a really good reason to do so.

It would be nice if it was possible to perform JSON streaming with very little code like:

~~~ scala
trait StreamingJson extends algebra.Endpoints with algebra.ChunkedJsonEntities {
  val foo = endpoint(get(path / "foo"), chunkedJsonResponse[Foo](framing = newLineDelimiterFraming))
}

object Documentation extends StreamingJson with openapi.Endpoints with openapi.ChunkedJsonEntities {
  val api = openApi(...)(foo) 
}

object Client extends StreamingJson with client.Endpoints with client.ChunkedJsonEntities

object Server extends StreamingJson with server.Endpoints with server.ChunkedJsonEntities
~~~

julienrf (commented):
Thank you Domantas! I’m trying to push again to reduce the number of traits, please let me know if you think it’s feasible, otherwise your design looks great!

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +93,62 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
+trait RequestFraming {
+
+  type RequestFraming
+  def newLineDelimiterRequestFraming[A]: RequestFraming
+}
+
+trait ResponseFraming {
 
+  type ResponseFraming
+  def newLineDelimiterResponseFraming[A]: ResponseFraming
```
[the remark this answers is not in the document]
ptrdom: Framing does two different things for requests and response that are the opposite of one another - one has to produce frames, the other has to consume them. I guess `Framing` could be a function that does one or the other based on a parameter that is passed to it, then we could have a single abstraction. How does that sound?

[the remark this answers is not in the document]
julienrf: Yes, or I thought interpreters that support both chunked requests and responses to model `Framing` like so:

~~~ scala
class Framing {
  def reframeRequestEntity(...) = ...
  def reframeResponseEntity(...) = ...
}
~~~

And then, in `chunkedJsonResponseEntity`, the interpreter would call `framing.reframeResponseEntity`.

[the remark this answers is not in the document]
ptrdom: Oh, that looks very clean, great idea! I will implement it.

akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
```diff
@@ -65,26 +70,59 @@ trait ChunkedJsonEntities
...
+          .left
+          .map(errors => new Throwable(errors.mkString(". ")))
+      },
+      framing.response
+    )
+  }
+
+  trait Framing {
+    def request: Flow[ByteString, ByteString, NotUsed]
+    def response: Flow[ByteString, ByteString, NotUsed]
```
julienrf, line 109: It would be great to document what is expected from these flows. Especially since their input and output types are both `ByteString`, it is easy to implement them with the “identity” flow, but that would lead to incorrect results. I think it is important to say that the request flow takes as input the frames, and outputs the frames “wrapped” in a specific way.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +92,53 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
       codec: JsonCodec[A]
   ): ResponseEntity[Chunks[A]]
+}
+
+/** Algebra interface for describing how chunks of chunked transfer-encoding requests and responses should be framed.
+  * Being explicit about how chunks are framed solves the issue of re-framing happening during transport.
+  */
+trait Framing {
 
+  type Framing
```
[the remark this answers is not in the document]
julienrf: This is just a stub of documentation, it would be good to emphasize that client and server interpreters have to be consistent (so that streams framed by a server are correctly decoded by a client that uses the same `Framing`).

julienrf (commented):
This looks great! I had a quick look and I think the design is now ideal! Please let me know if there are some specific parts you would like me to review… Or if the whole PR is ready to review?
Could you add some more scaladoc before I do the final pass? I added some suggestions but you are free to write different content. Only the `@group` annotations are important.

julienrf:
Ah, we also need to relax the compatibility intention of the algebra modules (the `versionPolicyIntention` key in the root `build.sbt` file) to be `Compatibility.BinaryCompatible`. It might not be enough to fix all the compatibility issues, but let’s see what its outcome is.

akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
ptrdom: Would it be best to document this on algebra or on each interpreter?

ptrdom: https://github.com/endpoints4s/endpoints4s/pull/972#discussion_r773998549 comment is about the same thing, as I understand.

ptrdom:
> Please let me know if there are some specific parts you would like me to review… Or if the whole PR is ready to review?

Since we agreed on most of the implementation, I was mainly expecting comments in regards to documentation and some minor implementation improvements, so thank you for providing that!

akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
julienrf: On the algebra, it is good to mention how client, server, and documentation interpreters must interpret the member. Like here, for instance:

https://github.com/endpoints4s/endpoints4s/blob/9493c2edf1e4d488228d98805104075e106bb2b0/algebras/algebra/src/main/scala/endpoints4s/algebra/Requests.scala#L132-L141

On interpreters, it is good to underline the important details that matter for the users of that interpreter, in case they want to customize, integrate, compose, etc.

ptrdom: I added additional documentation on the algebra level, but I am not too sure about what to add on interpreter level that is not already said on the algebra level. I was trying to not unnecessarily overexplain it, since I assume that developers implementing their own framing algorithms will have to have some prior knowledge for what framing is and will need to write their own tests too. But I am definitely open to any suggestions for documentation improvements!

One small thing to have in mind is that framing can do more than just wrap them, for example it could also chop them up when encoding for more efficient transport.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -27,39 +30,51 @@ package endpoints4s.algebra
...
+  /** A request entity carrying chunks of `Array[Byte]` values
     *
     * @group operations
     */
-  def textChunksResponse: ResponseEntity[Chunks[String]]
+  def bytesChunksRequest: RequestEntity[Chunks[Array[Byte]]]
+}
 
-  /** A request entity carrying chunks of `Array[Byte]` values
+trait ChunkedResponseEntities extends Chunks {
```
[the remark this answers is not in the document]
julienrf: Please include the above suggestion.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +94,63 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
+      codec: JsonCodec[A]
+  ): RequestEntity[Chunks[A]]
 
   /** A request entity carrying chunks of JSON values
     *
     * @tparam A Type of values serialized into JSON
     * @group operations
     */
-  def jsonChunksRequest[A](implicit
+  def jsonChunksRequest[A](framing: Framing)(implicit
```
julienrf, line 115: It would be good to add a `@param framing` documentation to the scaladoc of the method.

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +94,63 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
       codec: JsonCodec[A]
   ): RequestEntity[Chunks[A]]
+}
+
+trait ChunkedJsonResponseEntities extends ChunkedResponseEntities with JsonCodecs with Framing {
```
julienrf, line 120: This one too needs to be labelled `@group algebras`:
```suggestion
/** @group algebras */
trait ChunkedJsonResponseEntities extends ChunkedResponseEntities with JsonCodecs with Framing {
```

algebras/algebra/src/main/scala/endpoints4s/algebra/ChunkedEntities.scala
```diff
@@ -77,24 +94,63 @@ trait ChunkedEntities extends EndpointsWithCustomErrors {
...
+      codec: JsonCodec[A]
+  ): ResponseEntity[Chunks[A]]
 
   /** A response entity carrying chunks of JSON values
     *
     * @tparam A Type of values serialized into JSON
     * @group operations
     */
-  def jsonChunksResponse[A](implicit
+  def jsonChunksResponse[A](framing: Framing)(implicit
```
julienrf, line 135: Here too, it would be good to add a `@param framing` scaladoc documentation.

fetch/client/src/main/scala/endpoints4s/fetch/ReadableStream.scala
```diff
@@ -0,0 +1,253 @@
+package endpoints4s.fetch
+
+import org.scalajs.dom.ReadableStreamReader
+import org.scalajs.dom.WriteableStream
+
+import scala.annotation.nowarn
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+//FIXME copypasta from scalajs-dom https://github.com/scala-js/scala-js-dom/pull/628
```
julienrf, line 10: Should we use the same package, then? `org.scalajs.dom`?

openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
```diff
@@ -40,14 +40,27 @@ trait ChunkedJsonEntities
   private def jsonChunksEntity[A](codec: JsonCodec[A]) =
     Map("application/json" -> MediaType(Some(toSchema(codec.docs))))
```
julienrf, line 41: I think we should document the framing strategy as well.

I don’t remember what is the level of support in OpenAPI to do so, though.

julienrf (commented):
Thank you @ptrdom for the great work!

I have noticed a few places where we could improve even more the documentation. Also, it would be good to add a section about framing here: https://endpoints4s.github.io/algebras/chunked-entities.html

Other than that, the design and implementation look good to me! I see that there are some errors in the binary compatibility checks… I will have a look next week on how to fix them.

akka-http/client/src/main/scala/endpoints4s/akkahttp/client/ChunkedEntities.scala
julienrf: I think what you did is great!

fetch/client/src/main/scala/endpoints4s/fetch/ReadableStream.scala
ptrdom: We do use `org.scalajs.dom.ReadableStream[A]` for `Chunks[A]`, so to use this package for internal classes we would need to rename them too, right? What do we gain here then?

openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
ptrdom: I am not too experienced with OpenAPI, but there still is an open PR here for this https://github.com/OAI/OpenAPI-Specification/issues/1576. So we do have to come up with something of our own. Should we just add some text to the description field?

Also, `application/json` is an incorrect content type for `ChunkedJsonEntities`, because the request/response bodies are not valid JSON documents.

fetch/client/src/main/scala/endpoints4s/fetch/ReadableStream.scala
julienrf: Hm, I am not sure then, but since the definitions below are all private that’s not important anyway :)

fetch/client/src/main/scala/endpoints4s/fetch/ChunkedEntities.scala
```diff
@@ -0,0 +1,118 @@
...
+    )
+  }
+}
+
+trait ChunkedJsonResponseEntities
+    extends algebra.ChunkedJsonResponseEntities
+    with ChunkedResponseEntities
+    with JsonEntitiesFromCodecs {
+
+  type Framing = dom.ReadableStream[Uint8Array] => dom.ReadableStream[Uint8Array]
```
ptrdom, line 63: Now this only represents response chunk framing, because requests are not yet supported by browsers, but maybe it would make sense to make the type more flexible here, so we would not be required to introduce a breaking change once the browser support finally comes around? Or does it not make a lot of sense to think this far ahead? 😄

openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
julienrf: Thank you for the link!

I agree that `application/json` is wrong (unless we used a framing strategy that wraps the whole stream into a JSON array, where frames would be items of the array). I am not sure what we should use instead?

It seems the support for streaming in OpenAPI is very poor, so maybe there is very little we can do about it…

fetch/client/src/main/scala/endpoints4s/fetch/ChunkedEntities.scala
julienrf: It is OK to think far ahead :)

So far, we used to be less conservative about breaking changes in interpreters, though, but it’s always good (for the end-users) to reduce the amount of breaking changes!

You could do this to support future extensions:

```suggestion
  class Framing(val response: dom.ReadableStream[Uint8Array] => dom.ReadableStream[Uint8Array])
```

openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
ptrdom: Hmm, maybe we could express the content type's dependency on framing then? For example, `newLineDelimiterFraming` could set it to something like `text/plain`, but if someone would implement the JSON array framing, then it could set it to `application/json`?

fetch/client/src/main/scala/endpoints4s/fetch/ChunkedEntities.scala
ptrdom: And this type does not need to be moved to its own trait?

julienrf: I don’t think it _needs_ to, but feel free to do so, if you think that will help!

openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
julienrf: I suggest revisiting this later if someone actually implements a JSON array framing strategy.

algebras/testkit/src/main/scala/endpoints4s/algebra/client/ChunkedJsonEntitiesTestSuite.scala
```diff
@@ -68,17 +48,12 @@ trait ChunkedJsonEntitiesResponseTestSuite[T <: ChunkedJsonEntitiesTestApi]
...
+  "Succeed if server streams no chunks" in {
 
-    val expectedItems = List(Array[Byte](1, 2, 3), Array[Byte](4, 5, 6))
+    val expectedItems = List.empty
 
-    callStreamedEndpoint(uploadEndpointTest, expectedItems)
-      .map(_ => succeed)
+    callStreamedEndpoint(streamedEndpointEmptyTest, ())
+      .map(_ shouldEqual expectedItems)
   }
```
ptrdom, line 58: There can never be enough tests for silly errors - noticed this one while playing around with Fetch `ChunkedEntities` 😄

ptrdom:
> I have noticed a few places where we could improve even more the documentation. Also, it would be good to add a section about framing here: https://endpoints4s.github.io/algebras/chunked-entities.html

Should we alter the existing documentation for `ChunkedJsonEntities` or just add an additional section for `Framing`? I feel that with it being so critical for browser clients we should alter the existing docs, describing what problem does it solve, maybe mentioning the deprecation of methods without framing specified. Not sure how to best approach this. Any suggestions @julienrf?

julienrf:
@ptrdom Yes, I think it is a good idea to alter the existing docs to explain the problem you solved. I don’t think we should mention the deprecated method, though (but we will mention that deprecation in the release notes). How does that sound?

ptrdom:
That makes a lot of sense, I will do that, thanks!

documentation/manual/src/main/paradox/algebras/chunked-entities.md
```diff
@@ -32,7 +32,11 @@ which is supported by most HTTP 1.1 clients and servers.
 Clients and servers have to agree on the serialization format used by response
 chunks and WebSocket messages. The `ChunkedJsonEntities` module provides a `jsonChunksRequest`
 constructor and a `jsonChunksResponse` constructor for defining request entities and response
-entities carrying streams of values that are serialized into JSON:
+entities carrying streams of values that are serialized into JSON. In addition, entity constructors
+must also be provided with an implementation of chunk framing, which is defined by the `Framing` module.
+This is required to compensate for the fact that the chunks of “chunked transfer-encoding” can be re-framed 
+during transport, and thus cannot solely be relied on as a method of framing. 
+The library provides the constructor for `newLineDelimiterFraming`:
```
ptrdom, line 39: @julienrf What is your opinion, is this enough or do we need more docs?

julienrf: Let’s go with that. Thank you!

julienrf (approved):
Thank you for you work @ptrdom!

Issue #977, opened 2021-12-14
Aborting requests in Scala.js client interpreters


ptrdom:
While testing out `ChunkedJsonEntities` with Fetch I noticed that closing the response `ReadableStream` does not close its request. Having best front-end development practices for cleaning up opened resources in mind, endpoint4s should expose `abort` callbacks of both XHR and Fetch, ideally for any kind of endpoint.

References:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort
https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

julienrf:
Closed by #979

PR #979, opened 2021-12-16
Aborting requests in Scala.js client interpreters

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/977.


julienrf (commented):
Thank you @ptrdom, I believe this is going in the right direction. I only have one comment regarding the ergonomics of the API.

Also, you will have to relax the compatibility intention of the xhr and fetch modules.

xhr/client/src/main/scala/endpoints4s/xhr/future/Endpoints.scala
```diff
@@ -17,7 +18,7 @@ trait Endpoints extends xhr.Endpoints with EndpointsWithCustomErrors
 trait EndpointsWithCustomErrors extends xhr.EndpointsWithCustomErrors {
 
   /** Maps `Result` to `Future` */
-  type Result[A] = Future[A]
+  case class Result[A](value: Future[A], abort: js.Function1[Unit, Unit])
```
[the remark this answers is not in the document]
ptrdom: I was thinking that we should not try to make too many decisions about ergonomics, because users of the library might want to handle the calls in different ways, especially the streaming calls - a handler utility is basically a must for `ReadableStream` of `ChunkedEntities`.
Even for unary calls one might want to have something like:
```scala
// https://slinky.dev/docs/functional-components-and-hooks/
useEffect(
  () => {
    val abort = client
      .unaryCall(req)
      .handle(onSuccess, onFailure)
    () => abort()
  },
  Seq.empty
)
```
The least opinionated thing we can do is simply exposing the API. Does that make sense?

ptrdom:
Side note - I think multiplexed endpoints do not have tests, so my attempt to untangle the callbacks needs a careful look through.

ptrdom:
> Also, you will have to relax the compatibility intention of the xhr and fetch modules.

Could you elaborate on this? Do we need a major version upgrade?

xhr/client/src/main/scala/endpoints4s/xhr/future/Endpoints.scala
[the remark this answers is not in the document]
julienrf: Yes, but if I understood correctly here the only thing users want to do with a `Result` is to get its underlying `Future` value or maybe call `abort()` on it. I don’t think they will ever want to do `result.copy(abort = () => println())`?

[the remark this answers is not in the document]
ptrdom: Oh yes, sorry for not clarifying - I totally agree with your suggested code bit, I default to case classes too often 😅 I was commenting more on the first part about the cancellable `Future` - seemed to me like the amount of work for benefit gained does not balance out well, or maybe I just do not have a good idea for it.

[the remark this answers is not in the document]
julienrf: No problem! Yes, I agree about the cancellable thing.

julienrf:
> Could you elaborate on this? Do we need a major version upgrade?

Yes, but only in the xhr* and fetch* modules. You can find how to achieve that here: https://github.com/endpoints4s/endpoints4s/blob/master/CONTRIBUTING.md#breakage-policy

julienrf:
> Side note - I think multiplexed endpoints do not have tests, so my attempt to untangle the callbacks needs a careful look through.

I think they are covered by this test: https://github.com/endpoints4s/endpoints4s/blob/master/documentation/examples/cqrs/commands/src/test/scala/cqrs/commands/CommandsTest.scala

fetch/client/src/main/scala/endpoints4s/fetch/Endpoints.scala
```diff
@@ -281,68 +282,75 @@ trait EndpointsWithCustomErrors
       request: Request[A],
       response: Response[B],
       a: A
-  )(
-      onload: Either[Throwable, B] => Unit,
-      onerror: Throwable => Unit
-  ): Unit = {
+  ): (js.Promise[B], js.Function1[Unit, Unit]) = {
```
julienrf, line 285: Should it be?

```suggestion
  ): (js.Thenable[B], js.Function1[Unit, Unit]) = {
```

Maybe it doesn’t matter, I am not familiar with JS promises.

julienrf: Also, maybe a `Function0[Unit]` would work better?

julienrf: Last, I am not sure whether this is important or not, but we previously created either a `Thenable` or a `Future` depending on the concrete xhr interpreter. Now, we always create a `Thenable`, even when we return a `Future`.

xhr/client/src/main/scala/endpoints4s/xhr/Endpoints.scala
```diff
@@ -294,45 +294,49 @@ trait EndpointsWithCustomErrors
       request: Request[A],
       response: Response[B],
       a: A
-  )(
-      onload: Either[Throwable, B] => Unit,
-      onerror: XMLHttpRequest => Unit
-  ): Unit = {
+  ): (js.Promise[B], Unit => Unit) = {
```
julienrf, line 297: Here too, the fact that we always create a `Thenable` which we then convert to something else seems a bit weird.  Would it be possible to make the transformation of the “callback-based” version to the “future-like” version in every concrete interpreter? Maybe it does not really matter, actually, but I was just wondering…

ptrdom: I can see the yuckiness of mandatory intermediary `Promise`, I need to think about it for a bit.

ptrdom: Did I get your idea right?

fetch/client/src/main/scala/endpoints4s/fetch/Endpoints.scala
```diff
@@ -281,68 +283,75 @@ trait EndpointsWithCustomErrors
       request: Request[A],
       response: Response[B],
       a: A
-  )(
-      onload: Either[Throwable, B] => Unit,
-      onerror: Throwable => Unit
-  ): Unit = {
+  ): (FutureLike[B], js.Function0[Unit]) = {
```
julienrf, line 286: I am sorry if I am a bit nitpicky here, but I would rather not introduce the `FutureLike` abstraction. I prefer staying with callbacks here (and every concrete interpreter turns these callbacks into their `Future` thing).

So, the signature of `performFetch` would be the following:

~~~ scala
protected final def performFetch[A, B](
  request: Request[A],
  response: Response[B],
  a: A
)(
  onload: Either[Throwable, B] => Unit,
  onerror: Throwable => Unit
): js.Function0[Unit]
~~~

ptrdom: The reason why I went for the callback refactoring was `Thenable` interpreters - I could not come up with a pretty solution for them. I came up with this:
```scala
var outerResolve: js.UndefOr[js.Function1[B | js.Thenable[B], _]] = js.undefined
var outerError: js.UndefOr[js.Function1[scala.Any, _]] = js.undefined

val promise = new js.Promise[B]((resolve, error) => {
  outerResolve = resolve
  outerError = error
})

val jsAbort = performFetch(this.request, this.response, a)(
  _.fold(
    exn => outerError.asInstanceOf[js.Function1[scala.Any, _]].apply(exn.getMessage),
    b => outerResolve.asInstanceOf[js.Function1[B | js.Thenable[B], _]].apply(b)
  ): Unit,
  ex => outerError.asInstanceOf[js.Function1[scala.Any, _]].apply(ex): Unit
)

new Result(promise) { val abort = jsAbort }
```
I can implement it if it is fine, otherwise I would need some suggestions.

julienrf: Ah, indeed that’s because of how JS promises work. I am not sure we can do it without using an uninitialized variable…

I have pushed 88f5f37 with a design that I like for the xhr interpreter. What do you think? Do you think that would be appropriate for the fetch interpreter as well?

ptrdom: Either the result initialization or the abort callback needs to be moved to a mutable variable(s), I chose the former and you chose the latter 😄 So we can just go your route with Fetch interpreter too.

ptrdom:
@julienrf Do you think this PR is ready to be merged?

documentation/examples/quickstart/client/src/main/scala/quickstart/CounterClientFuture.scala
```diff
@@ -3,17 +3,20 @@ package quickstart
...
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
 object CounterClientFuture
     extends CounterEndpoints
     with xhr.future.Endpoints
     with xhr.JsonEntitiesFromSchemas {
 
+  implicit val ec: ExecutionContext = ExecutionContext.global
```
julienrf, line 14: Should we use this one instead?

```suggestion
  implicit val ec: ExecutionContext = scala.scalajs.concurrent.JSExecutionContext.queue
```

Or even use scalajs-macrotask-executor?

julienrf: BTW, why is this execution context now necessary?

fetch/client/src/main/scala/endpoints4s/fetch/Endpoints.scala
```diff
@@ -318,7 +322,7 @@ trait EndpointsWithCustomErrors
           .orElse(maybeClientErrors)
           .orElse(maybeServerError) match {
           case None =>
-            onload(Left(new Exception(s"Unexpected response status: ${fetchResponse.status}")))
+            onerror(new Exception(s"Unexpected response status: ${fetchResponse.status}"))
```
julienrf, line 325: That seems to be an important semantic change. Is this intended?

xhr/client/src/test/scala/endpoints4s/xhr/future/XhrClientEndpointsJsonSchemaTest.scala
```diff
@@ -11,11 +11,12 @@ import endpoints4s.xhr.BasicAuthentication
 import endpoints4s.xhr.EndpointsSettings
 import endpoints4s.xhr.JsonEntitiesFromSchemas
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutor
 import scala.concurrent.Future
 import scala.scalajs.concurrent.JSExecutionContext
 
-class TestJsonSchemaClient(val settings: EndpointsSettings)
+class TestJsonSchemaClient(val settings: EndpointsSettings)(implicit val ec: ExecutionContext)
```
julienrf, line 19: I did not follow why the execution context was necessary here as well.

xhr/client/src/test/scala/endpoints4s/xhr/future/XhrClientEndpointsTest.scala
```diff
@@ -5,11 +5,12 @@ import endpoints4s.xhr.BasicAuthentication
 import endpoints4s.xhr.EndpointsSettings
 import endpoints4s.xhr.JsonEntitiesFromCodecs
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutor
 import scala.concurrent.Future
 import scala.scalajs.concurrent.JSExecutionContext
 
-class TestClient(val settings: EndpointsSettings)
+class TestClient(val settings: EndpointsSettings)(implicit val ec: ExecutionContext)
```
julienrf, line 13: And here too

xhr/client-faithful/src/main/scala/endpoints4s/xhr/faithful/Endpoints.scala
```diff
@@ -3,14 +3,18 @@ package endpoints4s.xhr.faithful
...
 /** Implements [[xhr.Endpoints]] by using faithful.
   *
   * @group interpreters
   */
 trait Endpoints extends xhr.Endpoints {
 
   /** Maps `Result` to `Future` */
-  type Result[A] = Future[A]
+  abstract class Result[A](val future: Future[A]) {
+    val abort: js.Function0[Unit]
```
julienrf, line 16: May I request to change this one to:

```suggestion
    def abort(): Unit
```

For the sake of consistency with the other interpreters?

julienrf (commented):
Thank you @ptrdom for your patience!

I reviewed the whole PR and there are a couple of things that I think should be changed before we merge it.

documentation/examples/quickstart/client/src/main/scala/quickstart/CounterClientFuture.scala
ptrdom: `ExecutionContext`s are leftovers from `FutureLike`. Can be removed.

fetch/client/src/main/scala/endpoints4s/fetch/Endpoints.scala
ptrdom: Another leftover from `FutureLike`. It probably would have been a good idea to simply rollback `FutureLike` changes with Git instead of manually removing them 😅 

I will roll this back too for uniformity with XHR, but both `onerrror` and `onload(Left)` end up as `failure` in concrete interpreters for `Future` and `Promise`, so maybe the current callback abstractions are too powerful for the way they end up being used.

fetch/client/src/main/scala/endpoints4s/fetch/Endpoints.scala
```diff
@@ -31,7 +31,6 @@ trait EndpointsWithCustomErrors
     with StatusCodes {
 
   def settings: EndpointsSettings
-  implicit def ec: ExecutionContext
```
ptrdom, line 34: We actually did not notice this when finalizing https://github.com/endpoints4s/endpoints4s/pull/934 - after the removal of conversions of `Promise` to `Future`, the `ExecutionContext` became unused.

julienrf (approved):
Thank you @ptrdom!

PR #1017, opened 2022-02-23
Add methods Validated.traverse and Validated.sequence to validate a collection of values

GerretS:
This way, you can validate json arrays by building a validator for their individual objects and then mapping the result through this function.

### DIFF
--- a/json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
@@ -1,5 +1,6 @@
 package endpoints4s
 
+import scala.collection.compat._
 import scala.util.{Failure, Success, Try}
 
 /** A validated value of type `A` can either be `Valid` or `Invalid`
@@ -101,7 +102,8 @@ object Validated {
       case Right(a)     => Valid(a)
     }
 
-  /** Turns a `Success[A]` into a `Validated[A]`
+  /** Turns a `Success[A]` into a `Valid[A]`
+    *
     * Turns a `Failure[A]` into an invalid value, using the exception message as an 'error' message
     */
   def fromTry[A](tryA: Try[A]): Validated[A] =
@@ -109,5 +111,60 @@ object Validated {
       case Success(value) => Valid(value)
       case Failure(error) => Invalid(error.getMessage)
     }
+
+  /** Turns a collection of valid values into a valid collection of values.
+    * If all the values in the collection are `Valid`, the result will be `Valid`.
+    * If there is at least one `Invalid`, the result will be an `Invalid` containing the error messages of all `Invalid`s
+    * If the collection is empty, it will return a `Valid` empty collection.
+    *
+    * @see [[scala.concurrent.Future.sequence]]
+    *
+    * @param in   the collection of `Validated` values that will be sequenced
+    * @tparam A   the type of the `Validated` values (e.g. `Int`)
+    * @tparam CC  the type of the collection of `Validated` values (e.g. `List`)
+    * @return     the `Validated` of the resulting collection
+    */
+  def sequence[A, CC[X] <: IterableOnce[X]](
+      in: CC[Validated[A]]
+  )(implicit bf: BuildFrom[CC[Validated[A]], A, CC[A]]): Validated[CC[A]] = {
+    val builder = bf.newBuilder(in)
+
+    in.iterator
+      .foldLeft[Validated[Unit]](Valid(())) { case (folded, next) =>
+        folded.zip(next).map { n => builder += n; () }
+      }
+      .map(_ => builder.result())
+  }
+
+  /** Validates a collection of values using the provided function `A => Validated[B]` to validate every value.
+    * If all the values in the collection are `Valid`, the result will be `Valid`.
+    * If there is at least one `Invalid`, the result will be an `Invalid` containing the error messages of all `Invalid`s
+    * If the collection is empty, it will return a `Valid` empty collection.
+    *
+    *  {{{
+    *    val myValidatedList = Validated.traverse(myList)(x => myValidateFunc(x))
+    *  }}}
+    *
+    * @see [[scala.concurrent.Future.traverse]]
+    *
+    * @param in        the collection of values to be validated with the provided function
+    * @param fn        the validation function
+    * @tparam A        the type of the values to be validated (e.g. `String`)
+    * @tparam B        the type of the returned validated values (e.g. `Int`)
+    * @tparam CC       the type of the collection of `Validated` (e.g. `List`)
+    * @return          the `Validated` of the resulting collection
+    */
+  def traverse[A, B, CC[X] <: IterableOnce[X]](
+      in: CC[A]
+  )(fn: A => Validated[B])(implicit bf: BuildFrom[CC[A], B, CC[B]]): Validated[CC[B]] = {
+    val builder = bf.newBuilder(in)
+
+    in.iterator
+      .foldLeft[Validated[Unit]](Valid(())) { case (folded, next) =>
+        folded.zip(fn(next)).map { n => builder += n; () }
+      }
+      .map(_ => builder.result())
+
+  }
 
 }
--- a/build.sbt
+++ b/build.sbt
@@ -38,7 +38,7 @@ ThisBuild / sonatypeProjectHosting := Some(
 // We want to keep binary compatibility as long as we can for the algebra,
 // but it is OK to publish breaking releases of interpreters. So,
 // interpreter modules may override this setting.
-ThisBuild / versionPolicyIntention := Compatibility.BinaryAndSourceCompatible
+ThisBuild / versionPolicyIntention := Compatibility.BinaryCompatible
 // Ignore dependencies to modules with version like `1.2.3+n`
 ThisBuild / versionPolicyIgnoredInternalDependencyVersions := Some("^\\d+\\.\\d+\\.\\d+\\+n".r)
 // Default version, used by the algebra modules, and by the interpreters,
--- a/json-schema/json-schema/src/test/scala/endpoints4s/ValidatedTests.scala
+++ b/json-schema/json-schema/src/test/scala/endpoints4s/ValidatedTests.scala
@@ -1,0 +1,205 @@
+package endpoints4s
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import scala.util.{Failure, Success}
+
+class ValidatedTests extends AnyWordSpec with Matchers {
+
+  sealed trait FooBar
+
+  case class Foo(content: String) extends FooBar {
+    def toBar: Bar = Bar(content)
+  }
+
+  case class Bar(content: String) extends FooBar
+
+  private val inputFoo = Foo("input")
+  private val inputBar = Bar("input")
+
+  private val validFoo = Valid(inputFoo)
+  private val validBar = Valid(inputBar)
+  private val invalid: Validated[Foo] = Invalid(Seq("Error1", "Error2"))
+
+  "fold" should {
+
+    "apply the valid function when used on a Valid" in {
+      validFoo.fold(foo => foo.content, _ => "invalid") shouldBe "input"
+    }
+
+    "apply the invalid function when used on an Invalid" in {
+      invalid.fold(foo => foo.content, _ => "invalid") shouldBe "invalid"
+    }
+
+  }
+
+  "map" should {
+
+    "map the value in a valid to the result of the given function" in {
+      validFoo.map(_.toBar) shouldBe Valid(inputBar)
+    }
+
+    "leave an Invalid untouched" in {
+      invalid.map(_.toBar) shouldBe invalid
+    }
+
+  }
+
+  "mapErrors" should {
+
+    "leave a Valid untouched" in {
+      validFoo.mapErrors(_.map(_ + "a")) shouldBe validFoo
+    }
+
+    "map the errors in an Invalid to the result of the given function" in {
+      invalid.mapErrors(_.map(_ + "a")) shouldBe Invalid(Seq("Error1a", "Error2a"))
+    }
+
+  }
+
+  "flatMap" should {
+
+    "validate the value in a Valid" in {
+      validFoo.flatMap(foo => Valid(foo.toBar)) shouldBe validBar
+      validFoo.flatMap(_ => Invalid("incorrect")) shouldBe Invalid("incorrect")
+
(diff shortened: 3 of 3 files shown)


bmeesters:
Small remark: `Validated.sequence` would be consistent with other APIs such as used in `Future` and cats library.

GerretS:
> Small remark: `Validated.sequence` would be consistent with other APIs such as used in `Future` and cats library.

I am aware. I chose `fromSeq` for consistency with the other `from` functions in the same object. I have no strong preference either way.

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +110,18 @@ object Validated {
       case Failure(error) => Invalid(error.getMessage)
     }
 
+  /** Turns a `Seq[Validated[A]]`into a `Validated[Seq[A]]`.
+    * If all values in the sequence are `Valid`, the result will be `Valid`
+    * If there is at least one `Invalid`, the result will be an `Invalid` containing the error messages of all `Invalid`s
+    * If the `Seq` is empty, will return a `Valid[Seq.empty]`
+    */
+  def fromSeq[A](seq: Seq[Validated[A]]): Validated[Seq[A]] = {
```
julienrf, line 118: I think we should have a signature similar to the one of `Future.sequence`, to support any collection, not just `Seq`:

https://github.com/scala/scala/blob/99de4622e8e81e0adc4d4ea60f81bc42d40f84b9/src/library/scala/concurrent/Future.scala#L709

That would also be more efficient (constructing a collection with a `Builder` is the best approach to be collection agnostic).

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +110,18 @@ object Validated {
...
+    seq.foldLeft(Valid(Seq.empty[A]): Validated[Seq[A]]) {
+      case (folded: Valid[Seq[A]], Valid(nextValid)) => folded.map(_ :+ nextValid)
+      case (_: Valid[Seq[A]], invalid: Invalid) => invalid
+      case (folded: Invalid, _: Valid[A]) => folded
+      case (Invalid(foldedErrors), Invalid(nextErrors)) => Invalid(foldedErrors ++ nextErrors)
```
julienrf, line 123: I think it would be more readable to use `zip` and `map` instead of re-implementing the behavior of `zip` here, what do you think?

julienrf (commented):
Thank you @GerretS, this is a great addition!

I agree with the suggestion, it should probably be named `sequence` (and I believe `traverse` would be a useful addition too).

Could you also add some tests?

GerretS:
Thanks, some good suggestions. I'll look into this when I have some time.

The only thing I couldn't really figure out where to add tests - looking around the codebase I see all sorts of different tests in the different modules. Are there any guidelines for this?

GerretS:
I've made the sequence function more generic.

In fact, I've tried to make it mirror Future.sequence as much as possible because people are likely to be familiar with that.

I also implemented traverse while I was at it and added unit tests for the entire Validated class.

There's one thing I ran into which I don't really understand.
For some reason, the signature for Future.sequence has a type argument `CC` for the incoming collection and `To` for the returned collection. Future.traverse uses M for both collections.

I decided to keep it like this in my implementation but I don't really understand why this difference exists. It's not like you can just use any collection type for `To`, if it doesn't match `CC`, the compiler will simply say it can't build a To from that type. I was wondering if you have any clue @julienrf ?

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,60 @@ object Validated {
...
+    *
+    * @see [[scala.concurrent.Future.sequence]]
+    *
+    * @param in   the `IterableOnce` of Validateds which will be sequenced
+    * @tparam A   the type of the value inside the Validateds
+    * @tparam CC  the type of the `IterableOnce` of Validateds
+    * @tparam To  the type of the resulting collection
+    * @return     the `Validated` of the resulting collection
+    */
+  def sequence[A, CC[+X] <: IterableOnce[X], To](
```
GerretS, line 128: Note that CC needs to be covariant in X, otherwise the function would refuse a `Seq[Valid]` without an explicit cast.

julienrf: What is the error message? This is weird since in `traverse` the `M` type parameter is invariant and that does not seem to be an issue.

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,60 @@ object Validated {
...
+    * @see [[scala.concurrent.Future.traverse]]
+    *
+    * @tparam A        the type of the value inside the collection
+    * @tparam B        the type of the value of the returned Validateds
+    * @tparam M        the type of the collection of Validateds
+    * @param in        the collection to be mapped over with the provided function to produce a collection of Validateds that is then sequenced into a Validated collection
+    * @param fn        the function to be mapped over the collection to produce a collection of Validateds
+    * @return          the Validated of the collection of results
+    */
+  def traverse[A, B, M[X] <: IterableOnce[X]](
```
julienrf, line 158: I think this one too should return a `Validated[To]` so that we can return a `Map`, for instance:

~~~ scala
val validatedMaps: Validated[Map[Int, Int]] =
  Validated.traverse(List(1, 2, 3))(x => Valid(x -> x * x))
~~~

julienrf (approved):
Thank you @GerretS for all the added tests!

To be honest I don’t know why `Future.traverse` returns `M[B]` instead of `To`… I believe this is an oversight.

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
GerretS: Removing the `+` causes this error in my tests:

```
[error] .../endpoints4s/json-schema/json-schema/src/test/scala/endpoints4s/ValidatedTests.scala:155:17: no type parameters for method sequence: (in: CC[endpoints4s.Validated[A]])(implicit bf: collection.compat.BuildFrom[CC[endpoints4s.Validated[A]],A,To]): endpoints4s.Validated[To] exist so that it can be applied to arguments (Seq[endpoints4s.Valid[Product with ValidatedTests.this.FooBar with java.io.Serializable]])
[error]  --- because ---
[error] argument expression's type is not compatible with formal parameter type;
[error]  found   : Seq[endpoints4s.Valid[Product with ValidatedTests.this.FooBar with java.io.Serializable]]
[error]  required: ?CC[endpoints4s.Validated[?A]]
[error]       Validated.sequence(Seq(validFoo, validBar, validFoo)) shouldBe Valid(
[error]                 ^
[error] .../endpoints4s/json-schema/json-schema/src/test/scala/endpoints4s/ValidatedTests.scala:155:29: type mismatch;
[error]  found   : Seq[endpoints4s.Valid[Product with ValidatedTests.this.FooBar with java.io.Serializable]]
[error]  required: CC[endpoints4s.Validated[A]]
[error]       Validated.sequence(Seq(validFoo, validBar, validFoo)) shouldBe Valid(
[error]                             ^
[error] two errors found
```

It succeeds again if I make the type of Seq explicit:

```
    "turn an IterableOnce of all Valids into a Valid of an iterable" in {
      Validated.sequence(Seq[Validated[_]](validFoo, validBar, validFoo)) shouldBe Valid(
        Seq(inputFoo, inputBar, inputFoo)
      )
    }
```

julienrf: Ah, I see. I would say this is okay to have to explicitly write `Validated.sequence(Seq[Validated[FooBar]](…))` because the problem in my opinion is due to a limitation of the type inference capabilities of the Scala 2 compiler. I wonder if the type annotation would be unnecessary with Scala 3.

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
GerretS: I just tried, but this doesn't actually work. It returns `Validated[List[(Int, Int)]]`.

I'm not entirely sure, but I think this might be a result of using the Builder, the `zip` only exists to handle the invalid cases, and the map only does the side-effect of adding a value to the mutable builder. That's why I explicitly have the map call return Unit, to clarify that this is what happens. I could've built a Validated.foreach() and use that for the same result.

In my first attempt, the map returned the builder instance itself, but that doesn't change the fact that we depend on a side effect here which means the Tupler can't do anything useful.

julienrf: Ah, OK. Indeed, for a `Map` to be returned we need something more sophisticated.

Hmm, but I believe there is a good reason for having this `To` parameter in `Future.sequence`… I don’t know what is it so maybe it’s simpler to use `CC[_]` everywhere (and to get rid of `To` in `sequence`)?

GerretS: If you mean I should do this:

```
  def sequence[A, CC[_] <: IterableOnce[_]](
      in: CC[Validated[A]]
  )(implicit bf: BuildFrom[CC[Validated[A]], A, CC[A]]): Validated[CC[A]] = {
    val builder = bf.newBuilder(in)

    in.iterator
      .foldLeft[Validated[Unit]](Valid(())) { case (folded, next) =>
        folded.zip(next).map { n => builder += n; () }
      }
      .map(_ => builder.result())
  }
```

This gives the compiler errors:

```
[error] /home/gerret/IdeaProjects/endpoints4s/json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala:133:8: value iterator is not a member of type parameter CC[endpoints4s.Validated[A]]
[error]     in.iterator
[error]        ^
[error] /home/gerret/IdeaProjects/endpoints4s/json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala:135:16: value zip is not a member of Any
[error]         folded.zip(next).map { n => builder += n; () }
[error]                ^
```

julienrf: The error message is confusing but I think if we do:

~~~ diff
- def sequence[A, CC[_] <: IterableOnce[_]]
+ def sequence[A, CC[X] <: IterableOnce[X]]
~~~

That should solve the issue (at least I just tried with Scala 3, I hope this is okay for the other versions of the compiler)

GerretS: Alright, in my latest commit it's now consistent between both functions.

julienrf (approved):

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
       case Failure(error) => Invalid(error.getMessage)
     }
 
+  /** Turns an `IterableOnce[Validated[A]]`into a `Validated[IterableOnce[A]]`.
```
julienrf, line 115:
```suggestion
  /** Turns a collection of valid values into a valid collection of values.
```

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
...
+    * If all values in the sequence are `Valid`, the result will be `Valid`
```
julienrf, line 116:
```suggestion
    * If all the values in the collection are `Valid`, the result will be `Valid`.
```

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
...
+    * If there is at least one `Invalid`, the result will be an `Invalid` containing the error messages of all `Invalid`s
+    * If the `Seq` is empty, will return a `Valid[Seq.empty]`.
```
julienrf, line 118:
```suggestion
    * If the collection is empty, it will return a `Valid` empty collection.
```

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
...
+    *
+    * @see [[scala.concurrent.Future.sequence]]
+    *
+    * @param in   the `IterableOnce` of Validateds which will be sequenced
```
julienrf, line 122:
```suggestion
    * @param in   the collection of `Validated` values that will be sequenced
```

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
...
+    * @tparam A   the type of the value inside the Validateds
```
julienrf, line 123:
```suggestion
    * @tparam A   the type of the `Validated` values (e.g. `Int`)
```

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
...
+    * @tparam CC  the type of the collection of Validateds
```
julienrf, line 124:
```suggestion
    * @tparam CC  the type of the collection of `Validated` values (e.g. `List`)
```

json-schema/json-schema/src/main/scala/endpoints4s/Validated.scala
```diff
@@ -110,4 +112,58 @@ object Validated {
...
+    val builder = bf.newBuilder(in)
+
+    in.iterator
+      .foldLeft[Validated[Unit]](Valid(())) { case (folded, next) =>
+        folded.zip(next).map { n => builder += n; () }
+      }
+      .map(_ => builder.result())
+  }
+
+  /** Transforms a `IterableOnce[A]` into a `Validated[IterableOnce[B]]`
```
julienrf, line 139: Could you please apply similar changes here as well?

julienrf (commented):
Before we merge, I would like to suggest a few changes regarding the wording in the documentation (e.g. replace `IterableOnce` with “collection”). Do you mind also squashing your commits into a single one?

julienrf:
> This can be resolved by the user by just having the return type explicitly be Validated, so I don't think we should change anything for this.

I agree

julienrf:
Thank you for the contribution! It is available in version 1.8.0

PR #1018, opened 2022-02-28
[Draft] `oneOf` support for query strings

GrafBlutwurst:
As mentioned in gitter this tackles the oneOf support for query string. But I'll need some help with a couple bits.

1. OpenApi Generation: I am unsure what the correct way to flatten out nested options is.
2. Testing. I haven't quite found in which endpoints definition to hook this in (besides openapi) to make sure it works for all the implementations
3. Sealed Trait encodings? Is there a sensible way to have those in headers or is it strictly enums? Currently the encoding doesn't stop you from doing it since it's a simple "find first success" decoding strategy. So it has no tagging support or anything like that. I am unsure what exactly we need here.

Formatting and cleaning is not done yet as it's still a draft. 

I also added a possible encoding of `Alternative` and anolog to `Tupler` that could help people have less issues with heavily nested eithers. If we decide to retain this I'll probably use it for a bit more sizable test for query string enums.

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
@@ -83,6 +83,20 @@ trait Urls extends algebra.Urls {
   )
   type QueryStringParam[A] = DocumentedQueryStringParam[A]
 
+  def oneOfQueryStringParam[A,B](qspa: QueryStringParam[A], qspb: QueryStringParam[B]): QueryStringParam[Either[A,B]] = DocumentedQueryStringParam[Either[A,B]](
+    schema = Schema.OneOf(
+      alternatives = Schema.EnumeratedAlternatives(
+        qspa.schema :: qspb.schema :: Nil 
+      ),
+      description = None, //TODO: What to put here?
+      example = None, //TODO: What to put here?
+      title = None,//TODO: What to put here?
+    ),
+    isRequired = qspa.isRequired || qspb.isRequired, //TODO: This feels wrong?
+    encoder = ???, 
+
+  )
+
   implicit def optionalQueryStringParam[A](implicit
       param: QueryStringParam[A]
   ): QueryStringParam[Option[A]] =
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
@@ -83,19 +83,31 @@ trait Urls extends algebra.Urls {
   )
   type QueryStringParam[A] = DocumentedQueryStringParam[A]
 
-  def oneOfQueryStringParam[A,B](qspa: QueryStringParam[A], qspb: QueryStringParam[B]): QueryStringParam[Either[A,B]] = DocumentedQueryStringParam[Either[A,B]](
-    schema = Schema.OneOf(
-      alternatives = Schema.EnumeratedAlternatives(
-        qspa.schema :: qspb.schema :: Nil 
-      ),
-      description = None, //TODO: What to put here?
-      example = None, //TODO: What to put here?
-      title = None,//TODO: What to put here?
-    ),
-    isRequired = qspa.isRequired || qspb.isRequired, //TODO: This feels wrong?
-    encoder = ???, 
+  //TODO: This is kinda ugly. But discriminated schemas make little sense on query strings?
+  //Stil maybe this should be done better, but I haven't yet fully understood the openapi schema to figure out how to deal with nesting here
+  private def unpackSchema[A](qsp:QueryStringParam[A]):List[Schema] = qsp.schema match {
+      case oneOf : Schema.OneOf => oneOf.alternatives match {
+        case enumerated: Schema.EnumeratedAlternatives => enumerated.alternatives 
+        case _ => List(qsp.schema)
+      }
+      case _ => List(qsp.schema)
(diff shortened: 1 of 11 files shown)


GrafBlutwurst:
In particular what is going to need some design-elbow-grease is the question if we want to do this fallback strategy.

Right now there's nothing stopping you from doing something like this:

```
sealed trait Foo
case class Bar(value:String) extends Foo
case class Baz(value:String, value2:Int) extends Foo
```

Note that in terms of structural subtyping (depending on the particular encoding you choose for the case classes) we have `Baz <: Bar` so with a simple fallback decoding, without type tagging, assuming we try `Bar` first, any valid value of `Foo` will be decoded to `Bar`. This is quite the pitfall (though arguable you probably shouldn't encode something like this in a query string anyway). But the question remains how we want to deal with. If we want this to be as safe as `JsonSchema` we'll need an algebra of similar expressive power I think.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
```diff
@@ -83,6 +83,32 @@ trait Urls extends algebra.Urls {
   )
   type QueryStringParam[A] = DocumentedQueryStringParam[A]
 
+  //TODO: This is kinda ugly. But discriminated schemas make little sense on query strings?
+  //Stil maybe this should be done better, but I haven't yet fully understood the openapi schema to figure out how to deal with nesting here
+  private def unpackSchema[A](qsp:QueryStringParam[A]):List[Schema] = qsp.schema match {
```
julienrf, line 88: Indeed, this is a signal that tells us that something is wrong with our approach. Most likely we need an intermediate structure somewhere to avoid this.

openapi/openapi/src/main/scala/endpoints4s/openapi/Urls.scala
```diff
@@ -83,6 +83,32 @@ trait Urls extends algebra.Urls {
...
+    DocumentedQueryStringParam[Either[A,B]](
+      schema = Schema.OneOf(
+        alternatives = Schema.EnumeratedAlternatives(
+          unpackSchema(qspa) ::: unpackSchema(qspb)
+        ),
+        description = None, //TODO: What to put here?
+        example = None, //TODO: What to put here?
+        title = None,//TODO: What to put here?
+      ),
+      isRequired = qspa.isRequired || qspb.isRequired, //TODO: This feels wrong?
```
julienrf, line 106: Here too, this is a signal that the signature of `oneOfQueryStringParam` is wrong, or that the type of `QueryStringParam` is wrong in the openapi interpreter (maybe it should be something else than the current `DocumentedQueryStringParam`).

julienrf (commented):
Thank you @GrafBlutwurst.

May I ask you a bit more details about your use cases? It might be easier to support a special case of `oneOf`: the case of enumerations, where the type of all the alternatives is the same, but the set of valid values is a fixed subset of the possible values for that type.

Regarding the tests, the current approach consists in adding them [here for server interpreters](https://github.com/endpoints4s/endpoints4s/blob/master/algebras/testkit/src/main/scala/endpoints4s/algebra/server/EndpointsTestSuite.scala) and [here for client interpreters](https://github.com/endpoints4s/endpoints4s/blob/master/algebras/testkit/src/main/scala/endpoints4s/algebra/client/UrlEncodingTestSuite.scala).

GrafBlutwurst:
In most cases the usecase is really just straight up enums. So in particular we have a search endpoint where we'd like to have the search parameters in the URL so that this search can easily be bookmarked or shared by users.

Most of the parameters are e.g. `TYPE=x` where `x oneOf (A,B,C)` so fixed values enumerations. That should be easily doable in a way that also lends itself to generating openapi spec (which is what we're very interested in).

The other much harder case is e.g. `DATE=IS_BETWEEN(lower, upper)` where `lower` and `upper` are ISO date strings. And here is the scenario where you'd have a sealed trait that is decoded from a query string because `DATE=IS_NULL` is also fine for us. I'm not even sure if it's possible to have this automatically derived in openapi.

Right now we just work with a stringly types query string that we `xmapPartial` but that suppresses any structural info on the openapi

julienrf:
> Most of the parameters are e.g. `TYPE=x` where `x oneOf (A,B,C)` so fixed values enumerations. That should be easily doable in a way that also lends itself to generating openapi spec (which is what we're very interested in).

For that use case, an approach similar to `enumeration` in `JsonSchemas` should work. What do you think?

> The other much harder case is e.g. `DATE=IS_BETWEEN(lower, upper)` where `lower` and `upper` are ISO date strings. And here is the scenario where you'd have a sealed trait that is decoded from a query string because `DATE=IS_NULL` is also fine for us. I'm not even sure if it's possible to have this automatically derived in openapi.

Here, I am not even sure there is a way to express that in JSON Schema. It seems that “ranges” are supported for numeric values only. I think the only solution is to indicate in the description of the query parameter what are the validation constraints.

GrafBlutwurst:
I think you're right. I'll probably start from scratch again and restrict it to "true" enums. Maybe it would be interesting to explore to "derive" documentation for query params that actually carry data (so like the `DATE`) In our case we have in the backend a filter ADT that we decode from the query strings in an xmap. My current idea is to use the structure of that ADT to dervive the documentation text block explaining what goes into each possibility.

So I think I'll restrict this MR purely to enums and then internally develop the documentation block derivation, if it pans out I'll upstream it.

PR #1019, opened 2022-03-07
Redesign Request implementation in http4s and akka-http interpreters

julienrf:
The previous design was leading to brittle situations (see the changes on the BasicAuthentication interpreters). I believe the new design is simpler (it also requires less boilerplate).

I also changed the behavior of the BasicAuthentication interpreter to reject early incoming requests that don’t provide authentication credentials.

---

I only changed the interpreters’ implementation, not the algebra, which remains unchanged (with `mapRequest`, `mapResponse`, etc.).

For instance, the previous implementation of `Request` in the akka-http-server interpreter was the following:

~~~ scala
  trait Request[A] {
    type HeadersData
    type EntityData
    type UrlData
    def url: Url[UrlData]
    def headers: RequestHeaders[HeadersData]
    def method: Method
    def urlData(a: A): UrlData
    def entity: RequestEntity[EntityData]

    /** A directive that extracts an `A` from an incoming request */
    final lazy val directive: Directive1[A] =
      matchAndParseHeadersDirective.flatMap {
        case invalid: Invalid => handleClientErrors(invalid)
        case Valid((urlData, headersData)) =>
          entity.flatMap { entityData =>
            val validated = aggregateAndValidate(urlData, entityData, headersData)
            validated match {
              case Valid(value)     => Directives.provide(value)
              case invalid: Invalid => handleClientErrors(invalid)
            }
          }
      }

    /** Checks whether the incoming request matches this request description, and
      * if this is the case, parses its URL parameters and headers.
      */
    private[server] def matchAndParseHeadersDirective: Directive1[Validated[(UrlData, HeadersData)]]

    /** The URI of a request carrying the given `a` parameter */
    final def uri(a: A): Uri = url.uri(urlData(a))

    private[server] def aggregateAndValidate(
        urlData: UrlData,
        entityData: EntityData,
        headersData: HeadersData
    ): Validated[A]
  }
~~~

It contains 3 abstract type members, and several abstract methods. Some of them were even `private` (e.g., `aggregateAndValidate`) making it impossible to implement outside of endpoints4s! (I initially wanted to make them private to have some margin for changing the implementation details, but the result is impractical to use)

The new design is the following:

~~~ scala
  trait Request[A] {

    type UrlAndHeaders

    /** A directive that extracts an `A` from an incoming request */
    final lazy val directive: Directive1[A] =
      matchAndParseHeadersDirective.flatMap {
        case invalid: Invalid     => handleClientErrors(invalid)
        case Valid(urlAndHeaders) => parseEntityDirective(urlAndHeaders)
      }

    /** Checks whether the incoming request matches this request description, and
      * if this is the case, parses its URL parameters and headers.
      */
    def matchAndParseHeadersDirective: Directive1[Validated[UrlAndHeaders]]

    /** Parses the request entity.
      */
    def parseEntityDirective(urlAndHeaders: UrlAndHeaders): Directive1[A]

    /** The URI of a request carrying the given `a` parameter */
    def uri(a: A): Uri
  }
~~~

It contains only 1 abstract type member, and has fewer abstract methods. It has a public final member `directive` that provides the logic for processing a request. It works in two stages.

First, we check whether the incoming request matches the endpoint description (by its method and URL), and if this is the case we parse all the information from the URL and headers. This first stage is achieved by the abstract method `matchAndParseHeadersDirective`, which can either immediately return an error response, or provide the information parsed from the URL and headers.

Second, we parse the request entity and we combine it with the information previously parsed from the URL and headers to produce the resulting value of type `A` (for a `Request[A]`). That part is achieved by the abstract method `parseEntityDirective`.

The implementation of `Request[A]` in the http4s interpreter follows a similar design.

---

The concrete implementation of `Request[A]` usually does not matter much to the end users, except when they want to do advanced stuff where they need to implement their own instance of `Request[A]`. This is, for instance, the case with authentication support. In the future, the complete implementation of middlewares should allow the users to implement things like custom authentication even without having to implement their own instances of `Request[A]`, but there will probably be some cases where this will be still necessary.

Let’s look at the changes of the `example-authenticated` module (which uses http4s) to see the improvement. Before the PR, we implemented our instance of `Request` as follows:

~~~ scala
  def authenticatedRequest[U, E, UE, UET](
      method: Method,
      url: Url[U],
      entity: RequestEntity[E]
  )(implicit
      tuplerUE: Tupler.Aux[U, E, UE],
      tuplerUET: Tupler.Aux[UE, AuthenticationToken, UET]
  ): Request[UET] = {
    // alias parameters to not clash with `Request` members
    val urlArg = url
    val methodArg = method
    val entityArg = entity

    new Request[UET] {
      type UrlData = U
      type HeadersData = Option[AuthenticationToken]
      type EntityData = E

      def url: Url[U] = urlArg
      def headers: RequestHeaders[Option[UserInfo]] = authenticationTokenRequestHeaders
      def method: Method = methodArg
      def entity: RequestEntity[E] = entityArg

      def aggregateAndValidate(
          urlData: U,
          headersData: Option[UserInfo],
          entityData: E
      ): Validated[UET] =
        headersData match {
          case None        => sys.error("Unsupported")
          case Some(token) => Valid(tuplerUET(tuplerUE(urlData, entityData), token))
        }

      def matchAndParseHeaders(
          http4sRequest: org.http4s.Request[F]
      ): Option[Either[org.http4s.Response[F], Validated[(UrlData, HeadersData)]]] =
        matchAndParseHeadersAsRight(method, url, headers, http4sRequest).map(_.flatMap {
          case Valid((_, None /* credentials */ )) => Left(org.http4s.Response(Unauthorized))
          case validatedUrlAndHeaders              => Right(validatedUrlAndHeaders)
        })

    }
  }
~~~

In this PR, this is now as follows:

~~~ scala
  def authenticatedRequest[U, E, UE, UET](
      method: Method,
      url: Url[U],
      entity: RequestEntity[E]
  )(implicit
      tuplerUE: Tupler.Aux[U, E, UE],
      tuplerUET: Tupler.Aux[UE, AuthenticationToken, UET]
  ): Request[UET] =
    new Request[UET] {

      // Data extracted from the incoming request
      type UrlAndHeaders = (U, AuthenticationToken)

      def matchAndParseHeaders(
          http4sRequest: org.http4s.Request[F]
      ): Option[Either[org.http4s.Response[F], Validated[UrlAndHeaders]]] =
        // First, check whether the incoming request matches this request description
        matchAndParseHeadersAsRight(method, url, emptyRequestHeaders, http4sRequest)
          // If this is the case, check whether there is a token in the request headers or not
          .map { errorResponseOrValidatedUrlAndHeaders =>
            authenticationTokenRequestHeaders(http4sRequest.headers) match {
              // There is a token, just add it to the
              case Valid(Some(token)) =>
                errorResponseOrValidatedUrlAndHeaders
                  .map(_.map { case (u, _) => (u, token) })
              case _ => Left(org.http4s.Response(Unauthorized))
            }
          }

      def parseEntity(
          urlAndHeaders: UrlAndHeaders,
          http4sRequest: org.http4s.Request[F]
      ): Effect[Either[org.http4s.Response[F], UET]] =
        entity(http4sRequest).map(_.map { entityData =>
          val (urlData, token) = urlAndHeaders
          tuplerUET(tuplerUE(urlData, entityData), token)
        })

    }
~~~

Not only the new implementation is more concise, but it also removes the need to write `case None => sys.error("Unsupported")` that we previously had to handle even though that case was impossible to reach in practice.

With the design proposed in this PR, end-users have two “points of control” when they implement a `Request[A]`, which correspond to the two steps I explained above: checking whether the incoming request matches this request description (and parsing the URL and headers at the same time), or parsing the request entity (and aggregating it with the parsed URL and headers data to produce an `A`). Here, we act on the first step: we checked whether the incoming request matches this request description by calling `matchAndParseHeadersAsRight` and, in case the incoming request does match, we look into its headers to see if there is an authentication token. If it is missing, we can immediately return an error response. Otherwise, we add it to the content of the parsed URL. The implementation of the second stage parses the request entity and then uses the tuplers to build an `A` from the URL data, entity data, and token.

---

Possible variations around the design proposed in the PR:

- [ ] make the abstract methods in `Request[A]` `protected` instead of `public`
- [ ] change signature of `parseEntity` to return a `Validated[A]` instead of an `A`, to make it more straightforward to implement for end-users. Currently, users have to be careful to call `handleClientErrors` to handle validation errors. That could be done for them automatically if validation errors were returned as a `Validated` value.

### DIFF
--- a/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
+++ b/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
@@ -53,20 +53,24 @@ trait BasicAuthentication extends algebra.BasicAuthentication with EndpointsWith
             )
           case Valid(Some(credentials)) =>
             matchAndProvideParsedUrlAndHeadersData(method, url, headers)
-              .map(_.map { case (u, h) => (u, h, credentials) })
+              .map { validatedUrlAndHeaders =>
+                validatedUrlAndHeaders.map { case (urlData, headersData) =>
+                  (urlData, headersData, credentials)
+                }
+              }
           case invalid: Invalid => Directives.provide(invalid)
         }
 
       def parseEntityDirective(urlAndHeaders: UrlAndHeaders): Directive1[Out] =
-        entity.map { e =>
-          val (u, h, c) = urlAndHeaders
-          tuplerUEHCred(tuplerUE(u, e), tuplerHCred(h, c))
+        entity.map { entityData =>
+          val (urlData, headersData, credentials) = urlAndHeaders
+          tuplerUEHCred(tuplerUE(urlData, entityData), tuplerHCred(headersData, credentials))
         }
 
       def uri(out: Out): Uri = {
-        val (ue, _) = tuplerUEHCred.unapply(out)
-        val (u, _) = tuplerUE.unapply(ue)
-        url.uri(u)
+        val (urlAndEntityData, _) = tuplerUEHCred.unapply(out)
+        val (urlData, _) = tuplerUE.unapply(urlAndEntityData)
+        url.uri(urlData)
       }
 
     }
--- a/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
+++ b/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
@@ -6,11 +6,11 @@ import akka.http.scaladsl.model.headers.{
   HttpChallenges,
   `WWW-Authenticate`
 }
-import akka.http.scaladsl.model.{HttpHeader, HttpResponse, StatusCodes => AkkaStatusCodes}
+import akka.http.scaladsl.model.{HttpHeader, HttpResponse, Uri, StatusCodes => AkkaStatusCodes}
 import akka.http.scaladsl.server.{Directive1, Directives}
 import endpoints4s.algebra.BasicAuthentication.Credentials
 import endpoints4s.algebra.Documentation
-import endpoints4s.{Tupler, Valid, Validated, algebra}
+import endpoints4s.{Invalid, Tupler, Valid, Validated, algebra}
 
 /** @group interpreters
   */
@@ -27,33 +27,8 @@ trait BasicAuthentication extends algebra.BasicAuthentication with EndpointsWith
       tuplerHCred: Tupler.Aux[H, Credentials, HCred],
       tuplerUEHCred: Tupler.Aux[UE, HCred, Out]
   ): Request[Out] = {
-    val (m, u, e, h) = (method, url, entity, headers)
     new Request[Out] {
-      type UrlData = U
-      type HeadersData = (H, Option[Credentials])
-      type EntityData = E
-
-      def url: Url[UrlData] = u
-      def method: Method = m
-      def entity: RequestEntity[EntityData] = e
-      def headers: RequestHeaders[HeadersData] = h ++ authHeader
-
-      private[server] def aggregateAndValidate(
-          urlData: UrlData,
-          entityData: EntityData,
-          headersData: HeadersData
-      ): Validated[Out] =
-        headersData match {
-          case (_, None) =>
-            // Note: in practice that should not happen because the method `aggregateAndValidate` is
-            // only called from the final method `matches`, if `matchAndParseHeaders` succeeded.
-            // However, here we override `matchAndParseHeaders` to fail in case the credentials are missing.
-            sys.error(
-              "This request transformation is currently unsupported. You can't transform further an authenticated request."
-            )
-          case (h, Some(credentials)) =>
-            Valid(tuplerUEHCred(tuplerUE(urlData, entityData), tuplerHCred(h, credentials)))
-        }
+      type UrlAndHeaders = (U, H, Credentials)
 
       lazy val authHeader: RequestHeaders[Option[Credentials]] =
         httpHeaders =>
@@ -65,10 +40,9 @@ trait BasicAuthentication extends algebra.BasicAuthentication with EndpointsWith
             }
           )
 
-      private[server] def matchAndParseHeadersDirective
-          : Directive1[Validated[(UrlData, HeadersData)]] =
-        matchAndProvideParsedUrlAndHeadersData(method, url, headers).flatMap {
-          case Valid((_, (_, None /* credentials */ ))) =>
+      lazy val matchAndParseHeadersDirective: Directive1[Validated[UrlAndHeaders]] =
+        Directives.extractRequest.map(authHeader.decode).flatMap {
+          case Valid(None) =>
             Directives.complete(
               HttpResponse(
                 AkkaStatusCodes.Unauthorized,
@@ -77,14 +51,24 @@ trait BasicAuthentication extends algebra.BasicAuthentication with EndpointsWith
                 )
               )
             )
-          case validatedUrlAndHeaders => Directives.provide(validatedUrlAndHeaders)
+          case Valid(Some(credentials)) =>
+            matchAndProvideParsedUrlAndHeadersData(method, url, headers)
+              .map(_.map { case (u, h) => (u, h, credentials) })
+          case invalid: Invalid => Directives.provide(invalid)
         }
 
-      def urlData(out: Out): UrlData = {
+      def parseEntityDirective(urlAndHeaders: UrlAndHeaders): Directive1[Out] =
+        entity.map { e =>
+          val (u, h, c) = urlAndHeaders
+          tuplerUEHCred(tuplerUE(u, e), tuplerHCred(h, c))
+        }
+
+      def uri(out: Out): Uri = {
         val (ue, _) = tuplerUEHCred.unapply(out)
         val (u, _) = tuplerUE.unapply(ue)
-        u
+        url.uri(u)
       }
+
     }
   }
 
--- a/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
+++ b/akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
@@ -34,42 +34,14 @@ trait EndpointsWithCustomErrors
 
   trait Request[A] {
 
-    /** The type parameter of the headers, part of the whole complete input type A. */
-    type HeadersData
-
-    /** The type parameter of the request entity, part of the whole complete input type A. */
-    type EntityData
-
-    /** The type parameter of the headers, part of the whole complete input type A. */
-    type UrlData
-
-    /** The URL component of this request */
-    def url: Url[UrlData]
-
-    /** The headers component of this request */
-    def headers: RequestHeaders[HeadersData]
-
-    /** The method of this request */
-    def method: Method
-
-    /* Extract the UrlData from the complete output type of this request */
-    def urlData(a: A): UrlData
-
-    /** The entity of this request */
-    def entity: RequestEntity[EntityData]
+    /** Information extracted from the URL and the headers */
+    type UrlAndHeaders
 
     /** A directive that extracts an `A` from an incoming request */
     final lazy val directive: Directive1[A] =
       matchAndParseHeadersDirective.flatMap {
-        case invalid: Invalid => handleClientErrors(invalid)
-        case Valid((urlData, headersData)) =>
-          entity.flatMap { entityData =>
-            val validated = aggregateAndValidate(urlData, entityData, headersData)
-            validated match {
-              case Valid(value)     => Directives.provide(value)
-              case invalid: Invalid => handleClientErrors(invalid)
-            }
-          }
+        case invalid: Invalid     => handleClientErrors(invalid)
+        case Valid(urlAndHeaders) => parseEntityDirective(urlAndHeaders)
       }
 
     /** Checks whether the incoming request matches this request description, and
@@ -85,16 +57,21 @@ trait EndpointsWithCustomErrors
       *   - a value `Invalid(errors)` in case the URL and headers had validation
       *     errors.
       */
-    private[server] def matchAndParseHeadersDirective: Directive1[Validated[(UrlData, HeadersData)]]
+    def matchAndParseHeadersDirective: Directive1[Validated[UrlAndHeaders]]
+
+    /** Parses the request entity.
+      *
+      * The directive produces:
+      *
+      *   - a ''completion'' to immediately return a custom response,
+      *   - or, a value of type `A` containing all the information extracted
+      *     from the request (including URL, headers, and entity).
+      */
+    def parseEntityDirective(urlAndHeaders: UrlAndHeaders): Directive1[A]
 
     /** The URI of a request carrying the given `a` parameter */
-    final def uri(a: A): Uri = url.uri(urlData(a))
+    def uri(a: A): Uri
 
-    private[server] def aggregateAndValidate(
-        urlData: UrlData,
-        entityData: EntityData,
-        headersData: HeadersData
-    ): Validated[A]
   }
 
   // Default implementation for `matchAndParseHeadersDirective`, which always
@@ -189,28 +166,22 @@ trait EndpointsWithCustomErrors
 
   implicit def requestPartialInvariantFunctor: PartialInvariantFunctor[Request] =
     new PartialInvariantFunctor[Request] {
-
       def xmapPartial[A, B](fa: Request[A], f: A => Validated[B], g: B => A): Request[B] =
         new Request[B] {
-          type UrlData = fa.UrlData
-          type HeadersData = fa.HeadersData
-          type EntityData = fa.EntityData
+          type UrlAndHeaders = fa.UrlAndHeaders
 
-          def urlData(b: B): UrlData = fa.urlData(g(b))
-          def url: Url[UrlData] = fa.url
-          def method: Method = fa.method
+          def matchAndParseHeadersDirective: Directive1[Validated[UrlAndHeaders]] =
+            fa.matchAndParseHeadersDirective
 
-          def entity: RequestEntity[EntityData] = fa.entity
-          def headers: RequestHeaders[HeadersData] = fa.headers
-          def aggregateAndValidate(
-              urlData: fa.UrlData,
-              entityData: fa.EntityData,
-              headersData: fa.HeadersData
-          ): Validated[B] = fa.aggregateAndValidate(urlData, entityData, headersData).flatMap(f)
+          def parseEntityDirective(urlAndHeaders: UrlAndHeaders): Directive1[B] =
+            fa.parseEntityDirective(urlAndHeaders).flatMap { a =>
+              f(a) match {
+                case Valid(b)         => Directives.provide(b)
+                case invalid: Invalid => handleClientErrors(invalid)
+              }
+            }
 
-          private[server] lazy val matchAndParseHeadersDirective
-              : Directive1[Validated[(UrlData, HeadersData)]] =
-            matchAndProvideParsedUrlAndHeadersData(method, url, headers)
+          def uri(b: B): Uri = fa.uri(g(b))
         }
     }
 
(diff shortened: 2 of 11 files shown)


agboom:
Hey, this is great, thanks for your continued work on middlewares! Do you mind if I take a more detailed look at this in the coming days?

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
```diff
@@ -27,33 +27,8 @@ trait BasicAuthentication extends algebra.BasicAuthentication with EndpointsWith
...
-            // Note: in practice that should not happen because the method `aggregateAndValidate` is
-            // only called from the final method `matches`, if `matchAndParseHeaders` succeeded.
-            // However, here we override `matchAndParseHeaders` to fail in case the credentials are missing.
-            sys.error(
-              "This request transformation is currently unsupported. You can't transform further an authenticated request."
-            )
-          case (h, Some(credentials)) =>
-            Valid(tuplerUEHCred(tuplerUE(urlData, entityData), tuplerHCred(h, credentials)))
-        }
+      type UrlAndHeaders = (U, H, Credentials)
```
bmeesters, line 31: It is nice to see so much boilerplate go, though if I understand the changes correctly we do lose the ability to get the individual parts of a request in the interpreters. I would need to see in our own implementation of the auth to see whether that will give problems.

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
```diff
@@ -85,16 +57,21 @@ trait EndpointsWithCustomErrors
       *   - a value `Invalid(errors)` in case the URL and headers had validation
       *     errors.
       */
-    private[server] def matchAndParseHeadersDirective: Directive1[Validated[(UrlData, HeadersData)]]
+    def matchAndParseHeadersDirective: Directive1[Validated[UrlAndHeaders]]
```
bmeesters, line 60: Making these public is a good idea IMO. We also had to create a fork where we just made these public to implement some custom behavior ourselves.

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
```diff
@@ -77,14 +51,24 @@ trait BasicAuthentication extends algebra.BasicAuthentication with EndpointsWith
                 )
               )
             )
-          case validatedUrlAndHeaders => Directives.provide(validatedUrlAndHeaders)
+          case Valid(Some(credentials)) =>
+            matchAndProvideParsedUrlAndHeadersData(method, url, headers)
+              .map(_.map { case (u, h) => (u, h, credentials) })
```
bmeesters, line 56: It is a little bit of a nitpick. But I find nested maps quite hard to read without explicitly named arguments. I would replace the underscore with a sensible name.

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
```diff
@@ -449,63 +410,45 @@ trait EndpointsWithCustomErrors
...
-      private[server] lazy val matchAndParseHeadersDirective
-          : Directive1[Validated[(UrlData, HeadersData)]] =
-        matchAndProvideParsedUrlAndHeadersData(method, url, headers)
+      def parseEntityDirective(urlAndHeaders: (request.UrlAndHeaders, Q)): Directive1[tupler.Out] =
+        request
+          .parseEntityDirective(urlAndHeaders._1)
+          .map(a => tupler(a, urlAndHeaders._2))
+
+      def uri(out: tupler.Out): Uri =
+        request.uri(tupler.unapply(out)._1) // TODO Add query parameter
```
bmeesters, line 451: I suppose this TODO needs to be fixed before merging?

documentation/examples/authentication/src/main/scala/authentication/Authentication.scala
```diff
@@ -256,37 +257,34 @@ trait ServerAuthentication[F[_]]
...
-        matchAndParseHeadersAsRight(method, url, headers, http4sRequest).map(_.flatMap {
-          case Valid((_, None /* credentials */ )) => Left(org.http4s.Response(Unauthorized))
-          case validatedUrlAndHeaders              => Right(validatedUrlAndHeaders)
+      ): Option[Either[org.http4s.Response[F], Validated[UrlAndHeaders]]] =
+        // First, check whether the incoming request matches this request description
+        matchAndParseHeadersAsRight(method, url, emptyRequestHeaders, http4sRequest)
+          // If this is the case, check whether there is a token in the request headers or not
+          .map { errorResponseOrValidatedUrlAndHeaders =>
+            authenticationTokenRequestHeaders(http4sRequest.headers) match {
+              // There is a token, just add it to the
```
bmeesters, line 273: This sentence is incomplete.

bmeesters (commented):
Overall it seems like a nice simplification. I am wondering though whether we lose any power here? You cannot access the individual fields anymore, but it seems that most is caught by the new methods that can access the information before doing something with it (like parsing entities)? Only `method` is completely gone now, though I don't think we'd use that.

julienrf:
Thank you @agboom for your interest! Yes, I am very looking forward to your review ;)

Please @agboom and @bmeesters let me know what are your use cases that would be impacted by this PR (ie, the parts of your code that has to provide instances of `Request[A]` within server interpreters).

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
julienrf: That’s true, we might lose some power, but I believe in practice that should not be an issue (ie, we used to have “too much power”). But I might be wrong, so please let me know more precisely what are your needs.

The way it works in the use cases that are part of endpoints4s (`mapRequest` method and authentication support) is that we can always access the content of the headers, URL, and entity by calling `someRequest.matchAndParseUrlAndHeadersDirective` or `someRequest.parseEntityDirective`, which return the information parsed from the URL, headers, and entity.

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
julienrf: Yes, I forgot that one. And there is no test that covers it.

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
julienrf: That’s a good point! Addressed in add6b5c.

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/Endpoints.scala
julienrf: I juste pushed a fixup commit to address this point.

documentation/examples/authentication/src/main/scala/authentication/Authentication.scala
julienrf: Good catch, I just pushed a fixup commit that fixes that.

julienrf:
FYI here is the implementation of “high-level middlewares” based on this design: https://github.com/julienrf/endpoints/pull/2/files

akka-http/server/src/main/scala/endpoints4s/akkahttp/server/BasicAuthentication.scala
bmeesters: Yes, I have looked at our implementation of auth based on the current design and I believe the changes in this PR would still make that possible.

julienrf: @bmeesters Happy to know that the changes made in this PR still support your use case. Do you think it makes things simpler or more complex, though?

bmeesters: TBH I don't think it will make much difference. We did need to copy over some package private methods that I suppose are not needed anymore, so that is nice. But our implementation doesn't lean on the types and `url`, `entity` etc. methods of the interpreter `Request`. We instead override the apply method in the `Endpoint` instead to some auth there.

Also for more authorization use case we instead need to do `myEndpoint.handleBearerAuthentication.implementedBy` to ensure we have access in the server to user information after authentication is done. As far as I understood it was very hard to do in the algebra because this type of information is only necessary in the server, and it is in between the request (not an input to the endpoint, but only after handing the request). But I believe the middlewares could make this redundant.

That said, @agboom has been the one implementing auth in our internal library on top of endpoints4s, so he knows all the details that I might have missed.

julienrf: > Also for more authorization use case we instead need to do `myEndpoint.handleBearerAuthentication.implementedBy` to ensure we have access in the server to user information after authentication is done. As far as I understood it was very hard to do in the algebra because this type of information is only necessary in the server, and it is in between the request (not an input to the endpoint, but only after handing the request). But I believe the middlewares could make this redundant.

Yes, maybe, but I am not so sure of that. I would be interested in knowing more about this use case to validate the approach taken here https://github.com/julienrf/endpoints/pull/2.

agboom: I'm working on a barebones Authentication algebra based on our internal library, but adapted to your middlewares approach in https://github.com/julienrf/endpoints/pull/2. After I'm further along I'll share it somewhere in the coming week.

agboom: Hi @julienrf, sorry for my absence. I did make progress on the barebones algebra I mentioned above, but not as much as I wanted. I can push something this evening that will give an idea of my intended use-case.

julienrf: No need to hurry :)

agboom:
@julienrf I pushed a new commit with an `Authentication` algebra that almost fits our use-case based on your middlewares implementation. Almost, because I'm still working on the client part and I had to tweak the middleware code a little. I'll try to break it down here.

My aim of the authentication Middleware is to let it handle the server authentication logic (and later the client logic as well).
Given an `Endpoint[A, B]`, the goal is that at use-site, calling `withAuthentication` results in a signature `Endpoint[(A, AccountData), ValidatedAuthentication[B]]`. The `AccountData` is the data associated with the authenticated credentials, including the credentials themselves, but possibly also authorization data such as groups. This last one is up to the user and depends on the authentication server used.

I wasn't able to achieve this with the existing Middleware, because it needs to short circuit on invalid authentication and provide a response. That's why I created a new type of middleware called `ShortcircuitMiddleware`. It's similar to the original `Middleware`, but adds an error type `E`. The `Conditional` server action accepts an `Either` type so that the server can respond early depending on some logic in the middleware. In the case of authentication, `E` is substituted for `InvalidAuthentication`, i.e. `ForbiddenError` or `UnauthorizedError`.

```scala
trait ShortcircuitMiddleware[E, Request, Response, NewRequest, NewResponse] {
  def serverAction(request: Request): Middleware.Conditional[E, NewRequest]

  def fromNewResponse(newResponse: NewResponse): Response

  def fromNewRequest(newRequest: NewRequest): Request
}

case class Conditional[LeftResponse, NewRequest](
   condition: Either[LeftResponse, NewRequest]
) extends ServerAction[Either[LeftResponse, NewRequest], NewRequest]
```

This allows for the following authentication middleware:

```scala
  def basicAuthenticationMiddleware2[A, B, ACred, AAccountData](implicit
      tuplerACred: Tupler.Aux[A, Option[String], ACred],
      tuplerAAccountData: Tupler.Aux[A, AccountData, AAccountData]
  ): ShortcircuitMiddleware[InvalidAuthentication, ACred, B, AAccountData, B] =
    new ShortcircuitMiddleware[InvalidAuthentication, ACred, B, AAccountData, B] {
      def serverAction(
          request: ACred
      ): Middleware.Conditional[InvalidAuthentication, AAccountData] =
        Middleware.Conditional {
          val (a, credOpt) = tuplerACred.unapply(request)
          credOpt match {
            case Some(cred) =>
              decodeBasicCredentials(cred) match {
                case Valid(basicCredentials) =>
                  validateBasicCredentials(basicCredentials)
                    .map(tuplerAAccountData.apply(a, _))
                case Invalid(_) =>
                  Left(UnauthorizedError)
              }
            case None =>
              Left(UnauthorizedError)
          }
        }

      def fromNewRequest(newRequest: AAccountData): ACred = {
        val (a, accountData) = tuplerAAccountData.unapply(newRequest)
        tuplerACred(a, Some(extractCredentials(accountData)))
      }

      def fromNewResponse(newResponse: B): B = newResponse
    }
```

I've written a [server test case](https://github.com/julienrf/endpoints/compare/even-more-middlewares...agboom:even-more-middlewares?expand=1#diff-cd2000854a291a2c67ccd3d58acf58653626817384759e5000e00b84bf767366R7) to check that it works for the server.

There are still some holes in the implementation:
- The client implementation is non-existent in this commit, but I'm working on it right now. This will probably require an additional function to the middleware called `clientAction` or something. It will also change some of the types in the algebra. For example, `AccountData` as input type doesn't make sense for the client, but that will have to be the input credentials. I'm still trying to find a good name that covers both server and client, but the best I've come up with is `InputCredentials`. 
- The `ShortcircuitMiddleware` exists next to the original `Middleware`. I'm pretty sure it's possible to merge these two, but I don't know if both `serverAction`s are compatible in type.
- The middleware signature lacks support for effectful actions. It's pretty common to have authentication done by an external server, either on the client side or the server side. In our case I've seen both. So adding support for effectful actions would be a valuable addition in my opinion.
- It might have merit to abstract the `Either` from the `Conditional` server action and just allow any monadic structure. This might be a step-up to effectful server actions, but I'm not sure until I've tried. It would require some type class or similar signature added to the mix. 

I'll keep working to make the implementation more complete. I'm curious to know what you think so far.

agboom (approved):

agboom:
As far as this PR goes, I think that it's a great improvement! It's compatible with our current implementation and it makes your middlewares approach possible :ok_hand:

julienrf:
Thank you @agboom for your extended tests and review! I am merging this PR but I take note of your points regarding middleware implementation. In particular:

> * The middleware signature lacks support for effectful actions. It's pretty common to have authentication done by an external server, either on the client side or the server side. In our case I've seen both. So adding support for effectful actions would be a valuable addition in my opinion.

I think this is an important limitation, and I am not sure what we can do about it… I have found a couple of other limitations, and I am now less convinced of the value of such middlewares.

agboom:
Great that this PR is merged :tada:

> I think this is an important limitation, and I am not sure what we can do about it… I have found a couple of other limitations, and I am now less convinced of the value of such middlewares.

Can you elaborate in what way you're less convinced? What other limitations have you found? I actually like the approach you've taken with these declarative `Continue` and `Bypass` types and I think if we could somehow add the `Conditional` server action and a client action then we're an effect type short of a complete middleware structure.

Maybe it would help to create an issue for this, to continue the discussion?

julienrf:
The other “limitation” I have in mind is that, unlike most middlewares in HTTP servers, there is no simple way to “wrap” the server logic like the following, for instance:

~~~ scala
(request, logic) => {
  val startTime = System.currentTimeMillis()
  val response = logic(request)
  logger.info(System.currentTimeMillis() - startTime)
  response
}
~~~

Implementing a “metrics” middleware might be complicated.

But maybe we just didn’t find yet the right design…

agboom:
Good point. In my limited view that would just be a variant of a `ServerAction` with the right type signature. But you're probably right that there is no simple way right now.

I can do an attempt at a crude logging middleware to find out in more detail what is needed for this use-case. Although I'd like to get further with the auth middleware first, by creating the client part. Maybe we can do this in multiple steps? Should I create a separate issue for this?

julienrf:
Sure, please go ahead with the authentication use-case :) Would it be possible to work on it outside of endpoints4s first? I think it would be simpler and safer to first experiment outside of endpoints4s until you find a design that you are happy with, and then we can consider whether we want to include it or not.

agboom:
Yes, I think that's a sensible approach. I think experimenting with multiple use-cases will help get the general middleware implementation right, but the use-cases themselves don't necessarily have to be included in endpoints4s imho. Although it might help to have some examples in the same project to keep track of regressions, but that's a problem for later (or for unit tests).

PR #1042, opened 2022-06-29
Add timeout in clients

romainreuillon:
Here is the code to make it possible to set a timeout value in xhr.

Any comment or improvement to make to make it in the repo?

Best,
Romain

### DIFF
--- a/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
+++ b/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
@@ -13,7 +13,7 @@ trait ClientEndpointsTestApi extends EndpointsTestApi {
 
 trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] {
 
-  def clientTestSuite() = {
+  def clientTestSuite(timeoutTest: Boolean = true) = {
 
     "Client interpreted" should {
 
@@ -81,10 +81,12 @@ trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] 
         } yield succeed
       }
 
-      "timeout" in {
-        for {
-          _ <- call(client.slowResponseEndpoint, ()).failed.map(_ shouldBe a [scala.concurrent.TimeoutException])
-        } yield succeed
+      if(timeoutTest) {
+        "timeout" in {
+          for {
+            _ <- call(client.slowResponseEndpoint, ()).failed.map(_ shouldBe a[scala.concurrent.TimeoutException])
+          } yield succeed
+        }
       }
 
       "properly handle joined headers" in {
--- a/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
+++ b/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
@@ -13,7 +13,7 @@ trait ClientEndpointsTestApi extends EndpointsTestApi {
 
 trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] {
 
-  def clientTestSuite(timeoutTest: Boolean = true) = {
+  def clientTestSuite() = {
 
     "Client interpreted" should {
 
@@ -79,14 +79,6 @@ trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] 
           _ <- call(client.emptyResponseDetailedErrorEndpoint, ("userId", "name1", 18)).failed
             .map(x => x.getMessage shouldBe "Unable to process your request")
         } yield succeed
-      }
-
-      if(timeoutTest) {
-        "timeout" in {
-          for {
-            _ <- call(client.slowResponseEndpoint, ()).failed.map(_ shouldBe a[scala.concurrent.TimeoutException])
-          } yield succeed
-        }
       }
 
       "properly handle joined headers" in {
--- a/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
+++ b/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
@@ -81,7 +81,7 @@ trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] 
         } yield succeed
       }
 
-      "timeout in" in {
+      "timeout" in {
         for {
           _ <- call(client.slowResponseEndpoint, ()).failed.map(_ shouldBe a [scala.concurrent.TimeoutException])
         } yield succeed
--- a/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
+++ b/algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
@@ -81,6 +81,12 @@ trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] 
         } yield succeed
       }
 
+      "timeout in" in {
+        for {
+          _ <- call(client.slowResponseEndpoint, ()).failed.map(_ shouldBe a [scala.concurrent.TimeoutException])
+        } yield succeed
+      }
+
       "properly handle joined headers" in {
         val response = UUID.fromString("29d15495-55ea-431e-bef3-392b05b14fef").toString
 
--- a/fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
+++ b/fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
@@ -2,7 +2,6 @@ package endpoints4s.fetch
 
 import endpoints4s.Hashing
 
-import scala.annotation.nowarn
 import scala.concurrent.duration.FiniteDuration
 
 /** Settings for XHR interpreter.
@@ -22,7 +21,6 @@ final class EndpointsSettings private (val baseUri: Option[String], val timeout:
 
   override def hashCode(): Int = Hashing.hash(baseUri, timeout)
 
-  @nowarn("cat=unused")
   private[this] def copy(
       baseUri: Option[String] = baseUri,
       timeout: Option[FiniteDuration] = timeout
--- a/fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
+++ b/fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
@@ -3,11 +3,12 @@ package endpoints4s.fetch
 import endpoints4s.Hashing
 
 import scala.annotation.nowarn
+import scala.concurrent.duration.FiniteDuration
 
 /** Settings for XHR interpreter.
   * @param baseUri Base of the URI of the service that implements the endpoints, can be absolute or relative (e.g. "http://foo.com" or "/bar")
   */
-final class EndpointsSettings private (val baseUri: Option[String]) extends Serializable {
+final class EndpointsSettings private (val baseUri: Option[String], val timeout: Option[FiniteDuration]) extends Serializable {
 
   override def toString =
     s"EndpointsSettings($baseUri)"
@@ -15,7 +16,7 @@ final class EndpointsSettings private (val baseUri: Option[String]) extends Seri
   override def equals(other: Any): Boolean =
     other match {
       case that: EndpointsSettings =>
-        baseUri == that.baseUri
+        baseUri == that.baseUri && timeout == that.timeout
       case _ => false
     }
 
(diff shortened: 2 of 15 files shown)


xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
```diff
@@ -3,11 +3,12 @@ package endpoints4s.xhr
 import endpoints4s.Hashing
 
 import scala.annotation.nowarn
+import scala.concurrent.duration.{FiniteDuration}
 
 /** Settings for XHR interpreter.
   * @param baseUri Base of the URI of the service that implements the endpoints, can be absolute or relative (e.g. "http://foo.com" or "/bar")
   */
-final class EndpointsSettings private (val baseUri: Option[String]) extends Serializable {
+final class EndpointsSettings private (val baseUri: Option[String], val timeout: Option[FiniteDuration]) extends Serializable {
```
julienrf, line 11: Could you please also update the implementation of `hashCode`, `equals`, and `toString` to take into account the `timeout`?

julienrf (commented):
Hey @romainreuillon, thank you for this contribution!

Would it make sense to add it to the `fetch` interpreter as well? Maybe @ptrdom would be interested to do it?

ptrdom:
@romainreuillon Thank you for also updating `fetch` interpreter. This change would also benefit from tests, I think it should be possible to set up a relevant interaction with `stub-server`.

fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
```diff
@@ -3,38 +3,45 @@ package endpoints4s.fetch
...
 
   override def equals(other: Any): Boolean =
     other match {
       case that: EndpointsSettings =>
-        baseUri == that.baseUri
+        baseUri == that.baseUri && timeout == that.timeout
       case _ => false
     }
 
   override def hashCode(): Int = Hashing.hash(baseUri)
```
julienrf, line 23: You need to add the timeout here as well

julienrf (commented):
Thank you for the update @romainreuillon!

As @ptrdom mentioned, would it be possible to add a test? You could create an endpoint in the stub server that would hang for some time (e.g. 2 sec) before returning a response. You could then check what happens with a longer or shorter timeout.

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
```diff
@@ -3,38 +3,45 @@ package endpoints4s.xhr
...
       case that: EndpointsSettings =>
-        baseUri == that.baseUri
+        baseUri == that.baseUri && timeout == that.timeout
       case _ => false
     }
 
-  override def hashCode(): Int = Hashing.hash(baseUri)
+  override def hashCode(): Int = Hashing.hash(baseUri, timeout)
 
   @nowarn("cat=unused")
```
julienrf, line 25: It seems this annotation is not necessary?

romainreuillon:
@julienrf Thanks for the guideline.

I'm gonna try to implement a few test. I am trying to understand how the test suite works and where the endpoints used by the tests are being implemented.

fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
romainreuillon: Good catch, I just added it

xhr/client/src/main/scala/endpoints4s/xhr/EndpointsSettings.scala
romainreuillon: The annotation was in the code already. I don't know, which warning it solves.

Should I remove it?

romainreuillon:
I implemented a unit test for the timeout. I implemented the timeout in http4s client, fetch and xhr. For these client the test passes.

I didn't find how to implement it in akka-client in a timely manner, could you give me a hand on this one?

Are there other clients that I have missed?

ptrdom:
Of course you can also implement timeouts for akka-http client, but then you would need to look into how to implement it, probably this page is relevant - https://doc.akka.io/docs/akka-http/current/common/timeouts.html.

ptrdom:
Also you can see in CI logs that `sttp-client` has also failed, so you need to implement timeouts for it too. Otherwise you could just create a new suite with its new endpoint for timeout testing and only have it implemented by specific interpreters you have implemented the timeouts for. Now your tests are in the main suite, which is implemented by all interpreters, even those you currently have not modified.

romainreuillon:
sttp-client timeout is now implemented and all test are okay on my computer

romainreuillon:
I am unsure on how the modification I made triggered this test errors. It seems related to some akka server config... Any idea?

ptrdom:
The errors I see for 2.13 workflow seem to be related to `@nowarn` annotations.

romainreuillon:
The @nowarn were in the code already. Should I remove them (as it seems unrelated to the timeout implem) ?

julienrf:
Hey @romainreuillon, thank you so much for addressing all the points we’ve made. Yes, it’s true that the `@nowarn` annotations were there before your changes. I don’t remember why we introduced them, to be honest. It seems that only now they create these compilation errors, so could you try to remove them?

romainreuillon:
I've just removed them

romainreuillon:
Anything left preventing to merge this PR?

algebras/testkit/src/main/scala/endpoints4s/algebra/client/EndpointsTestSuite.scala
```diff
@@ -13,7 +13,7 @@ trait ClientEndpointsTestApi extends EndpointsTestApi {
 
 trait EndpointsTestSuite[T <: ClientEndpointsTestApi] extends ClientTestBase[T] {
 
-  def clientTestSuite() = {
+  def clientTestSuite(timeoutTest: Boolean = true) = {
```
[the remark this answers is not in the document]
romainreuillon: The test is not supported by akka-http yet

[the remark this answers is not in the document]
romainreuillon: The difference between the ChunkedJsonEntities tests and the Client tests is that the Client tests are encapsulated in a method, whereas the ChunkedJsonEntities tests are implement directly within the trait.

Making this test modular using cake pattern would require to inject a new method and calling it in each test suite supporting the timeout settings. Would that be a solution or should we keep the boolean parameter until all clients implement the timeout setting?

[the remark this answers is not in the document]
julienrf: I don’t think we need the extra method in the trait (even that’s what in most of the existing test suites, I am not sure this provides value). I recommend doing something like this:

https://github.com/endpoints4s/endpoints4s/blob/3545ad7c118b73bf6f44cc0fefd1b08b2804953d/algebras/testkit/src/main/scala/endpoints4s/algebra/client/ChunkedEntitiesTestSuite.scala#L12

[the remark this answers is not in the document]
julienrf: I would prefer if you could move the test to a separate trait that you would mix only to the relevant interpreters (instead of using the boolean param). Would you mind making that change?

fetch/client/src/main/scala/endpoints4s/fetch/EndpointsSettings.scala
```diff
@@ -2,39 +2,44 @@ package endpoints4s.fetch
...
 import endpoints4s.Hashing
 
-import scala.annotation.nowarn
+import scala.concurrent.duration.FiniteDuration
 
 /** Settings for XHR interpreter.
   * @param baseUri Base of the URI of the service that implements the endpoints, can be absolute or relative (e.g. "http://foo.com" or "/bar")
   */
-final class EndpointsSettings private (val baseUri: Option[String]) extends Serializable {
+final class EndpointsSettings private (val baseUri: Option[String], val timeout: Option[FiniteDuration]) extends Serializable {
```
julienrf, line 10: To keep the backward binary compatibility, you need to add back the previous constructor as a secondary constructor:

```suggestion
final class EndpointsSettings private (val baseUri: Option[String], val timeout: Option[FiniteDuration]) extends Serializable {

  def this(baseUri: Option[String]) = this(baseUri, None)
```

http4s/client/js/src/test/scala/endpoints4s/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
```diff
@@ -48,7 +50,7 @@ class Http4sClientEndpointsJsonSchemaTest
       port = Some(stubServerPort)
     ),
     Uri.Scheme.http,
-    FetchClientBuilder[IO].create
+    FetchClientBuilder[IO].withRequestTimeout(FiniteDuration(2, TimeUnit.SECONDS)).create
```
julienrf, line 53: Why is this change necessary?

sttp/client/src/main/scala/endpoints4s/sttp/client/Endpoints.scala
```diff
@@ -37,7 +21,8 @@ import _root_.sttp.client3.{
   */
 class Endpoints[R[_]](
     val host: String,
-    val backend: SttpBackend[R, Any]
+    val backend: SttpBackend[R, Any],
+    val timeout: Option[FiniteDuration] = None
```
julienrf, line 25: You will probably have to re-introduce the old constructor signature as a secondary constructor here as well.

sttp/client/src/main/scala/endpoints4s/sttp/client/Endpoints.scala
```diff
@@ -332,11 +323,15 @@ trait EndpointsWithCustomErrors[R[_]]
       //#endpoint-type
       {
     def apply(a: A): R[B] = {
-      val result = backend.send(request(a).response(asStringAlways))
-      backend.responseMonad.flatMap(result) { sttpResponse =>
-        decodeResponse(response, sttpResponse)
+        val result = backend.send(request(a).response(asStringAlways))
+        val responseFuture = backend.responseMonad.handleError(result) {
+          case e if e != null && e.getCause.getClass.getName.toLowerCase.contains("timeout") =>
```
julienrf, line 328: Would it be possible to pattern match on the cause using a “typed pattern” instead of manipulating the class name? Something like:

~~~ scala
e.getCause match {
  case timeout: TimeoutException => ...
}
~~~

julienrf (changes requested):
Thank you again Romain for your patience! Overall, I am in favor of merging your contribution, but there are still a couple of points that I would like to see addressed.

sttp/client/src/main/scala/endpoints4s/sttp/client/Endpoints.scala
romainreuillon: It has been implemented

romainreuillon:
Dear @julienrf, 

thank you for having taken the time to review this PR. I addressed the issue you raised about this PR.

best,
Romain

http4s/client/js/src/test/scala/endpoints4s/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
julienrf: Oh, I see, there is no need to introduce the timeout configuration in the Http4s interpreter trait (like you did with the fetch interpreter, for instance), because the Http4s client can already be configured with a timeout by the users. That’s great!

I wonder if we should follow the same approach with the sttp interpreter then? Let the users configure the timeout in the underlying backend? Would that be possible?

romainreuillon: I extracted the config from the mixin in sttp tests, is that what you meant?

romainreuillon:
@julienrf I'm back from vacation. I saw that you requested one change concerning sttp tests. Is that all what is left before merging or did I miss some other points?

http4s/client/js/src/test/scala/endpoints4s/http4s/client/Http4sClientEndpointsJsonSchemaTest.scala
julienrf: Hey Romain ! No, I am sorry, I wasn't clear. What I meant is that maybe instead of introducing a timeout setting in the sttp interpreter it would be possible to do nothing about the timeout in endpoints4s and let sttp handle it itself. Just like you did here with this http4s interpreter: you didn't change anything, but you simply configured the timeout in http4s itself.

romainreuillon: Okay I understand. I think this was the first thing I tried.

I just tried it again, using this config:

```scala
val backend = TryHttpURLConnectionBackend(options = SttpBackendOptions.connectionTimeout(FiniteDuration(2, TimeUnit.SECONDS)))
```

but unfortunately, the interpreter doesn't timeout

romainreuillon: I think that the global config it made for the connection timeout (at tcp level) and the other for the read.

It is not the same timeout:
https://sttp.softwaremill.com/en/latest/conf/timeouts.html

romainreuillon: Okay, I think I've actually found how to it.

I just committed it.

julienrf (approved):

julienrf:
Thank you so much Romain for your work! I’ve pushed an additional commit to my fork to resolve the binary compatibility issues (hopefully). Do you mind fast-forward-merging it to your fork?

https://github.com/openmole/endpoints4s/compare/timeout...julienrf:endpoints:timeout

julienrf:
I just fixed the code style, please merge my last commit again: https://github.com/openmole/endpoints4s/compare/timeout...julienrf:endpoints:timeout

julienrf:
Thank you Romain, this has been released in 1.8.0

Issue #1096, opened 2022-11-12
Use `bgRun` for stub-server in CI


ptrdom:
Worth it to test out if it works fine, would save some CI time having to fiddle with Docker images.

One issue I noticed is that `bgRun` requires sbt to be kept running, so you cannot just run `sbt stub-server/bgRun` because it will just terminate upon return to the shell, but it seems to work fine with the native thin client - `sbtn stub-server/bgRun`.

ptrdom:
It will probably won't work - `sbtn` seems to unpredictably terminate background jobs.

julienrf:
> It will probably won't work - `sbtn` seems to unpredictably terminate background jobs.

Is that the expected behavior? Or is it a corresponding open issue on sbt?

ptrdom:
> Is that the expected behavior? Or is it a corresponding open issue on sbt?

Have not found anything about it, but then again, `sbtn` and `bgRun` are not too documented. Might be worth it to inquire in discord. But this also got me thinking that we probably do not even need that.

While this won't work:
```
- run: sbtn server/bgRun
- run: sbt ++2.12.13 test
```
This should work:
```
- run: sbt server/bgRun ++2.12.13 test
```
We do not really need to split the calls anymore 😅

julienrf:
Note that `sbt-revolver` is added to the build, so we could also do something like `sbt stub-server/reStart ++2.12 test` in case there is an issue with `bgRun`.
That being said, I do agree that `bgRun` would be worth investigating.

PR #1098, opened 2022-11-14
Use `bgRun` for stub-server in CI

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/1096.


ptrdom:
@julienrf Looking good, I think we can merge it.

julienrf (approved):

Issue #566, opened 2020-05-12
Scala 3 support


julienrf:
#559 shows that the project compiles with Dotty.

The next steps are:

- [x] compile all the modules with Scala 3
- [ ] run the tests with Dotty (requires scalatest to be published for recent Dotty versions)
- [ ] publish artifacts for Dotty (requires adding an sbt command to the `.travis.yml`)
- [x] migrate the `endpoints-json-schema-generic` module (should we use Shapeless 3? Magnolia? A vanilla Scala 3 solution?)

stefanholzmueller:
According to https://github.com/scalatest/scalatest/issues/1865, scalatest 3.2.1 is published for Dotty 0.26.0-RC1 now. I quickly tried ` sbt "++ 0.25.0-RC2 -v compile"`, but it seems that circe does not support newer Dotty versions yet.

When I tried to run the tests, I even got some NullPointerExceptions from Dotty. I won't dig deeper today, I was just curious and wanted to share my results.

harpocrates:
Dotty issue I opened for a NPE in json-schema's testsuite: https://github.com/lampepfl/dotty/issues/9572.

MasseGuillaume:
@julienrf Coucou, what can I do to make this happen?

GrafBlutwurst:
I would also be willing to chip into this effort. Lemme know if i can help somehow.

julienrf:
Thank you for your messages!

All the code compiles with Scala 3, but the real blocker is the state of the module `json-schema-generic`. That module contains implementation details that need to be made `public` because of a limitation of Scala 3 (IIUC, see lampepfl/dotty#11706):

https://github.com/endpoints4s/endpoints4s/blob/6ab7db825c4e567e77d3a5abf669ad7829df9050/json-schema/json-schema-generic/src/main/scala-3/endpoints4s/generic/JsonSchemas.scala#L122

Furthermore, the current implementation of the macros produces code that gives warnings at compile-time: https://github.com/endpoints4s/endpoints4s/runs/5138703907?check_suite_focus=true#step:8:51

Based on the Sonatype stats, I believe the module `json-schema-generic` is used a lot, so I think it’s better to wait for this module to be ready before we publish Scala 3 artifacts (alternatively, we could publish the Scala 3 artifacts for all the other modules except `json-schema-generic`, but I don’t think it would be helpful since many people use `json-schema-generic`).

[...]

PR #1099, opened 2022-11-14
Cross-test with Scala 3

julienrf:
Remaining issues

- [x] Fix compilation of http4s-server tests. There seems to be an issue in the usage of the Akka-http graph DSL in Scala 3. We use it to convert between Akka-streams and http4s streams.
- [x] Compilation of openapi module loops indefinitely

Relates to #566

### DIFF
--- a/akka-http/build.sbt
+++ b/akka-http/build.sbt
@@ -55,7 +55,6 @@ val `akka-http-server` =
       excludeDependencies ++= {
         if (scalaBinaryVersion.value.startsWith("3")) {
           List(
-            ExclusionRule("org.scala-lang.modules", "scala-xml_3"),
             ExclusionRule("org.scala-lang.modules", "scala-collection-compat_2.13")
           )
         } else Nil
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       with:
         java-version: adopt@1.11
     - run: npm install jsdom@12.2.0
-    - run: sbt stub-server/bgRun ++2.12.13 test
+    - run: sbt stub-server/bgRun "++ 2.12 test"
 
   test213:
     name: Test (Scala 2.13)
@@ -33,7 +33,7 @@ jobs:
         java-version: adopt@1.11
     - run: npm install jsdom@12.2.0
     - run: sudo apt-get install graphviz
-    - run: sbt stub-server/bgRun ++2.13.8 coverage test coverageReport coverageAggregate manual/makeSite
+    - run: sbt stub-server/bgRun "++ 2.13 ;coverage ;test ;coverageReport ;coverageAggregate ;manual/makeSite"
     - run: bash <(curl -s https://codecov.io/bash)
 
   test3:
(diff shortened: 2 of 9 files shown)


julienrf:
Any help to move forward would be appreciated, I don’t have time to look at the remaining issues.

ptrdom:
I do not have permissions to push to your branch, so here is commit to my fork that fixes http4s-server test compilation - https://github.com/ptrdom/endpoints4s/commit/d3e4f09a9d7785965d65233eff90b00d68c7438e.

ptrdom:
And this fixes openapi tests - https://github.com/ptrdom/endpoints4s/commit/f3b2b4d9829fe4960e0698a756218186fe640905.

ptrdom:
Remaining Scala 3 errors are all in example code, at a quick glance they do look like issues in build definition.

ptrdom:
You will probably notice, but just a note that you might not want to take the following change from my commits -
https://github.com/ptrdom/endpoints4s/commit/d3e4f09a9d7785965d65233eff90b00d68c7438e#diff-993fc08217429dd4f90458a62bd3f0c461dd5c8c8031173e323d6fa94e95f21bR67-R70.

julienrf:
Thank you very much @ptrdom!

> I do not have permissions to push to your branch, so here is commit to my fork that fixes http4s-server test compilation - [ptrdom@d3e4f09](https://github.com/ptrdom/endpoints4s/commit/d3e4f09a9d7785965d65233eff90b00d68c7438e).

IIUC we only need to bump akka and use `createGraph` instead of `create`? Nice!

> And this fixes openapi tests - [ptrdom@f3b2b4d](https://github.com/ptrdom/endpoints4s/commit/f3b2b4d9829fe4960e0698a756218186fe640905).

Interesting! How did you figure out the cause of the issue? I wonder if there is a bug in Scala 3 or if this is expected.

ptrdom:
> IIUC we only need to bump akka and use createGraph instead of create? Nice!

Yeah, I noticed a comment about it in the code of the first Akka release for Scala 3 - figured it was very likely they ran into this issue too.

> Interesting! How did you figure out the cause of the issue? I wonder if there is a bug in Scala 3 or if this is expected.

I just noticed that the class in error had private implicits in inner class, changed that - did not seem to help, then I moved the inner class definitions out to the companion object and that did it.

I see we have more errors now, will check them out.

julienrf:
Ah, about f3b2b4d, I think this is related to a limitation that has been fixed in lampepfl/dotty#15847, which has been shipped in Scala 3.2.0. We could also try to bump the Scala 3 version to 3.2.1, maybe?

ptrdom:
> Ah, about [f3b2b4d](https://github.com/endpoints4s/endpoints4s/commit/f3b2b4d9829fe4960e0698a756218186fe640905), I think this is related to a limitation that has been fixed in [lampepfl/dotty#15847](https://github.com/lampepfl/dotty/pull/15847), which has been shipped in Scala 3.2.0. We could also try to bump the Scala 3 version to 3.2.1, maybe?

Neat, will try it out!

ptrdom:
~We could also try to bump the Scala 3 version to 3.2.1, maybe?~ (actually, I am fine with your change)

All good. That change does improve the code quality, so to speak - those definitions do not benefit from being inside the class, so might as well be in the companion object.

akka-http/build.sbt
```diff
@@ -55,7 +55,6 @@ val `akka-http-server` =
       excludeDependencies ++= {
         if (scalaBinaryVersion.value.startsWith("3")) {
           List(
-            ExclusionRule("org.scala-lang.modules", "scala-xml_3"),
```
ptrdom, line 58: @julienrf Is this change fine? Tests will not work on Scala 3 with this exclusion.

ptrdom (changes requested):

ptrdom (commented):
Checks pass, just need to double check that exclusion change.

akka-http/build.sbt
julienrf: Yes, that’s even better! Eventually, we should get rid of those exclusions.

julienrf: That makes me think that we should probably not publish Scala 3 artifacts for the modules that still use `CrossVersion.for3Use2_13` in their dependencies… :thinking:

julienrf:
That’s amazing, thank you @ptrdom!

Issue #1103, opened 2022-11-20
Akka license changes


ptrdom:
With Akka's move to BSL license and its ASF fork getting traction (https://github.com/apache/incubator-pekko), I think we should consider doing one of the following upon Pekko's release:
- Change all Akka-based code to use Pekko.
- Create Pekko-based alternatives for all Akka-based code.

pjfanning:
Pekko HTTP 1.0.0 should be released this week.

https://pekko.apache.org/docs/pekko-http/current/

julienrf:
I see in the stats that the Akka HTTP interpreter is still used by ~15% of the users of endpoints4s.

For the sake of simplicity, I would like to avoid having both Akka HTTP and Pekko HTTP in this repository (endpoints4s/endpoints4s). It would be better to move the existing Akka HTTP interpreter to its own repository (endpoints4s/akka-http) and have the Pekko HTTP interpreter here. Is there anyone interested in this task?

ptrdom:
I was planning to pick this up eventually, and I think I should be able to get started even this week. Moving Akka HTTP to a separate repository and replacing it in this repository with Pekko certainly simplifies things. In regards to the new repository for Akka HTTP interpreter, I assume that because of permissions and secrets required for the initial repository setup, @julienrf you are the only person that can do that, am I correct?

ptrdom:
@julienrf I pushed the initial split to https://github.com/endpoints4s/akka-http, but the branch was called `split` and it got set as default branch. Could you set `main` as default? I do not have the permissions to do that.

julienrf:
Thank you very much @ptrdom! I changed the default branch to `main` and also gave you admin access to the repo in case this is needed in the future…

PR #1104, opened 2022-11-22
Remove no longer necessary uses of CrossVersion.for3Use2_13

ptrdom:
References https://github.com/endpoints4s/endpoints4s/issues/566.

### DIFF
--- a/akka-http/build.sbt
+++ b/akka-http/build.sbt
@@ -3,10 +3,7 @@ import EndpointsSettings._
 val `algebra-jvm` = LocalProject("algebraJVM")
 val `algebra-testkit-jvm` = LocalProject("algebra-testkitJVM")
 val `algebra-circe-testkit-jvm` = LocalProject("algebra-circe-testkitJVM")
-val `algebra-playjson-jvm` = LocalProject("algebra-playjsonJVM")
-val `json-schema-circe-jvm` = LocalProject("json-schema-circeJVM")
 val `json-schema-generic-jvm` = LocalProject("json-schema-genericJVM")
-val `json-schema-playjson-jvm` = LocalProject("json-schema-playjsonJVM")
 val `openapi-jvm` = LocalProject("openapiJVM")
 
 val `akka-http-client` =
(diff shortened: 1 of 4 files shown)


ptrdom:
This PR is not that useful:

- `play-json` is still pending a release for Scala 3. Seems like it could be expected relatively soon
- `akka-http` is also pending a release for Scala 3. It also could be expected relatively soon, but it will be only under new BSL license. Relates to https://github.com/endpoints4s/endpoints4s/issues/1103.

julienrf (approved):
Thank you very much!

I guess we will (unfortunately) have to wait before we can publish the akka-http modules for Scala 3, right?

ptrdom:
> I guess we will (unfortunately) have to wait before we can publish the akka-http modules for Scala 3, right?

Yes, indeed.

julienrf:
I’ve pushed a commit to not publish the Scala 3 modules that still depend on Scala 2.13 artifacts.

I’ve seen there are some errors in the CI, do you know what’s going on?

ptrdom:
I saw some errors in Akka HTTP interpreters related to removal of `CrossVersion.for3Use2_13` for actors and streams, have not looked into the errors further.

julienrf:
Unfortunately, bumping the Scala 3 compiler to 3.2.x does not seem to be easy. I’ve pushed another commit that fixes the CI but I realize that it basically undoes what you previously did: it adds back all the removed `cross(CrossVersion.for3Use2_13)`.

At least that allows us to publish Scala 3 modules for the algebras, the http4s interpreters, and the Scala.js interpreters…

ptrdom:
Yeah, the name of this PR has kind of lost its meaning now 😄 Oh well, we will have a chance to get back to this topic in the future.

julienrf:
What would you like to do? I am fine with waiting a bit longer in case you would be interested in investigating how to use the Scala 3 artifacts of Akka… Otherwise, I am happy to roll a 1.9.0 release of endpoints4s with Scala 3 artifacts for the algebras, the sttp interpreter, the http4s interpreters, and the Scala.js interpreters.

ptrdom:
We can definitely just go forward with 1.9.0 release now, there is not much point in trying to resolve Akka issues at this moment because we are anyways blocked by Akka HTTP.

PR #1136, opened 2023-02-08
Add redirect status codes

laurin-no:
Add status codes for:
- [307 Temporary Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307)
- [308 Permanent Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308)


julienrf (approved):
Thank you for your contribution @laurin-no!

julienrf:
You need to [relax the compatibility intention](https://github.com/endpoints4s/endpoints4s/blob/master/CONTRIBUTING.md#introduction-of-source-or-binary-incompatibilities) to `Compatibility.BinaryCompatible` in the main [build.sbt](https://github.com/endpoints4s/endpoints4s/blob/8137c72922371938bd1dd880e8ec7e5459296fb7/build.sbt#L45) (which is used by default by all the modules unless they override it)

laurin-no:
thanks for the feedback @julienrf, i updated the pr

PR #1162, opened 2023-05-15
Add Server and ServerVariable objects to the OpenAPI model

### LINKED ISSUE #1139: OpenApi: Add "Servers" top-level field to OpenAPI model and renderer
The openAPI model would benefit from the addition of ["Server Object"](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#server-object)

As a starting point, here's a case class representation of Server:
```scala
/** see <https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#server-object> */
case class Server(
  url: String,
  description: Option[String] = None,
  variables: Map[String, ServerVariable] = Map.empty
)

/** see <https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#server-variable-object> */
case class ServerVariable(
  `enum`: Option[Seq[String]] = None,
  default: String,
  description: Option[String] = None
)
```

Some alternative modelling choices could be made, but I went for "as direct a mapping as possible".

Here's some starter code for rendering the model in OpenApiRenderer:

```scala
  private def serverVariableJson(variable: ServerVariable): ujson.Obj = variable match {
    case ServerVariable(enumAlternatives, default, description) =>
      val fields: mutable.LinkedHashMap[String, ujson.Value] = mutable.LinkedHashMap(
        "default" -> ujson.Str(default)
      )
      description.foreach(desc => fields += "description" -> ujson.Str(desc))
      enumAlternatives.foreach(alternates => fields += "enum" -> ujson.Arr.from(alternates.map(ujson.Str)))
      ujson.Obj(fields)
  }
  private def serversJson(servers: Seq[Server]): ujson.Arr =
    ujson.Arr.from(servers.map { case Server(url, description, variables) =>
      val fields: mutable.LinkedHashMap[String, ujson.Value] = mutable.LinkedHashMap(
        "url" -> ujson.Str(url)
      )
      description.foreach(desc => fields += "description" -> ujson.Str(desc))
      if (variables.nonEmpty) {
        val variablesFields: mutable.LinkedHashMap[String, ujson.Value] = mutable.LinkedHashMap()
        variables.foreach { case (k, v) => variablesFields += k -> serverVariableJson(v) }
        fields += "variables" -> ujson.Obj(variablesFields)
      }
      
      ujson.Obj(fields)
    })
```

Due to time limitations, I am unable to write the remainder of the code, especially in light of the compatibility constraints preventing the use of case classes.

julienrf:
Fixes #1139

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -58,3 +58,9 @@ releaseProcess := Seq[ReleaseStep](
   tagRelease,
   pushChanges
 )
+
+import com.typesafe.tools.mima.core._
+ThisBuild / mimaBinaryIssueFilters ++= Seq(
+  // OK, constructor is private (but accessed from within the companion)
+  ProblemFilters.exclude[DirectMissingMethodProblem]("endpoints4s.openapi.model.OpenApi.this")
+)
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -11,31 +11,33 @@ import scala.collection.mutable
 final class OpenApi private (
     val info: Info,
     private val innerPaths: collection.Map[String, PathItem],
-    val components: Components
+    val components: Components,
+    val servers: Seq[Server]
 ) extends Serializable {
 
   def paths: Map[String, PathItem] = innerPaths.toMap
 
   override def toString =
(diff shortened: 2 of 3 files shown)


emanb29:
This looks perfect and addresses my intent in #1139 -- thanks for finding the time to flesh it out!

PR #1194, opened 2023-08-28
Introduce a rawField JSON schema

ant8e:
A field JSON Schema allowing fine control of the absence or presence (with a value, that can be null) of a JSON property.

Use case: Be able to implement endpoints conforming to [rfc7386 JSON Merge Patch](https://www.rfc-editor.org/rfc/rfc7386)  where it is mandatory to distinguish between a null and an absent property.

### DIFF
--- a/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
+++ b/json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
@@ -364,7 +364,8 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
     */
   def preciseField[A](name: String, documentation: Option[String] = None)(implicit
       tpe: JsonSchema[A]
-  ): Record[JsonSchemas.PreciseField[A]]
+  ): Record[JsonSchemas.PreciseField[A]] =
+    sys.error("Unsupported algebra version: 1.11.0. Please update your interpreter dependency.")
 
   /** The JSON schema of a record with a single optional field with the given `name`
     *
--- a/json-schema/json-schema-circe/src/main/scala/endpoints4s/circe/JsonSchemas.scala
+++ b/json-schema/json-schema-circe/src/main/scala/endpoints4s/circe/JsonSchemas.scala
@@ -242,7 +242,7 @@ trait JsonSchemas extends algebra.NoDocsJsonSchemas with TuplesSchemas {
       )
     )
 
-  def preciseField[A](name: String, documentation: Option[String] = None)(implicit
+  override def preciseField[A](name: String, documentation: Option[String] = None)(implicit
       tpe: JsonSchema[A]
   ): Record[PreciseField[A]] =
     Record(
--- a/json-schema/json-schema-generic/src/test/scala/endpoints4s/generic/JsonSchemasTest.scala
+++ b/json-schema/json-schema-generic/src/test/scala/endpoints4s/generic/JsonSchemasTest.scala
@@ -123,7 +123,7 @@ class JsonSchemasTest extends AnyFreeSpec {
     ): String =
       s"$name:$tpe?${docs.fold("")(doc => s"{$doc}")}"
 
-    def preciseField[A](name: String, docs: Option[String])(implicit
+   override def preciseField[A](name: String, docs: Option[String])(implicit
         tpe: String
     ): String =
       s"$name:$tpe?${docs.fold("")(doc => s"{$doc}")}"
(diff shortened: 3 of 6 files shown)


julienrf:
Thank you Antoine for your pull request.

 I won't be able to review it before a couple of weeks. If anyone else wants to review it in the meantime, feel free to do so.

julienrf (commented):
I had a quick look and I think this is the right approach.
Do you mind showing what the final code of your use case looks like? Maybe add it as a test case if that's easy. It is just to see if we should add more methods to the type RawField.

julienrf:
Hey @ant8e, are you still interested in working on this?

ant8e:
Yes. I have something drafted. Need to find the time to finish

julienrf:
Great, thank you! As alternative names for “raw”, I thought of “precise”. Maybe `PreciseJson`?

ant8e:
I want to add also an example of implementing a JSON Merge patch, but I'm not sure where to put it
Any suggestion ?

julienrf:
> I want to add also an example of implementing a JSON Merge patch, but I'm not sure where to put it Any suggestion ?

Great! This file is the best place, IMHO:

https://github.com/endpoints4s/endpoints4s/blob/master/json-schema/testkit/src/main/scala/endpoints4s/algebra/JsonSchemasOptionalFieldsTest.scala

The testing infrastructure is a bit messy, but for now the simplest path is to add your test to the end of that file. How does that sound?

json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
```diff
@@ -349,6 +349,23 @@ trait JsonSchemas extends TuplesSchemas with PartialInvariantFunctorSyntax {
...
+    *     it is present but has the value `null`. They fail if the field is
+    *     present but contains an invalid value,
+    *   - Documentation interpreters produce the JSON schema of a JSON object with an
+    *     optional property of the given `name`.
+    *
+    * @group operations
+    */
+  def preciseField[A](name: String, documentation: Option[String] = None)(implicit
+      tpe: JsonSchema[A]
+  ): Record[JsonSchemas.PreciseField[A]]
```
julienrf, line 367: To “fix” the CI failure, we need to implement this method. Indeed, leaving it abstract in binary incompatible change (existing code that mixes-in the trait `JsonSchemas` won’t have an implementation of the method). In endpoints4s, we decided to handle such situations by providing a default implementation that throws an error at run-time. This is not a problem because existing code couldn’t call this method, so there is not much risk for them to get a run-time exception. The only issue that can happen is if someone wants to upgrade to the new version of the algebra without upgrading the interpreters: in that case, they will get the run-time exception. Therefore, the message in the exception should tell the users to upgrade their interpreters:

~~~ scala
sys.error("Unsupported algebra version: 1.11.0. Please update your interpreter dependency.")
~~~

(Note that we do have a method specifically for that purpose in the `algebra` module, but we can not use it from the `json-schema` module…)

julienrf:
> I want to add also an example of implementing a JSON Merge patch

After reading your comment again, maybe my suggestion was not the best one. Could you please elaborate on the content of your test?

json-schema/json-schema/src/main/scala/endpoints4s/algebra/JsonSchemas.scala
ant8e: This is a nice trick ! It will be quite useful to do this in my $WORK code base.

julienrf (approved):
Thank you @ant8e for your contribution! Do you still want to add the test case showing a JSON merge patch?

ant8e:
What I have in mind is to code some example that does implement a patch endpoint, conforming to the JSON Merge patch RFC.  `preciseField`  will allow doing that in the presence of optional fields.  That could also be just documentation.
Anyway, I can do this in a subsequent PR, if you want to go forward with merging this one.

julienrf:
It would be cool to have that example!

You can add a `JsonMergeTestApi` trait [here](https://github.com/endpoints4s/endpoints4s/tree/master/algebras/testkit/src/main/scala/endpoints4s/algebra) defining the endpoint that uses `preciseField`, and then write a [client test suite](https://github.com/endpoints4s/endpoints4s/tree/master/algebras/testkit/src/main/scala/endpoints4s/algebra/client) and a [server test suite](https://github.com/endpoints4s/endpoints4s/tree/master/algebras/testkit/src/main/scala/endpoints4s/algebra/server) that both use that endpoint. Finally, update the interpreters (at least pekko-http and http4s, if possible also the fetch and sttp interpreters) to mix-in your test suites.

julienrf:
@ant8e I propose that we go ahead and merge your work now. If you have time and interest, you are welcome to add the JSON merge patch example in a separate PR!

PR #1204, opened 2023-10-18
Make OpenApi.jsonEncoder public

danarmak:
To serialize the json using a non-default ujson renderer, e.g. to produce indented output, I need access to the json structure. Right now the only way to do this is to re-parse the output of `stringEncoder`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ ThisBuild / version := (LocalProject("algebraJVM") / version).value
 // We want to keep binary compatibility as long as we can for the algebra,
 // but it is OK to publish breaking releases of interpreters. So,
 // interpreter modules may override this setting.
-ThisBuild / versionPolicyIntention := Compatibility.BinaryAndSourceCompatible
+ThisBuild / versionPolicyIntention := Compatibility.BinaryCompatible
 // Ignore dependencies to modules with version like `1.2.3+n`
 ThisBuild / versionPolicyIgnoredInternalDependencyVersions := Some("^\\d+\\.\\d+\\.\\d+\\+n".r)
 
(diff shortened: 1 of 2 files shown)


julienrf:
The _source_ compatibility is affected here, not the binary compatibility. There are some messages about binary compatibility in the logs because we check the forward binary compatibility as a proxy for the source compatibility, but the relevant message is this one:

https://github.com/endpoints4s/endpoints4s/actions/runs/6562714377/job/17825354883?pr=1204#step:4:182

This can be addressed by relaxing the compatibility intention, but I would like to learn more about the motivation before doing that change, could you please elaborate on that?

danarmak:
Right now the only way to export an `OpenApi` to json that can be used by some other tool is by using `OpenApi.stringEncoder: Encoder[OpenApi, String]`. This produces a compact json string without any whitespace, but I want nicely indented json. Right now I'm re-parsing the string into json, and then writing it again with `ujson.StringRenderer(indent=4)`; this seems silly when we can just expose the original `ujson.Value`.

And although I don't need to do this right now, I can also imagine wanting to use some other `ujson.Visitor`, e.g. for writing Yaml instead of json, or embedding the json value in another, bigger value.

julienrf:
OK, I understand. Then you need to relax the compatibility intention [here](https://github.com/endpoints4s/endpoints4s/blob/1a4d9627d6df9a0e3b39f6cf778836361a0d0014/build.sbt#L44) to `BinaryCompatible`.

danarmak:
Why not relax the compatibility intent only for the openapi sub-project? Or do you mean I should do that, but in /build.sbt and not in /openapi/build.sbt?

julienrf:
This is a good question. We could indeed do it only in the openapi project. We would also have to do the change in the dependent projects, though (all the interpreters). I think we will anyway have to do it at the algebra level in #1194, that’s why I suggest we do it now.

danarmak:
Alright, I replaced the last commit with one making the change globally.

julienrf (approved):

julienrf:
Thank you!

PR #1205, opened 2023-10-18
OpenApi for `application/octet-stream` should have type=string, format=binary

danarmak:
[OAPI reference](https://swagger.io/docs/specification/data-models/data-types/#file) specifies this for 'files' i.e. opaque binary request/response bodies.

The previous value of `MediaType(None)` means anything is allowed, which is technically correct; but Swagger Codegen refuses to parse a schema that doesn't have a `type` specified, and if I specify `type="string"`, then `format="binary"` is also required for the schema to mean the right thing. (This might have changed in the latest swagger-codegen release.)

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/ChunkedEntities.scala
@@ -19,7 +19,7 @@ trait ChunkedEntities extends algebra.ChunkedEntities with EndpointsWithCustomEr
   def textChunksResponse: ResponseEntity[Chunks[String]] = textChunksEntity
 
   private lazy val bytesChunksEntity =
-    Map("application/octet-stream" -> MediaType(None))
+    Map("application/octet-stream" -> MediaType(Some(Schema.simpleString.withFormat(Some("binary")))))
 
   def bytesChunksRequest: RequestEntity[Chunks[Array[Byte]]] = bytesChunksEntity
 


julienrf (approved):
I approve this change if it increases the compatibility with the tooling that relies on OpenAPI files.

Did you try with the latest version of the Swagger code generator ?

Also, I’ve noticed that in OpenAPI 3.1 the specification has changed and we may not use `format` anymore: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#considerations-for-file-uploads. I would be curious to know what happens if you change the [version](https://github.com/endpoints4s/endpoints4s/blob/39f1b8af3edb9e77909be17fb8495083df9158c2/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala#L57) of the OpenAPI document we produce to 3.1.0.

Do we know in the wild the level of support of OpenAPI 3.1? Would it be a good idea to switch to that version?

danarmak:
I can't reproduce the error with the latest (online) version of swagger-codegen. So this was probably fixed at some point. I have an older 5.x version in my CI flow (and I'll update it now).

The change to the schema is still good, according to OAS 3.0.0, but it would make it backward-incompatible with OAS 3.1.0 when you do switch - I hadn't realized that! So perhaps it's better not to merge this, and keep the schema slightly non-explicit but more forward compatible.

Do you want me to push a commit fixing the tests here? Or close the PR?

julienrf:
> So perhaps it's better not to merge this, and keep the schema slightly non-explicit but more forward compatible.

OK, let’s do that.

Thank you for the investigation!

PR #1210, opened 2023-11-21
Publish Algebra and Pekko interpreters for Scala 3

ptrdom:


ptrdom:
@julienrf Could you check the errors?

`faithful-cats` issue can be fixed be either:

1. `faithful-cats` Scala 3 release. If you have time to work on it then we can wait.
1. Dropping `faithful-cats` in the example code.

`for3Use2_13` will not work here, I assume.

As for CQRS query example I am unsure how to make that code work on Scala 3.

julienrf:
Using the 2.13 binaries should work, but we might end up with two binaries of cats on the class path, which will fail the build unless we add an exclusion rule. If it gets more complicated than that, I am fine with dropping faithful-cats.

ptrdom:
> Using the 2.13 binaries should work, but we might end up with two binaries of cats on the class path, which will fail the build unless we add an exclusion rule. If it gets more complicated than that, I am fine with dropping faithful-cats.

This was not enough - now there are errors in code where `faithful-cats` are used. How to we proceed?

julienrf:
I think we should update the example to use the `fetch-client` interpreter instead, and return a standard `Future`. Does that make sense?

ptrdom:
> I think we should update the example to use the `fetch-client` interpreter instead, and return a standard `Future`. Does that make sense?

Sounds good!

Just a remimder if you could check that Scala 3 error on CQRS query example - it seems very basic, but I have no clue why Scala 3 does not like those mux endpoints.

julienrf:
Ah sorry I missed that one. That could be because of differences in type inference between Scala 2 and Scala 3. But I think we do have examples of multiplexed endpoints that compile with Scala 3 somewhere else in the codebase, no? If not, it is possible that they are not easy to support on Scala 3, and to be honest I would suggest not using them in the example (so, it means defining multiple endpoints), or simply not enabling Scala 3 on that specific example (I think updating those examples is a bit out of the scope of this PR, but if you are motivated to clean up the legacy, you are more than welcome, of course!).

ptrdom:
Ah, I kind of assumed that I _have_ to deal with examples too, but I guess that those exclusion rules should help with that.

julienrf (approved):
Sounds good to me!

ptrdom:
@julienrf This is ready for review, I totally misunderstood the errors I was getting - I did not need to touch the example projects at all 🤦

ptrdom:
And you just approved it - thank you!

project/EndpointsSettings.scala
```diff
@@ -66,7 +66,7 @@ object EndpointsSettings {
   )
   val `scala 2.12 to dotty` = Seq(
     scalaVersion := "2.13.10",
-    crossScalaVersions := Seq("2.13.10", "3.1.3", "2.12.13")
+    crossScalaVersions := Seq("2.13.10", "3.3.0", "2.12.13")
```
julienrf, line 69: You had to update the Scala 3 version because Pekko HTTP requires at least Scala 3.3.x, right?

julienrf:
> I think we should update the example to use the `fetch-client` interpreter instead, and return a standard `Future`. Does that make sense?

FYI, I decided to do that in https://github.com/endpoints4s/endpoints4s/pull/1214. I also copied the content of the `xhr` interpreters in a separate repo (https://github.com/endpoints4s/xhr) like you did with Pekko HTTP. After that, we should be able to safely delete the `xhr` interpreters implementation from the core repository.

project/EndpointsSettings.scala
ptrdom: Yep, that was the reason.

OlegYch:
was this published? i'm trying to update play module, and the only things missing are `algebra-testkit` and `algebra-circe-testkit`

julienrf:
Hey Oleg! Not yet, we will cut a release next week.

ptrdom:
> was this published? i'm trying to update play module, and the only things missing are `algebra-testkit` and `algebra-circe-testkit`

Awesome initiative @OlegYch! Are you going to update to Play 3.0 or 2.9? I guess that 3.0 probably makes more sense for endpoints4s going forward.

OlegYch:
@ptrdom play3, here is draft pr https://github.com/endpoints4s/play/pull/5

PR #1222, opened 2023-12-05
Test that basic auth endpoints do not prevent unprotected endpoints from matching

OlegYch:


algebras/testkit/src/main/scala/endpoints4s/algebra/server/BasicAuthenticationTestSuite.scala
```diff
@@ -62,6 +62,23 @@ trait BasicAuthenticationTestSuite[T <: BasicAuthenticationTestApi] extends Endp
       }
     }
 
+    "fall through to another endpoint if url doesn't match" in {
```
julienrf, line 65: In the case you provided, the URL _does_ match, I believe the right behavior is to reject the request with an `Unauthorized` response.

OlegYch: @julienrf does it make sense?

julienrf: Aah OK, I see now that there are several endpoints starting with `/users`: `protectedEndpoint`, `unprotectedEndpoint`, and `protectedEndpointWithParameter`. I thought you were using both `protectedEndpoint` and `unprotectedEndpoint` in the tests, but now I see that you use `unprotectedEndpoint` and `protectedEndpointWithParameter`.

In that case, indeed, the URL `/users` does not match the `protectedEndpointWithParameter` but still matches the `unprotectedEndpoint`, so the response should indeed be `"Unprotected"`! :+1:

julienrf (approved):

julienrf:
The compatibility guarantees need to be relaxed in the following projects (uncomment these lines):

https://github.com/endpoints4s/endpoints4s/blob/b57330c62f1ed7c9c16b88a6b18716a2edab8742/algebras/build.sbt#L29
https://github.com/endpoints4s/endpoints4s/blob/b57330c62f1ed7c9c16b88a6b18716a2edab8742/algebras/build.sbt#L77

julienrf:
Thank you @OlegYch, I will trigger a new release soon.

PR #1234, opened 2024-02-12
Update ujson to `3.2.0`

lolgab:
The version of `ujson` currently used is ancient and creates conflicts when endpoints4s is used together with other libraries depending on newer `upickle` or `geny`.
Moreover ujson `3+` fixes a DOS vulnerability (more info here https://github.com/com-lihaoyi/upickle/pull/449)

### DIFF
--- a/algebras/build.sbt
+++ b/algebras/build.sbt
@@ -26,7 +26,7 @@ val `algebra-testkit` =
       publishSettings,
       `scala 2.12 to dotty`,
       name := "algebra-testkit",
-//      versionPolicyIntention := Compatibility.None,
+      versionPolicyIntention := Compatibility.None,
       libraryDependencies ++= Seq(
         "org.apache.pekko" %% "pekko-http" % pekkoHttpVersion,
         "org.apache.pekko" %% "pekko-actor" % pekkoActorVersion,
@@ -43,9 +43,9 @@ val `algebra-testkit` =
     .dependsOnLocalCrossProjectsWithNative("json-schema-testkit")
     .configurePlatforms(JSPlatform, NativePlatform)(_.disablePlugins(ScoverageSbtPlugin))
 
-val `algebra-testkit-js` = algebra.js
-val `algebra-testkit-jvm` = algebra.jvm
-val `algebra-testkit-native` = algebra.native
+val `algebra-testkit-js` = `algebra-testkit`.js
+val `algebra-testkit-jvm` = `algebra-testkit`.jvm
+val `algebra-testkit-native` = `algebra-testkit`.native
 
 val `algebra-circe` =
   crossProject(JSPlatform, JVMPlatform)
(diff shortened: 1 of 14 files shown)


julienrf:
Hey @lolgab, thank you for opening the PR. This is a breaking change because we bump the major version of upickle. As a consequence, we will have to bump the major version of all the interpreters that depend the `openapi` module (ie, the server interpreters pekko-http and http4s).

Before we go further, do you think the API of upickle is stable enough to do the change? I would prefer not doing it again in a few months.

julienrf:
I meant the API stability, which you answered perfectly! Would it be wiser to wait for upickle 4 to make the bump?

lolgab:
About the API I used, it seemed to me that the existing code was written in a lower-level way for performance reasons.
So I changed `+= key -> value` into `.put(key, value)` which doesn't require the allocation of tuples.
Otherwise, the API is still available. If you prefer, I can revert to use `+=`.
About uPickle 4, consider that the blocking issue (https://github.com/com-lihaoyi/upickle/issues/528) has no PR yet and nobody is working on it. So it might take time.
It would be nice if endpoints4s would not depend on ujson on its core but this means redesigning the lib altogether.

julienrf:
> It would be nice if endpoints4s didn't depend on `ujson` on its core but this means redesigning the lib altogether.

The core of endpoints4s does not depend on `ujson`, only the server interpreters that produce OpenAPI documentation.

If you say that uPickle 4 may take some time to land, then let’s move forward with your PR! Do you need any help to fix the tests?
Regarding the `versionPolicyCheck` failure, what you need to do is to [relax the compatibility intention](https://github.com/endpoints4s/endpoints4s/blob/master/CONTRIBUTING.md#introduction-of-source-or-binary-incompatibilities) in the `openapi` module and all the modules that depend on it (off the top of my head: the Pekko HTTP server interpreter and the http4s server interpreter).

julienrf:
Thank you for updating the PR. Actually, I see that the Scala.js bump requires relaxing the [default](https://github.com/endpoints4s/endpoints4s/blob/bac47f5e355a9c38e077bf685580c6ea0bb32cfe/build.sbt#L39-L43) version policy to `BinaryCompatible` in addition to relaxing the version policy of the server interpreters to `None`.

lolgab:
Ok, CI is green now. Ready for a review round @julienrf

julienrf (approved):

julienrf:
Thank you @lolgab for your help!

I see several people following the PR and reacting to it. I am wondering if the binary compatibility issue has been a real problem for you? Please feel free to comment on that! It is important that we know if there are any friction areas when you use endpoints4s.

julienrf:
Thank you again @lolgab, I’ve just released version 1.12.0 of endpoints4s with your changes.

PR #1281, opened 2024-08-30
Mark request bodies as required, closes #1280

### LINKED ISSUE #1280: Request bodies should be marked as required
using openapi module v4.5.1 i get a request descriprion like this:
```
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/CompleteRequest"
              }
            }
          }
        },
```
and it's missing `"required": true` field which means someone could call the endpoint without body

OlegYch:


julienrf:
Thank you! Could you please add a test to validate that the property is not set when we use `emptyRequest`?

OlegYch:
@julienrf this is already tested 
emptyRequest does not create a requestBody

julienrf (approved):

julienrf:
Thank you again @OlegYch, I’ve just released version 1.12.0 of endpoints4s with your changes.

PR #1288, opened 2024-09-17
Make openapi json more stable

OlegYch:
sort by name or preserve order defined in code where possible

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
@@ -29,8 +29,9 @@ trait EndpointsWithCustomErrors
     for (e <- endpoints) {
       val key = e.path
       val newValue = pathItems.get(key) match {
-        case Some(current) => PathItem(current.operations ++ e.item.operations)
-        case None          => PathItem(e.item.operations)
+        case Some(current) =>
+          PathItem(collection.immutable.ListMap(current.operations.toSeq: _*) ++ e.item.operations)
+        case None => PathItem(collection.immutable.ListMap(e.item.operations.toSeq: _*))
       }
       pathItems.update(key, newValue)
     }
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -70,7 +70,13 @@ object OpenApi {
 
   private def mapJson[A](map: collection.Map[String, A])(f: A => ujson.Value): ujson.Obj = {
     val result = ujson.Obj()
-    map.foreach { case (k, v) => result.value.put(k, f(v)) }
+    //preserve order defined by user or sort by key to minimize diff
+    val stableMap = map match {
+      case map: collection.mutable.LinkedHashMap[String, A] => map
+      case map: collection.immutable.ListMap[String, A]     => map
+      case map                                              => map.toSeq.sortBy(_._1)
+    }
+    stableMap.foreach { case (k, v) => result.value.put(k, f(v)) }
     result
   }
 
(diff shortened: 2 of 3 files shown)


julienrf:
Hello, thank you for your contribution! Would you mind elaborating on the reasons why minimizing the diffs of the OpenAPI document is important?

OlegYch:
well i just store it in git to check if it evolves compatibly

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -70,7 +70,12 @@ object OpenApi {
 
   private def mapJson[A](map: collection.Map[String, A])(f: A => ujson.Value): ujson.Obj = {
     val result = ujson.Obj()
-    map.foreach { case (k, v) => result.value.put(k, f(v)) }
+    val stableMap = map match {
+      case map: collection.mutable.LinkedHashMap[String, A] => map
+      case map: collection.immutable.ListMap[String, A]     => map
+      case map                                              => map.toSeq.sortBy(_._1)
+    }
+    stableMap.foreach { case (k, v) => result.value.put(k, f(v)) }
```
julienrf, line 78: Would you mind adding a brief comment in the code explaining what it does?

Also, I think it could be simplified to:

```suggestion
    val stableMap = map match {
      case seqMap: collection.SeqMap[String, A] => seqMap
      case map                                  => map.toSeq.sortBy(_._1)
    }
    stableMap.foreach { case (k, v) => result.value.put(k, f(v)) }
```

openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
```diff
@@ -29,8 +29,9 @@ trait EndpointsWithCustomErrors
     for (e <- endpoints) {
       val key = e.path
       val newValue = pathItems.get(key) match {
-        case Some(current) => PathItem(current.operations ++ e.item.operations)
-        case None          => PathItem(e.item.operations)
+        case Some(current) =>
+          PathItem(collection.immutable.ListMap(current.operations.toSeq: _*) ++ e.item.operations)
+        case None => PathItem(collection.immutable.ListMap(e.item.operations.toSeq: _*))
```
julienrf, line 34: What do you think of moving this to the implementation of `PathItem`? Ie, you could change the type of `operations` to be a `SeqMap`.

OlegYch: didn't want to break compatibility

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
OlegYch: makes sense

openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
julienrf: I believe this can be done without breaking the compatibility, no? Since `operations` is private in `PathItem`.

openapi/openapi/src/test/scala/endpoints4s/openapi/StableJsonTest.scala
julienrf, line 1: The test is not super robust but let’s keep it like that for now, that’s ok.

openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
OlegYch: withOperations is public

openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
OlegYch: oops SeqMap is not part of 2.12

openapi/openapi/src/main/scala/endpoints4s/openapi/Endpoints.scala
julienrf: Good point. And actually `operations` is public. Only the constructor is private…

julienrf:
Thank you again @OlegYch, I’ve just released version 1.12.0 of endpoints4s with your changes.

OlegYch:
cheers

julienrf:
Actually I made a mistake, but I will release 1.12.1 with your changes…

PR #1371, opened 2026-08-26
Add support for externalDocs in OpenAPI object

OlegYch:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#openapi-object

### DIFF
--- a/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
+++ b/openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
@@ -10,7 +10,8 @@ final class OpenApi private (
     val info: Info,
     private val innerPaths: collection.Map[String, PathItem],
     val components: Components,
-    val servers: Seq[Server]
+    val servers: Seq[Server],
+    val externalDocs: Option[ExternalDocumentationObject]
 ) extends Serializable {
 
   def paths: Map[String, PathItem] = innerPaths.toMap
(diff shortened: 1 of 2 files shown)


openapi/openapi/src/main/scala/endpoints4s/openapi/model/OpenApi.scala
```diff
@@ -48,25 +50,37 @@ final class OpenApi private (
...
+    new OpenApi(info, paths, components, servers, None)
+
+  def apply(
+      info: Info,
+      paths: collection.Map[String, PathItem],
+      components: Components,
+      servers: Seq[Server],
+      externalDocs: ExternalDocumentationObject
+  ): OpenApi =
+    new OpenApi(info, paths, components, servers, Some(externalDocs))
```
julienrf: Do you think we could avoid adding this overload? I don't remember why we already have that many overload but I think you could write `OpenApi(info, paths, components).withExternalDocs(Some(externalDocs))`. Would that work for you?

OlegYch: certainly

julienrf (approved):

julienrf:
Thank you @OlegYch!

julienrf:
Hey @OlegYch, I triggered a release workflow but it failed the `sbtVersionPolicyCheck` step. Would you mind having a look and relaxing the compatibility guarantees where necessary?
https://github.com/endpoints4s/endpoints4s/actions/runs/33206228768/job/98967857142

OlegYch:
@julienrf [fixed](https://github.com/endpoints4s/endpoints4s/pull/1372)

julienrf:
Thank you. Unfortunately, the build is not up to date and it is not anymore able to publish to Sonatype. Would you be interested in having a look?

OlegYch:
@julienrf this should work i think https://github.com/endpoints4s/endpoints4s/pull/1373

